PR #21, opened 2019-07-25
Add some conferences

philippemizrahi:
### Summary of Changes

Added some conferences + links to the readme


philippemizrahi:
@ttannis

ttannis:
Looks like you did it on top of your old work, so that PR makes it look like you're adding the roadmap again. You'll have to update your branch with the latest code from `master`, then push this PR back up. (and make sure that `.DS_store` is gone. 

Let me know on slack if you need clarification about what commands to run.

philippemizrahi:
gotcha, let me try that

feng-tao:
+1 on @ttannis for removing the unnecessary files.

ttannis:
> Thanks for cleaning up the history. Still have that mysterious `.DS_Store` file though. If you can delete that we're all set here.

@philippemizrahi when you get a chance

PR #125, opened 2019-10-07
addition of amundseninfrastructure via k8s + helm

javamonkey79:
### Summary of Changes

Added basic k8s+helm infrastructure for amundsen.

### Documentation

See the included readme

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,3 +17,6 @@
 [submodule "amundsencommon"]
 	path = amundsencommon
 	url = https://github.com/lyft/amundsencommon
+[submodule "amundseninfrastructure"]
+	path = amundseninfrastructure
+	url = https://github.com/lyft/amundseninfrastructure
--- a/amundseninfrastructure/README.md
+++ b/amundseninfrastructure/README.md
@@ -1,0 +1,39 @@
+# Amundsen Infrastructure
+
+## What is this?
+This is setup templates for deploying [amundsen](https://github.com/lyft/amundsen) on [k8s (kubernetes)](https://kubernetes.io/), using [helm.](https://helm.sh/) 
+
+## How do I get started?
+1. Make sure you have the following command line clients setup:
+    - k8s (kubectl)
+    - helm
+    - tiller
+2. Build out a cloud based k8s cluster, such as [amazon eks.](https://aws.amazon.com/eks/)
+3. Ensure you can connect to your cluster with cli tools in step 1.
+
+## How do I use this?
+You will need a values file to merge with these templates, in order to create the infrastructure. Here is an example:
+```
+environment: "dev"
+provider: aws
+dnsZone: teamname.company.com
+dockerhubImagePath: amundsendev
+searchServiceName: search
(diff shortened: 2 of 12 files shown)


.gitmodules
```diff
@@ -17,3 +17,6 @@
 [submodule "amundsencommon"]
 	path = amundsencommon
 	url = https://github.com/lyft/amundsencommon
+[submodule "amundseninfrastructure"]
```
feng-tao: gitsubmodules is for submodule actual repos here. I don't think we need to add a new submodules givens it is mostly doc.

javamonkey79: > gitsubmodules is for submodule actual repos here. 
I'm new to using submodules, but, seemed to me that this would ultimately end up as a submodule. I just wasn't sure how to make it work from a PR. If you don't think it should be a submodule, I can remove it. lmk

>  I don't think we need to add a new submodules givens it is mostly doc.
I don't think it is mostly documentation. There should be several template files used to create infrastructure. Am I missing something?

javamonkey79:
> thanks @javamonkey79 for the pr, could we make it a similar like [lyft/amundsenfrontendlibrary#224](https://github.com/lyft/amundsenfrontendlibrary/pull/224) which put it under the doc section. We could update the readme about this deployment option?

I could put the readme under docs/installation-k8s if you think that makes sense, but, I think the yaml files are infrastructure as code, so, I don't know that it would make sense under docs.

.gitmodules
feng-tao: sorry for the wait, how about like superset(https://github.com/apache/incubator-superset/tree/master/contrib/docker)? We create an folder like `contrib/infrastructure/k8s` ?
cc @jinhyukchang @bolkedebruin in case they have other ideas.

feng-tao:
@javamonkey79 let me know what your think of my comment, also the CLA check fails. You may need to configure your github user/email locally first.

.gitmodules
javamonkey79: @feng-tao I'm not sure the necessity of the contrib folder, but, I don't mind it either. Just lmk what you prefer, I just went with what we had internally for another k8s based project.

feng-tao:
the cla check is still not happy :(

.gitmodules
feng-tao: contrib is for `contribute` which I borrowed from superset or Airflow :)

javamonkey79:
> the cla check is still not happy :(

Thanks, I'll correct it today.

javamonkey79:
@feng-tao I'm not sure why I am getting the CLA check; I am going to close this PR and open a new one. I'll also leave out the module bit.

dmateusp:
@javamonkey79 I was looking for Helm resources for this project, was there a new PR containing this work ?

javamonkey79:
@dmateusp the PR was merged a little while ago, and you can find those resources here https://github.com/lyft/amundsen/tree/master/amundsen-kube-helm

jai2033shankar (commented):
Is there anything I am missing here?

PR #140, opened 2019-10-18
Rename Amundsen infrastructure to amundsen-kube-helm

feng-tao:
looking at https://github.com/BrechtDeVlieger/airflow-kube-helm which {project}-kube-helm seems to be more intuitive than current existing name. I propose to switch to that name pattern. Let me know what you think.


javamonkey79:
Seems fine to me; the name I used was just what we decide on internally, but, k8s+helm is new to use, so, makes sense that naming is developing.

PR #151, opened 2019-10-27
Local docker dev speedup and simplification

jornh:
### Summary of Changes

Cut some time and things to handle in your head/keystrokes away:
* Little need to limit the build to single containers. Due to docker caching additional build time is negligible. 
* No need to `stop` as part of compose relaunch, `up` will detect changed containers and only relaunch as needed
* Also, indent code snippets in the numbered steps

### Documentation

yep!

### CheckList
- [x] I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -29,22 +29,24 @@ $ git submodule update --remote
 
 ### Building local changes
 1. First, be sure that you have first followed the [installation documentation](https://github.com/lyft/amundsen/blob/master/docs/installation.md#bootstrap-a-default-version-of-amundsen-using-docker) and can spin up a default version of Amundsen without any issues. If you have already completed this step, be sure to have stopped and removed those containers by running:
-```bash
-$ docker-compose -f docker-amundsen.yml down
-```
-2. Launch the containers needed for local development by running:
-```bash
-$ docker-compose -f docker-amundsen-local.yml up
-```
-3. After making local changes rebuild the containers for any submodule that was modified. For example, to build a local change in `amundsen/amundsenmetadatalibrary` run:
-```bash
-$ docker-compose -f docker-amundsen-local.yml build amundsenmetadata
-```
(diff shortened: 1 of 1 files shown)


jornh:
@ttannis (or Daniel?) PTAL when you have a few cycles ...

I think I shaved off quite some time on specially docker stop/up waiting. 

Ultimately we want to be able to do hot reloading on the frontend with volume-mount running in a container. But that can come as another PR. This current PR mostly is good for Flask dev ... though it shouldn't hurt the TypeScript dev trying to use Docker.

docs/developer_guide.md
```diff
@@ -29,22 +29,24 @@ $ git submodule update --remote
...
+    $ docker-compose -f docker-amundsen-local.yml up -d
+    ```
+3. After making local changes rebuild and relaunch modified containers:
+    ```bash
+    $ docker-compose -f docker-amundsen-local.yml build \
+      && docker-compose -f docker-amundsen-local.yml up -d
+    ```
+4. Optionally, to still tail logs, in a different terminal you can:
+    ```bash
+    $ docker-compose -f docker-amundsen-local.yml logs --tail=3 -f
```
feng-tao: not familiar with the context, but --tail=3 and --tail 10 work for both scenarios?

jornh: Yes, I think they do. What I did actually just reflected the `--help` guidance given for each of the commands 🤷‍♂ 

We can drop the `=` sign if it bothers you. The difference in 3 vs 10 lines was actually on purpose because having 10 lines from one container makes sense where you don't want to read through 50 lines of log across 5 containers. You just want to see they're all alive and then follow the tail from there ...

feng-tao: it is ok

feng-tao:
thanks @jornh

PR #192, opened 2019-11-26
Adding resources, tolerations, node selectors, and affinities to the helm charts

lukelowery:
I have added resources, tolerations, node selectors, and affinities to the helm charts. We need these to deploy to our kubernetes cluster. I also added some default value files to charts.

### DIFF
--- a/amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
+++ b/amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
@@ -7,12 +7,24 @@ spec:
   selector:
     matchLabels:
       run: {{ .Chart.Name }}-{{ .Values.searchServiceName }}
-  replicas: 1
+  replicas: {{ default 1 .Values.search.replicas }}
   template:
     metadata:
       labels:
         run: {{ .Chart.Name }}-{{ .Values.searchServiceName }}
     spec:
+      {{- with default .Values.nodeSelector .Values.search.nodeSelector }}
+      nodeSelector:
+{{ toYaml . | indent 8 }}
+      {{- end }}
+      {{- with default .Values.affinity .Values.search.affinity }}
+      affinity:
+{{ toYaml . | indent 8 }}
+      {{- end }}
+      {{- with default .Values.tolerations .Values.search.tolerations }}
+      tolerations:
+{{ toYaml . | indent 8 }}
+      {{- end }}
       containers:
       - name: {{ .Chart.Name }}-{{ .Values.searchServiceName }}
         image: {{- if .Values.searchServiceImage }} {{.Values.searchServiceImage}}{{- else }} {{ .Values.dockerhubImagePath }}/{{ .Chart.Name }}-{{ .Values.searchServiceName }}:{{ .Values.searchImageVersion }}{{- end }}
@@ -21,6 +33,10 @@ spec:
         env:
         - name: PROXY_ENDPOINT
           value: elasticsearch  
+        {{- with .Values.search.resources }}
+        resources:
+{{ toYaml . | indent 10 }}
+        {{- end }}
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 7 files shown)


feng-tao:
cc @javamonkey79 Shaun, could you take a look given you have the most context on this :) ?Thanks.

javamonkey79:
> cc @javamonkey79 Shaun, could you take a look given you have the most context on this :) ?Thanks.

Yes, I'll check it out.

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
```diff
@@ -44,6 +72,10 @@ spec:
         env:
         - name: PROXY_HOST
           value: bolt://neo4j
+        {{- with .Values.metadata.esources }}
```
javamonkey79: I think this should be `resources`?

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
```diff
@@ -30,12 +46,24 @@ spec:
...
     matchLabels:
       run: {{ .Chart.Name }}-{{ .Values.metadataServiceName }}
-  replicas: 1
+  replicas: {{ .Values.metadataReplicas }}
   template:
     metadata:
       labels:
         run: {{ .Chart.Name }}-{{ .Values.metadataServiceName }}
     spec:
+      {{- with default .Values.nodeSelector .Values.metadata.nodeSelector }}
```
javamonkey79: wdyt about wrapping the nodeSelector, affinity, tolerations sections in an if block, such that they are not required in all values files?

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
```diff
@@ -7,12 +7,24 @@ spec:
   selector:
     matchLabels:
       run: {{ .Chart.Name }}-{{ .Values.searchServiceName }}
-  replicas: 1
+  replicas: {{ .Values.search.replicas }}
```
javamonkey79: Could you set a default here?

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
javamonkey79: Actually, it looks like `with default` will cause this to not be used, if neither nodeSelector is defined?

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
```diff
@@ -53,12 +85,24 @@ spec:
   selector:
     matchLabels:
       run: {{ .Chart.Name }}-{{ .Values.frontEndServiceName }}
-  replicas: 1
+  replicas: {{ .Values.frontEnd.replicas }}
```
javamonkey79: default here too?

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
lukelowery: Yeah. If both are set the `metadata.nodeSelector` one overrides the `nodeSelector`. If neither are set you don't get anything in the with clause.

javamonkey79: Cool, thanks TIL

javamonkey79:
@lukelowery FYI, also, you are not passing the CLA check...

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
lukelowery: Alright. 

I have the default set in the values file. I think that is a better practice, but it is not a big deal.

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
lukelowery: oops! Thanks :)

amundsen-kube-helm/templates/helm/amundsen/templates/deployment.yaml
javamonkey79: Yeah, I think overall having it in the values file makes sense too, but, defaulting it here means you don't have to have it in the values file... which is a little more succinct. Anyways, thanks for the change.

javamonkey79:
@feng-tao it looks good to me. I've approved it, but, it does not seem to show as approved. I'm guessing I lack permission.

feng-tao:
thanks @javamonkey79 for taking a look, I could do that.

@lukelowery could you configure your github email/user name locally and repush it again? This will fix the CLA check :)

lukelowery:
@feng-tao Looks like I got my gpg keys working right now.

feng-tao:
@lukelowery sorry for the delay(was in OOO last week). merge it now.

PR #212, opened 2019-12-20
setting up a pvc to allow for neo4j backups

javamonkey79:
### Summary of Changes

* In order to allow other pods to access neo4j disk, we need to setup a pvc and point neo4j at it

### Documentation

* No doc changes

### CheckList
- [X] PR title addresses the issue accurately and concisely.

### DIFF
--- a/amundsen-kube-helm/templates/helm/neo4j/templates/deployment.yaml
+++ b/amundsen-kube-helm/templates/helm/neo4j/templates/deployment.yaml
@@ -38,15 +38,19 @@ spec:
         volumeMounts:
         - name: conf
           mountPath: /conf
+        {{- if (or (kindIs "invalid" .Values.neo4jpvc) .Values.neo4jpvc)}}
         - name: data
           mountPath: /var/lib/neo4j/data
+        {{- end}}
       volumes:
         - name: conf
           configMap:
             name: {{ .Chart.Name }}-configmap
+        {{- if (or (kindIs "invalid" .Values.neo4jpvc) .Values.neo4jpvc)}}
         - name: data
           persistentVolumeClaim:
             claimName: neo4j-pvc
+        {{- end}}
       {{- with .Values.resources }}
       resources:
 {{ toYaml . | indent 8 }}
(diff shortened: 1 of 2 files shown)


javamonkey79:
@bryfox @lukelowery can you take a look at this as well? I'm also a bit new to k8s\helm, and hence I'm not sure if there is a better way to do this. Basically, we need to open up /var/lib/neo4j/data such that other pods\services can then access it and back it up. 

cc @feng-tao

javamonkey79:
Also, FYI, I've tested this on a dev cluster and it seems to work. After I made this update, I ran this cron pod:
```
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: test-mount
spec:
  schedule: "5 * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: test
            image: "appropriate/curl:latest"
            command:
             - "/bin/sh"
             - "-c"
             - |
               echo "hello"
               ls /var/lib/neo4j/data
               echo "goodbye"
            volumeMounts:
              - name: data
                mountPath: /var/lib/neo4j/data
          restartPolicy: OnFailure
          volumes:
            - name: data
              persistentVolumeClaim:
                claimName: neo4j-pvc
```

And, I did note from the logs:
```
➜  neo4j-backup kubectl logs test-mount-1576865100-6cln9
hello
databases
dbms
goodbye
```

lukelowery:
This is a good way to do it if you don't need more than one neo4j pod. 

I would wrap the PVC stuff in an if statement so people can turn it off easily if they want.

I think the best way is to have a real persistent storage volume that's backed by something robust, then you don't need to make backups at all, but I totally get this approach.

javamonkey79:
@lukelowery thanks for that tip; I've made the changes, to allow the pvc to be omitted. It seems the sprig-go default function would not work as I expected, since it evaluates the property and if it is found to be false, then it uses the default value. So, to work around this, I had to check `kindIs "invalid" .Values.neo4jpvc` to see if the value is not populated. If you know a better way, lmk.

javamonkey79:
@feng-tao I think this is ready to do, can you merge?

PR #240, opened 2020-01-21
change default ports to 5001,5002 for search, library

friendtocephalopods:
### Summary of Changes

Change ports to 5001, 5002 for search, metadatalibrary. Related to https://github.com/lyft/amundsenmetadatalibrary/pull/102 and https://github.com/lyft/amundsensearchlibrary/pull/80

### Documentation

None

### CheckList

### DIFF
--- a/docker-amundsen-atlas.yml
+++ b/docker-amundsen-atlas.yml
@@ -14,7 +14,7 @@ services:
       image: amundsendev/amundsen-search:1.4.2
       container_name: amundsensearch
       ports:
-        - 5001:5000
+        - 5001:5001
       environment:
         - CREDENTIALS_PROXY_USER=admin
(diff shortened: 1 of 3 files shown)


docker-amundsen-atlas.yml
```diff
@@ -14,7 +14,7 @@ services:
       image: amundsendev/amundsen-search:1.4.2
       container_name: amundsensearch
       ports:
-        - 5001:5000
+        - 5001:5001
```
feng-tao: given the tag image is same as the pypi version, I think we don't need to change it now. Once we bump the pypi version, we could change that. Otherwise the docker file may not work.

My bad, initially I thought we need to change the port for every file.

PR #298, opened 2020-02-21
Update READ.md adding service version guide

jinhyukchang:
### Summary of Changes

Service version guide on v2.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,20 +79,6 @@ The databases we see most frequently used in the community are:
 
 Please visit [Installation guideline](docs/installation.md) on how to install Amundsen.
 
-## Service version guide
-### Amundsen 1.x -> 2.x Migration Guide:
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -79,6 +79,20 @@ The databases we see most frequently used in the community are:
 
 Please visit [Installation guideline](docs/installation.md) on how to install Amundsen.
 
+## Service version guide
```
feng-tao: about a migration guide or faq doc?

ttannis: I like the idea of a dedicated dos because over time we will keep adding to it. This section could stay here for high visibility, but just link to the separate doc. Fine to merge though.

feng-tao: yeah, once we have a faq doc, we could copy some of the repetitive slack q & a into that doc as well.

jinhyukchang: +1 on separate doc. Will update!

jinhyukchang:
@ttannis @feng-tao Created faq.md . Let me know.

PR #354, opened 2020-03-30
[WIP] Configurable .Values.frontEnd.OIDC_PROVIDER

jornh:
### Summary of Changes

Adds use of a new `.Values.frontEnd.OIDC_PROVIDER`.

If the value is not set or set to something different than "azure" or "google" it should behave as before. (I could need help with testing that)

Note setting to either "azure" or "google" essentially skips token validation by setting `"token_introspection_uri": ""` (to a blank string).

⚠️ of course it shouldn't be taken lightheartedly to skip id token validation (through a _remote_ endpoint, which neither Google, nor Azure AD appears to offer), which is _exactly_ why for those providers we _NEED_ to pair with `offline` validation instead - as can be added through the other WIP PR lyft/amundsenfrontendlibrary#421 (hope that made sense?)

### Documentation

**TODO** - needs addition in the values file and in README.md

In addition https://github.com/lyft/amundsen/blob/master/docs/authentication/oidc.md#using-okta-with-amundsen-on-k8s should be adjusted/generalized a bit

### CheckList

### DIFF
--- a/amundsen-kube-helm/templates/helm/templates/amundsen-deployment.yaml
+++ b/amundsen-kube-helm/templates/helm/templates/amundsen-deployment.yaml
@@ -149,6 +149,10 @@ spec:
           - name: FRONTEND_BASE
             value: http://{{ .Values.frontEnd.FRONTEND_BASE }}
           {{ end }}
+          {{- if .Values.frontEnd.OIDC_OVERWRITE_REDIRECT_URI }}
+          - name: FLASK_OIDC_REDIRECT_OVERRIDE
+            value: {{ .Values.frontEnd.OIDC_OVERWRITE_REDIRECT_URI }}
+          {{- end }}
           - name: SEARCHSERVICE_BASE
             value: http://{{ .Chart.Name }}-{{ .Values.search.serviceName }}:5001
           - name: METADATASERVICE_BASE
--- a/amundsen-kube-helm/templates/helm/templates/amundsen-oidc-config.yaml
+++ b/amundsen-kube-helm/templates/helm/templates/amundsen-oidc-config.yaml
@@ -31,7 +31,7 @@ stringData:
         "userinfo_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}/v1/userinfo",
         {{- end }}
         "redirect_uris": [
-          "http://localhost/oidc_callback"
+          "{{ .Values.frontEnd.OIDC_OVERWRITE_REDIRECT_URI }}"
         ],
         {{- if (eq .Values.frontEnd.OIDC_PROVIDER "azure") or (eq .Values.frontEnd.OIDC_PROVIDER "google") }}
         "token_introspection_uri": ""
--- a/amundsen-kube-helm/templates/helm/templates/amundsen-oidc-config.yaml
+++ b/amundsen-kube-helm/templates/helm/templates/amundsen-oidc-config.yaml
@@ -11,14 +11,33 @@ stringData:
       "web": {
         "client_id": "{{ .Values.frontEnd.OIDC_CLIENT_ID }}",
         "client_secret": "{{ .Values.frontEnd.OIDC_CLIENT_SECRET }}",
+
+        {{- if eq .Values.frontEnd.OIDC_PROVIDER "azure" }}
+        "auth_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/v2.0/authorize",
+        "token_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/v2.0/token",
+        "issuer": "{{ .Values.frontEnd.OIDC_ORG_URL }}/v2.0",
+        "userinfo_uri": "https://graph.microsoft.com/oidc/userinfo",
+
+        {{- else if eq .Values.frontEnd.OIDC_PROVIDER "google" }}
+        "auth_uri": "https://accounts.google.com/o/oauth2/v2/auth",
+        "token_uri": "https://oauth2.googleapis.com/token",
+        "issuer": "https://accounts.google.com",
+        "userinfo_uri": "https://openidconnect.googleapis.com/v1/userinfo",
+
+        {{- else }}
         "auth_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}/v1/authorize",
         "token_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}/v1/token",
         "issuer": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}",
         "userinfo_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}/v1/userinfo",
+        {{- end }}
         "redirect_uris": [
           "http://localhost/oidc_callback"
         ],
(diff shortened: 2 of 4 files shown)


amundsen-kube-helm/templates/helm/templates/amundsen-oidc-config.yaml
```diff
@@ -11,14 +11,33 @@ stringData:
...
+        "auth_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/v2.0/authorize",
+        "token_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/v2.0/token",
+        "issuer": "{{ .Values.frontEnd.OIDC_ORG_URL }}/v2.0",
+        "userinfo_uri": "https://graph.microsoft.com/oidc/userinfo",
+
+        {{- else if eq .Values.frontEnd.OIDC_PROVIDER "google" }}
+        "auth_uri": "https://accounts.google.com/o/oauth2/v2/auth",
+        "token_uri": "https://oauth2.googleapis.com/token",
+        "issuer": "https://accounts.google.com",
+        "userinfo_uri": "https://openidconnect.googleapis.com/v1/userinfo",
```
josibake: this looks good. long term, it might be better to move the logic around which uri's to use into an `oidc-config.py` or whatever is responsible for creating the secrets file. this is because google uri's will change , which would then require updating the helm chart every time an endpoint changes. you can always get the latest uri's from google from `https://accounts.google.com/.well-known/openid-configuration`. i might take a stab at this in a later PR

jornh: Got it. Just wanted to go with the simpler static approach to get off the ground. 

I recently learned about Helm hooks - maybe they're useful to run the code`(would that be the right time/place). I think it's called OpenID Connect Discovery - there's already a tool for it here: 

https://github.com/puiterwijk/flask-oidc/blob/7f16e27b926fc12953d6b2ae78a9b9cc9b8d1769/flask_oidc/registration_util.py#L65

Oh, yeah and if you want to get cracking away on something you should actually be able to get this to login a user with Google, as I did, right now (there is just no validation of the token). Having that up and running on your end would mean you're prepared for testing the validation, figuring out other parts needed like documentation etc. and you have a demo environment.

josibake: Great! thanks! I'll start playing around with it. I hacked around on a chart earlier and got it working with the frontend, but the problem I ran into was calling the metadata service and search service because of the lack of token validation. Once I created a dev Okta account it was working tho, so I think thats the only missing piece. I'll looking into the helm hooks too, I'd never heard of that

jornh: Excellent! In the meantime I've got myself setup with a clone of needed intermediate pieces (Verdan's `flaskoidc` wrapper + `ammundsenfrontendlibrary` + local Kubernetes on Docker Desktop + Helm chart PR above) and a way to locally build ... 

* I'll try to get WiP PR's for that pushed for you (& possibly others) to see. Maybe also description of the build steps (it's a bit of Russian doll game due to all the layers, rebuild, testing, rinse & repeat 🙂)
* So far I've successfully swapped in `flask-oidc-ex` and added env vars from Damian's `dcrdev/flaskoidc` fork and my Google testing login still works with the frontend (without validation disabled as per Helm above)
* On the sad side even though I added Damian's CLOCK_SKEW I'm still not able to login with Azure.
* Next up is adding the `offline` validation as a config option, then dig more into Azure login.

I guess with your Okta account you can cover testing that with the changes we introduce nothing gets broken with that provider. We need to get similar for Keycloak as well. 

I think we should also move out discussion from here and create a proper issue to track across PRs and get more visibility.

jornh:
Update: I’ll declare a mild success 😉 with AAD. I got past the clock skew/validation error.

I pushed https://github.com/jornh/flaskoidc/commit/b08f659a98886070d60c9d8fef869a09476de47e which can be 
```
RUN pip install https://github.com/jornh/flaskoidc/archive/b08f659a98886070d60c9d8fef869a09476de47e.zip
```
... installed in the frontend `public.Dockerfile` in the OIDC build stage before setup.py
The dockerfiles OIDC version then needs to be set to `0.1.0b4` and rebuilt

amundsen-kube-helm/templates/helm/templates/amundsen-oidc-config.yaml
```diff
@@ -11,14 +11,33 @@ stringData:
...
+
+        {{- else }}
         "auth_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}/v1/authorize",
         "token_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}/v1/token",
         "issuer": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}",
         "userinfo_uri": "{{ .Values.frontEnd.OIDC_ORG_URL }}/oauth2/{{ .Values.frontEnd.OIDC_AUTH_SERVER_ID }}/v1/userinfo",
+        {{- end }}
         "redirect_uris": [
           "http://localhost/oidc_callback"
```
hashhar, line 34: Is this expected to be without the port? I'm not sure if anything's running on 80.

hashhar:
@jornh It would be great if you could add the implications of skipping token validation to the opening comment of the PR (and maybe the commit message as well). I think the Okta docs at https://developer.okta.com/docs/guides/validate-id-tokens/overview/#validating-a-token-remotely-with-okta mention it clearly.

amundsen-kube-helm/templates/helm/templates/amundsen-oidc-config.yaml
jornh: Good catch .. I'll look into it - it might have to become configurable. Though, I think that particular line is just the as-is state of the Helm Chart.

jornh:
Thank you for Okta doc pointer. Will look into how to cover it, most likely also need to be included in updates to `/docs/authentication/oicd.md` where I have a draft sitting locally.

jornh:
Hah! yes I see, was re-reading PR opening comment above now... and of course it shouldn't be taken lightheartedly to skip id token validation (through a _remote_ endpoint, which neither Google, nor Azure AD appears to offer), which is _exactly_ why for those providers we _NEED_ to pair with `offline` validation instead - as can be added through the other WIP PR lyft/amundsenfrontendlibrary#421 (hope that made sense?)

amundsen-kube-helm/templates/helm/templates/amundsen-deployment.yaml
```diff
@@ -149,6 +149,10 @@ spec:
           - name: FRONTEND_BASE
             value: http://{{ .Values.frontEnd.FRONTEND_BASE }}
           {{ end }}
+          {{- if .Values.frontEnd.OIDC_OVERWRITE_REDIRECT_URI }}
+          - name: FLASK_OIDC_REDIRECT_OVERRIDE
```
pPanda-beta, line 153: How are you configuring the FLASK_OIDC_REDIRECT_OVERRIDE ? 
Shouldn't it be FLASK_OIDC_OVERWRITE_REDIRECT_URI? as the official documentations says : OVERWRITE_REDIRECT_URI 

Here: https://flask-oidc.readthedocs.io/en/latest/

jornh: It’s configured here: https://github.com/jornh/flaskoidc/commit/9368ec24ffacbd03590a03b2986fcc5792fa5deb You are right there’s a naming inconsistency across the layers. I choose to carry it forward to be compatible with the other fork of Verdan’s `flaskoidc` wrapper layer.

AlejoGB:
@jornh Thank you for your solution.
 I made Azure SSO work with offline validation, but the api/auth_user GET request returns a 500 response:

`
{
   "msg": "Encountered exception: Expecting value: line 1 column 1 (char 0)"
}`

I understand this may be related to an error obtaining the user_id from id_token, but i cant find it on flask_oidc_ex process_callback.

PR #355, opened 2020-03-30
Repairing the link to sample DAG

fBedecarrats:
In 
> Users could either load the data with [a python script](https://github.com/lyft/amundsendatabuilder/blob/master/example/scripts/sample_data_loader.py) with the library or with an [Airflow DAG](https://github.com/lyft/amundsendatabuilder/blob/master/example/dags/sample_dag.py) importing the library.

the link to the DAG example was broken, as the folder and file it pointed to no longer exist. I replaced the url with the new corresponding folder, but I didn't point to a specific file as there are now 3 different samples provided as example. So the link points to the folder: https://github.com/lyft/amundsendatabuilder/tree/master/example/dags


feng-tao:
thanks.

PR #384, opened 2020-04-15
Align helm images with docker-compose yaml

ankusharya:
### Summary of Changes

I have aligned the helm chart with docker-compose.yaml for amundsen. Also fixed the PROXY_ENDPOINT url, which was set to namespace instead of name. Tested in local helm with docker-for-mac, also tested in aws EKS.

### DIFF
--- a/amundsen-kube-helm/templates/helm/templates/amundsen-deployment.yaml
+++ b/amundsen-kube-helm/templates/helm/templates/amundsen-deployment.yaml
@@ -32,7 +32,7 @@ spec:
         - containerPort: 5000  
         env:
         - name: PROXY_ENDPOINT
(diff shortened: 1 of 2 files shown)


ankusharya:
As i can see all checks failed, i am closing the pull request

jornh:
Hi @ankusharya I think I spotted what the problem could be before you closed the issue.

The CLA - open source signature form often fails when people do their first commit. This usually is the case because the email address of the commit does not match the email for the GitHub user (you see the )Please have user on https://github.com/lyft/amundsen/pull/384/commits/caf8d803f3bacbeb7a7930780e796a2341a03281 does not have same user/icon as your GitHub user @ankusharya green avatar 

So, please have another go at it - just as a new PR - with commit email matching the one on your GitHub account then I'm pretty sure it will work

😄 The alignment of the Helm chart would be very welcome, now that you already tested it. Thanks!

feng-tao:
@ankusharya normally the CLA check just requires you first config your github setting(email, id) locally before pushing the pr:)

PR #395, opened 2020-04-17
Add additional features to Helm Chart

ajbosco:
### Summary of Changes
* Separates microservices into their own templates for easier readability
* Adds ability to add `annotations` for services
* Adds ability to add `podAnnotations` for deployments
* Adds ability to change `serviceType` rather than hardcode to `LoadBalancer`
* Adds additional metadata and labels to each component
* Adds optional `PersistentVolume` for neo4j backups that works with AWS EFS, which has 
 `ReadWriteMany` access mode. Previously, the pvc was hardcoded to a `ReadWriteOnce` so the backup pod was unable to mount the volume since it was in use by neo4j itself.

### Documentation
* Updated `README.md` for helm chart with new chart values

### CheckList


feng-tao:
this is great!
cc @javamonkey79 @lukelowery some folks to take a look.

feng-tao:
also @jornh

amundsen-kube-helm/templates/helm/templates/pv-neo4j.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if and .Values.neo4j.enabled .Values.neo4j.persistence .Values.neo4j.persistence.efs }}
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: neo4j-pvc
```
javamonkey79, line 5: MINOR: name should be `neo4j-pv`

amundsen-kube-helm/templates/helm/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,100 @@
...
+      {{- end }}
+      containers:
+      - name: {{ .Chart.Name }}-{{ .Values.frontEnd.serviceName }}
+        image: {{ .Values.frontEnd.image }}:{{ .Values.frontEnd.imageTag }}
+        imagePullPolicy: Always
+        ports:
+          - containerPort: 5000
+        env:
+          - name: FRONTEND_BASE
+            value: {{ .Values.frontEnd.fqdn }}
```
jornh, line 56: Where does `.Values.frontEnd.fqdn` come from? Should it also be documented in README.md?

Admittedly pretty new to kube and helm, so this could just be my lack of understanding.

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -40,6 +40,10 @@ affinity: {}
 ## tolerations -- amundsen application wide configuration of tolerations. This applies to search, metadata, frontend and neo4j. Elasticsearch has it's own configuation properties for this. [ref](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#taints-and-tolerations-beta-feature)
 ##
 tolerations: []
+##
+## podAnnotations -- amundsen application wide configuration of podAnnotations. This applies to search, metadata, frontend and neo4j. Elasticsearch has it's own configuation properties for this. [ref](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
+##
+podAnnotations: {}
```
jornh, line 46: Neat trick!

amundsen-kube-helm/templates/helm/templates/deployment-frontend.yaml
ajbosco: Ah, good catch! I missed that one. Updated and added to README.md and values.yaml.

ajbosco:
> Overall, I like what you have done - nice job! I only did a single pass, and I don't have a lot of time atm; I'll take a 2nd pass later this week.
> 
> My only major consideration at this point, is various name construction e.g. this type of behavior:
> 
> ```
> {{ template "amundsen.fullname" . }}-{{ .Values.frontEnd.serviceName }}
> ```
> 
> I found those to be cumbersome and unnecessary. I think it would be better if it were either baked in as a static value, or, allow the whole thing to be changed, with the static value as a default.
> 
> Unless there is some case I'm missing?

Using a name value like this has similar behavior to the existing names like [here](https://github.com/lyft/amundsen/blob/master/amundsen-kube-helm/templates/helm/templates/amundsen-deployment.yaml#L5), but allows for overrides as well. It seemed to be a common pattern in some of the other Helm charts I've used like this [Airflow](https://github.com/helm/charts/blob/master/stable/airflow/templates/deployments-web.yaml#L4) one.

feng-tao:
thanks @jornh @javamonkey79 for the review, and thanks @ajbosco for the contribution!

PR #398, opened 2020-04-17
Fix a badge to be https and change color

feng-tao:
### Summary of Changes

Screen Shot 2020-04-17 at 4 36 31 PM

All the supported color could be found at https://www.materialui.co/colors . Felt the new one looks better, but ok to not change.

Also fix a badge link to be https

### CheckList

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -13,8 +13,8 @@ theme:
   logo: img/logos/amundsen_mark_orange.svg
   favicon: 'img/logos/amundsen_mark_orange.svg'
   palette:
-    primary: 'blue'
-    accent: 'pink'
+    primary: 'deep purple'
+    accent: 'deep orange'
   feature:
     tabs: true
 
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -13,8 +13,8 @@ theme:
   logo: img/logos/amundsen_mark_orange.svg
   favicon: 'img/logos/amundsen_mark_orange.svg'
   palette:
-    primary: 'blue'
-    accent: 'pink'
+    primary: 'deep purple'
```
ttannis, line 16: Do they only support specific color names? Any place where we can understand the options and see if we can manually enter colors by "hex code"?

ttannis: As in, I see the colors we could choose from but looking for if this supports hex codes. Our main purple hex code is `#2B1B81` and the "orange" colors is `#FF623E`

feng-tao: @ttannis it supports, update the pr to use that main purple color.

MJFND:
This Color looks awesome.

ttannis:
Weird, so the background of the header on the current site isn't the same as the `primary` color we entered. It's fine though I'll look at it in some free to to understand how the site works.

PR #422, opened 2020-04-28
Add "A Data Engineer's Perspective On Data Democratization"

dmateusp:
Adds a blog post I wrote, in the related posts sections

The blog post covers Metadata hubs in general and Amundsen

I asked @feng-tao on Slack if it was relevant and where I could share it, so thank you for the initial review!

This is a friends link, so readers won't be stopped by the Medium paywall


feng-tao:
thanks!

PR #471, opened 2020-05-28
docs: Updates PR template and adds document for labeling explanation

Golodhros:
### Summary of Changes
Updates PR template
Adds documentation about labeling

### CheckList


docs/issue_labeling.md
```diff
@@ -0,0 +1,60 @@
+# Issue and Feature Labeling
```
feng-tao, line 1: should we add it under this dev section (https://github.com/lyft/amundsen/blob/master/mkdocs.yml#L61) of our doc?

Golodhros: That makes sense, I guess I just need to add another entry on that yml file?

feng-tao: correct, then follow https://github.com/lyft/amundsen/blob/master/deploy_website.sh#L8-L12 to test locally on the doc side and verify.

Golodhros: Awesome, thanks!

jornh: I like the labeling.

How will it work for non-Lyft contributors? Currently I think it’s only available for you people. Maybe it’s just a GitHub setting adjustment or something else needed ...

Golodhros: Oh, true, thanks for the heads up @jornh!

We'll probably start with keeping it only for Lyft developers, but it might have sense to open it up.

Issue #486, opened 2020-06-05
Using helm 3 with the default values.yaml search service is not able to connect to the ES


ibnipun10:
Following https://github.com/lyft/amundsen/tree/master/amundsen-kube-helm to install amundsen to kube using helm 3. Using the default values without providing the ES endpoint in the values.yaml, it search service uses `{{ .Release.Namespace }}-elasticsearch-client.{{ .Release.Namespace }}.svc.cluster.local` as the url to connect to ES. It gives error as it does not find this service. It should be Release.Name in the beginning of the url to connect

## Expected Behavior
Search service should connect to the ES provided by the url in the deployment-searhc.yml file

## Current Behavior
With helm 3 it gives error using the default values 

## Possible Solution
Replace {{ .Release.Namespace }}-elasticsearch-client.{{ .Release.Namespace }}.svc.cluster.local with
{{ .Release.Name }}-elasticsearch-client.{{ .Release.Namespace }}.svc.cluster.local

## Steps to Reproduce
1. Clone https://github.com/lyft/amundsen/tree/master/amundsen-kube-helm
2. WIth helm 3, install using the command : helm install my-amundsen ./templates/helm --values impl/helm/dev/values.yaml
3. Push sample data using databuilder
4. On Amundsen UI, you will not be able to search. 

## Context
Not able to search in the amundsen UI. 

## Your Environment
* Amunsen version used: Master branch as of today
* Data warehouse stores: 
* Deployment (k8s or native): AKS
* Link to your fork or repository: Master branch

PR #488, opened 2020-06-06
chore: Updated the deployment search file with Release.Name instead of Release.Namespace

ibnipun10:
Check this issue : #486

Issue
Following https://github.com/lyft/amundsen/tree/master/amundsen-kube-helm to install amundsen to kube using helm 3. Using the default values without providing the ES endpoint in the values.yaml, it search service uses {{ .Release.Namespace }}-elasticsearch-client.{{ .Release.Namespace }}.svc.cluster.local as the url to connect to ES. It gives error as it does not find this service. It should be Release.Name in the beginning of the url to connect

Summary of Changes
Replaced {{ .Release.Namespace }}-elasticsearch-client.{{ .Release.Namespace }}.svc.cluster.local with {{ .Release.Name }}-elasticsearch-client.{{ .Release.Namespace }}.svc.cluster.local

Documentation
NA

### DIFF
--- a/amundsen-kube-helm/templates/helm/templates/deployment-search.yaml
+++ b/amundsen-kube-helm/templates/helm/templates/deployment-search.yaml
@@ -46,7 +46,7 @@ spec:
         - containerPort: 5001
         env:
         - name: PROXY_ENDPOINT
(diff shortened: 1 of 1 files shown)


feng-tao:
thanks @ibnipun10

PR #528, opened 2020-06-26
fix: update helm chart to prevent installation of elastic search resources when it's disabled

diegoep:
Currently the property elasticsearch.enabled is ignored so that the elastic search is always installed. I've added a checking for this property to the deployment and service templates so that it would be possible to use another elasticsearch installation without having to install a fresh one.


feng-tao:
not sure why prlint fails,let me close and reopen the pr.

feng-tao:
cc @javamonkey79

javamonkey79:
This is turning off the search microservices when elastic search is disabled, which, I think is a bit of a confusing coupling. iirc, when you set elasticsearch.enabled = false, it should not install the dependent ES chart. That is what [this is for.](https://github.com/lyft/amundsen/blob/master/amundsen-kube-helm/templates/helm/requirements.yaml#L8). I think if it is still installing then, we should get to the bottom of why, try to reproduce it, etc. 

If ES is not installing when enabled is set to false, then disabling the search service is another topic; will Amundsen function ok with it off? Specifically, the metadata service, front end, etc. If it does still work, then I think changing the flag to `{{- if .Values.searchservice.enabled }}` makes more sense.

diegoep:
Yes, it was a bit of confusion regarding the ES configuration. To disable it is just to remove it from the requirements.yaml. I'll close this PR.

PR #699, opened 2020-09-15
other: Add persistent volumes for docker-compose images

madison-ookla:
### Summary of Changes

This PR includes two changes:
1. Adds persistent Docker volumes for Neo4J and Elasticsearch so data in both services is preserved across restarts.
1. Renames the default compose file from `docker-amundsen.yml` to `docker-compose.yml`. This simplifies the commands used to run the default docker compose setup, since compose will automatically use an available `docker-compose.yml` file.

I sequestered both changes into separate commits, so if it's preferred to keep `docker-amundsen.yml` over `docker-compose.yml` I can jettison the second commit and we can retain the mounted volume changes in this PR.

I've also tested this locally to make sure that changes persist 😄 

### Documentation

Updated all references to `docker-amundsen.yml` to `docker-compose.yml` and simplified commands where possible.

### CheckList

- [X] PR title addresses the issue accurately and concisely, including a title prefix.

### DIFF
--- a/docker-amundsen.yml
+++ b/docker-amundsen.yml
@@ -16,6 +16,7 @@ services:
           - ./example/docker/neo4j/conf:/conf
           - ./example/docker/neo4j/plugins:/plugins
           - ./example/backup:/backup
+          - neo4j_data:/neo4j/data
       networks:
         - amundsennet
   elasticsearch:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -16,17 +16,16 @@ services:
           - ./example/docker/neo4j/conf:/conf
           - ./example/docker/neo4j/plugins:/plugins
           - ./example/backup:/backup
+          - neo4j_data:/neo4j/data
```
feng-tao, line 19: we have 3 sets of docker compose file for different purpose. I think let's keep the file name as it is for now and only apply the persistent volume change. WDYT?

madison-ookla: Totally fine with that! It was more of a convenience thing on my end in the first place 🙂  I'll drop that commit

feng-tao (approved):
thanks for the change

PR #706, opened 2020-09-20
fix: Update helm Neo4j deployment

brunsgaard:
### Summary of Changes

Use [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) for Neo4j deployment plugin volumes instead of [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath).
With this change we get rid of the dependency on hostPath and thus the dependency on the hosts filesystem and pods permissions to the hostPath.

The custom Neo4j deployment can now also run on GKE (and other k8s clusters).

I also sorted the config options for for Neo4j and disabled auth. The disabling of Neo4j auth seems to be a consistent setting through this repo (docker examples, and ECS examples).

### DIFF
--- a/amundsen-kube-helm/templates/helm/templates/configmap-neo4j.yaml
+++ b/amundsen-kube-helm/templates/helm/templates/configmap-neo4j.yaml
@@ -12,13 +12,14 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   neo4j.conf: |-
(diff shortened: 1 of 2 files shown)


feng-tao:
cc @javamonkey79

javamonkey79 (approved):
lgtm

feng-tao (approved):

PR #783, opened 2020-10-26
chore: Helm will keep persistent volume claims on delete

brunsgaard:
I suggest that helm delete should persist neo4j pvc's. Those should be deleted manually afterward to prevent dataloss.

### DIFF
--- a/amundsen-kube-helm/templates/helm/templates/pvc-neo4j.yaml
+++ b/amundsen-kube-helm/templates/helm/templates/pvc-neo4j.yaml
@@ -4,6 +4,8 @@ kind: PersistentVolumeClaim
 metadata:
   name: neo4j-pvc
   namespace: {{ .Release.Namespace }}
+  annotations:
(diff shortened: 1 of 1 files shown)


feng-tao (approved):

mgorsk1:
just to add my 2 cents, maybe it would make sense to just make neo4j statefulset instead of annotating the pvc. the result will be `almost` the same - when deleting neo4j pvc will not be deleted, but it won't get orphaned too. cc @feng-tao 

Following docs (https://helm.sh/docs/howto/charts_tips_and_tricks/):

`The annotation "helm.sh/resource-policy": keep instructs Helm to skip deleting this resource when a helm operation (such as helm uninstall, helm upgrade or helm rollback) would result in its deletion. However, this resource becomes orphaned. Helm will no longer manage it in any way. This can lead to problems if using helm install --replace on a release that has already been uninstalled, but has kept resources.`

PR #927, opened 2021-02-24
feat: Add imagePullSecrets to support amundsen private images

joeybaer:
### Summary of Changes

Added imagePullSecrets option to Amundsen images to support pulling from a private repository. 

### Documentation

Added references and example for using imagePullSecrets to values.yaml

### CheckList

- [X] PR title addresses the issue accurately and concisely, including a title prefix.

### DIFF
--- a/amundsen-kube-helm/README.md
+++ b/amundsen-kube-helm/README.md
@@ -27,8 +27,6 @@ This is setup templates for deploying [amundsen](https://github.com/amundsen-io/
 
 ## Chart Values
 
-The following table lists the configurable parameters of the Amundsen charts and their default values.
-
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | LONG_RANDOM_STRING | int | `1234` | A long random string. You should probably provide your own. This is needed for OIDC. |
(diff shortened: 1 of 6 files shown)


joeybaer:
@feng-tao I had two questions on this one.

1. In regards to the readme , I was unable to generate with `helm-docs` to match the current readme.md  against master prior to my changes. TLDR; missing the longer descriptions on keys. Are there certain flags I should be passing or a specific version I should be using?
1. Moving forward with this idea, do we think it would be a value add to rework the image details into a map which I have seen as a pattern. IE:
```
image:
  repository:
  tag:
  pullPolicy:
  pullSecrets: <have seen this separate and included in map>
```

Anyways the second is just to get your thoughts however the helm-docs issue is something I want to get fixed up prior to merge.

feng-tao:
I am actually not sure. tag @javamonkey79 who contributed the helm chart

javamonkey79:
@joeybaer 

> Are there certain flags I should be passing or a specific version I should be using?

I've just checked my command history, and it does not appear there are any flags on helm-docs that I ran. My version of helm-docs is 0.9.0 and my helm client version is 2.14.3.

> Moving forward with this idea, do we think it would be a value add to rework the image details into a map which I have seen as a pattern. IE:

Sure. I would advise taking a look at other open source charts and seeing how they do it, if you haven't already (it sounds like you have). If moving image details to a map adds value for you and\or makes the chart easier to use, I'd say go for it. 

Thanks for your contributions, thus far. I look forward to seeing the finished work.

joeybaer:
@javamonkey79 thanks for your input , as far as reworking the images to objects, I am going to leave that for now just because the value add isn't strong enough imo to warrant work for end users. 

I did however update the values.yaml to support the latest release of helm-docs which includes some updates to the readme outside of changes here just due to some drift in updates and the readme table being updated.

javamonkey79 (approved):
lgtm

feng-tao (approved):

PR #1015, opened 2021-04-27
feat: Add dbt extractor for table lineage + metadata

sewardgw:

### Summary of Changes

Added a dbt extractor that is able to extract the following:
- Tables / views + corresponding definitions
- Columns + corresponding definitions
- dbt tags (converted to Amundsen badges by default but configurable to ingest as Amundsen tags)
- Table lineage

A sample implementation has been created in the examples directory as well.

Implementers must provide both a dbt `catalog.json file` as well as a `manifest.json`. Both of these output files are available via the `dbt docs generate` CLI function. The `DbtExtractor` configs can accept the inputs as a string (that references a local file) as well as a `json.dumps(...)` string which can be loaded into a dictionary in order to allow implementers to reference dbt files that are stored locally as well as in the cloud.

### Documentation

Basic documentation has been added to the README and tests have been added for the new dbt extractor

### CheckList

- [ ] PR title addresses the issue accurately and concisely, including a title prefix.


databuilder/example/scripts/sample_dbt_loader.py
```diff
@@ -0,0 +1,178 @@
...
+
+    job = DefaultJob(conf=job_config,
+                     task=task,
+                     publisher=ElasticsearchPublisher())
+    return job
+
+
+if __name__ == "__main__":
+    # Uncomment next line to get INFO level logging
+    logging.basicConfig(level=logging.INFO)
```
dorianj, line 164: comment?

dorianj (approved):

sewardgw:
Updated with feedback from Convoy! Much appreciated @lexisantoro Jay W, Daniel D + Sam :D

databuilder/databuilder/extractor/dbt_extractor.py
```diff
@@ -196,43 +200,87 @@ def _default_sanitize(self, s: str) -> str:
...
+        return desc, desc_src
+
+    def _get_table_tags_badges(self, manifest_content: Dict) -> Tuple[Optional[List[str]], Optional[List[str]]]:
+        """
+        Gets tags or badges for a given table. At most one of these values will not be null.
+        """
+        tags, tbl_badges = None, None
+        if self._extract_tags:
+            if self._dbt_tag_as == DBT_TAG_AS.BADGE:
+                tbl_badges = manifest_content.get('tags')
```
feng-tao, line 220: why tbl_badges the same as tags ?

sewardgw: This was a request from some members of the community that I was collaborating with. The original implementation was to only use badges since the dbt tags (IMHO) behave more closely to Amundsen badges rather than Amundsen tags but after the feedback we made it configurable with the default being dbt tags are converted to Amundsen badges.

databuilder/README.md
```diff
@@ -468,6 +468,39 @@ job = DefaultJob(
...
+- Tables and their definitions
+- Columns and their definitions
+- Table level lineage
+- dbt tags (as Amundsen badges or tags)
+- Table Sources (e.g. link to GitHib where the dbt template resides)
+
+```python
+job_config = ConfigFactory.from_dict({
+    # Required args
+    f'extractor.dbt.{DbtExtractor.DATABASE_NAME}': 'snowflake',
```
feng-tao, line 485: if we use dbt for extraction, should the database be called dbt?

sewardgw: I think the database should be named the same database that Amundsen is also extracting metadata from so that metadata will be attached to existing tables / columns.

To give a practical example, if we use the Snowflake extractor then the value for the database is aptly called `snowflake`. The key for a given table would be something akin to `snowflake://cluster.schema/table`

In this use-case if we want to use dbt to help fill in the additional metadata and to populate table descriptions, lineage, etc. we will need to create the same key. If we pass in the value "dbt" here would yield something like `dbt://cluster.schema/table`

feng-tao: make sense, thanks for the info.

dorianj:
i fixed some CI stuff, you'll either need to rebase or merge with main for CI to pass

databuilder/example/scripts/sample_dbt_loader.py
sewardgw: commented this line

sewardgw:
> What about update [Table connectors](https://github.com/amundsen-io/amundsen#table-connectors) section in the README

Good call, will update this as well

feng-tao (approved):

apalmercari:
@sewardgw  Found issue in dbt extract lineage  
DBT manifest child map also have resource_type  'source' so below line gives key error as we are filtering only model types in [step](https://github.com/amundsen-io/amundsen/blob/4de3f9bb07d93b18b9ad91216cb0be6ce74e2a9d/databuilder/databuilder/extractor/dbt_extractor.py#L238) before that.

https://github.com/amundsen-io/amundsen/blob/4de3f9bb07d93b18b9ad91216cb0be6ce74e2a9d/databuilder/databuilder/extractor/dbt_extractor.py#L292

apalmercari:
Also why we included only dbt models in extraction, not dbt sources.

sewardgw:
Hey @apalmercari - we weren't using sources in the environment where this was originally implemented. Handling sources would be a great extension! Would you be able to submit a PR for this?

apalmercari:
Hey @sewardgw, ahh I understand. Sure,  I will try to.

PR #1037, opened 2021-05-03
fix:Ignore extra user fields

benrifkind:
Signed-off-by: benrifkind <ben.rifkind@gmail.com>

### Summary of Changes

This updates the `UserSchema` to ignore the extra fields. I am doing this because I am getting extra fields from the get_auth_user method (https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/oidc_config.py#L25) and this causes an error like:
```
 {'exp': ['Unknown field.'], 'sub': ['Unknown field.'], 'at_hash': ['Unknown field.'], 'iss': ['Unknown field.'], 'amr': ['Unknown field.'], 'idp': ['Unknown field.'], 'auth_time': ['Unknown field.'], 'jti': ['Unknown field.'], 'ver': ['Unknown field.'], 'aud': ['Unknown field.'], 'iat': ['Unknown field.']}
2021-05-03T16:34:17+0000.056 [ERROR] v0.current_user:48 (11:MainThread) - Encountered exception: 'dict' object has no attribute 'user_id'
```
I believe this is because the method returns more fields than the `UserSchema` expects. 

I am unsure of how to add tests to this because the tests seem like they belong in https://github.com/amundsen-io/amundsen/blob/main/frontend/tests/unit/models/test_user.py but I am not sure how the tests link with an update to the common module.
This is the test I want to add
```
    def test_extra_key_does_not_raise(self) -> None:
        """
        Handle extra keys in the user data
        :return:
        """
        test_user = {
            'email': 'test@test.com',
            'foo': 'bar',
        }
        with app.test_request_context():
            self.assertEqual(dump_user(load_user(test_user)).get('email'), 'test@test.com')
```

### CheckList

- [x] PR title addresses the issue accurately and concisely, including a title prefix.

### DIFF
--- a/common/amundsen_common/models/user.py
+++ b/common/amundsen_common/models/user.py
@@ -4,7 +4,7 @@
 from typing import Any, Optional, Dict
 
 import attr
-from marshmallow import ValidationError, validates_schema, pre_load
+from marshmallow import EXCLUDE, ValidationError, validates_schema, pre_load
 from marshmallow3_annotations.ext.attrs import AttrsSchema
 
 """
@@ -46,6 +46,7 @@ class UserSchema(AttrsSchema):
     class Meta:
         target = User
         register_as_scheme = True
+        unknown = EXCLUDE
 
     # noinspection PyMethodMayBeStatic
     def _str_no_value(self, s: Optional[str]) -> bool:
--- a/common/tests/unit/models/test_user.py
+++ b/common/tests/unit/models/test_user.py
@@ -89,3 +89,12 @@ class UserTest(unittest.TestCase):
         :return:
         """
         self.assertEqual(UserSchema()._str_no_value(" "), True)
+
+    def test_extra_key_does_not_raise(self) -> None:
+        """
+        Handle extra keys in the user data
+        :return:
+        """
+        test_user = {"email": "test@test.com", "foo": "bar"}
+        with app.test_request_context():
+            self.assertEqual(UserSchema().load(test_user).email, "test@test.com")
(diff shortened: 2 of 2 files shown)


sewardgw:
I think the unit tests for this should go into the commons test directory: https://github.com/amundsen-io/amundsen/tree/main/common/tests It doesn't look like there is a directory for testing models but there probably should be and IMHO it should handle all of the ser/de for the models.

The ones in the frontend service are really testing that the dependency works, is backwards compatible, etc. It should be considered more of an integration test than a unit test.

benrifkind:
I added two commits. 

The first ports all the tests from https://github.com/amundsen-io/amundsen/blob/main/frontend/tests/unit/models/test_user.py to this repo since to my mind they seem to be testing the functionality of the model more than being integration tests.  I can also remove them if you think this is duplicating tests. 

The second commit adds the test for this new behavior I added.

sewardgw:
Porting makes a ton of sense, especially w/ the new mono repo. You are keeping the existing tests in place for the frontend because those are pegged to a specific version of the common lib?

benrifkind:
I left those tests because you mentioned they were used as integration tests. I am happy to remove them. I would remove all the tests except the one for profile id since that is the one that tests extra functionality in the frontend.

benrifkind:
@sewardgw gentle reminder 🙏

sewardgw (approved):
Leaving the tests in frontend for now makes sense, once the build / test processes are updated to use mono repo common instead of pypi common they can be removed.

feng-tao (approved):

PR #1041, opened 2021-05-04
fix: Database filter to return string, not object

sewardgw:

### Summary of Changes

Previously, when selecting the database from the table details page the value placed into the `source` field was an object, not a string. This PR updates the value being passed to be the actual value of the database.

Previous:

Screen Shot 2021-04-23 at 10 30 32 AM

Current:

Screen Shot 2021-04-23 at 10 30 39 AM

This is inline with how other filters are being passed, for example, dashboard and table tags: https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/static/js/components/Tags/TagInfo/index.tsx#L63

### Documentation

N/A

### CheckList

- [ ] PR title addresses the issue accurately and concisely, including a title prefix.

### DIFF
--- a/frontend/amundsen_application/static/js/pages/TableDetailPage/TableHeaderBullets/index.tsx
+++ b/frontend/amundsen_application/static/js/pages/TableDetailPage/TableHeaderBullets/index.tsx
@@ -68,7 +68,7 @@ export const mapDispatchToProps = (dispatch: any) =>
       searchDatabase: (databaseText: string) =>
         updateSearchState({
           filters: {
-            [ResourceType.table]: { database: databaseText }
+            [ResourceType.table]: { database: databaseText },
           },
           submitSearch: true,
         }),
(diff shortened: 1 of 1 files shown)


Golodhros (approved):
LGTM

markgrover:
Awesome, thanks! Out of scope of this PR, but it'd be great to make this is a dropdown in the future because most orgs only have a handful of these sources (like Bigquery, Redshift, etc.) and we should get a list and show them. We'd also need a default all in that case.

Golodhros:
> Awesome, thanks! Out of scope of this PR, but it'd be great to make this is a dropdown in the future because most orgs only have a handful of these sources (like Bigquery, Redshift, etc.) and we should get a list and show them. We'd also need a default all in that case.

There is new stuff coming soon for this, they are still figuring out the proposal, but stay tuned!

dorianj:
Test failure is real (lint), but this LGTM

PR #1047, opened 2021-05-05
test: update Mode dashboard extractor tests

youngyjd:
### Summary of Changes

Update some Mode dashboard extractor tests which will be more like an end-to-end test and increase test coverage a little bit

### Tests

As summary says

### CheckList


feng-tao (approved):
@youngyjd @allisonsuarez do you want to bump a new version for databuilder?

youngyjd:
@feng-tao will bump version later. Will have a few more PRs. We switched to discovery to get spaces https://github.com/amundsen-io/amundsen/commit/04f1e27a3f337fb16af24ded4b8074ee5fe24543. However, discovery api are authenticated with bearer token, while some following calls to main api, which is authenticated by a user token and password, to get reports under a space is broken because the user does not have access to that space. Space discovery api returns all spaces including restricted ones, while the user calling Mode main api does not have access to some of them.

PR #1056, opened 2021-05-06
fix: fix bug in rest_api_query

youngyjd:
### Summary of Changes

Potential bugs in rest api query:
The yield `record_dict` for example `{'product': 'mode'}` may be transformed in an extractor to a dict like `{'product': 'mode', 'new_field': 'new_field'}`. When the next yield record is extracted, it still should be `{'product': 'mode'}` instead of `{'product': 'mode', 'new_field': 'new_field'}` since two records should be unrelevant.

See this extractor as an example that changes record_dict https://github.com/amundsen-io/amundsen/blob/1cece5b4461c8623414ccfb635e495e794a55988/databuilder/databuilder/extractor/dashboard/mode_analytics/mode_dashboard_queries_extractor.py#L74

This is not causing a problem on our end right now. It is a proactive change to avoid possible problems. 

### CheckList

### DIFF
--- a/databuilder/databuilder/rest_api/rest_api_query.py
+++ b/databuilder/databuilder/rest_api/rest_api_query.py
@@ -177,10 +177,10 @@ class RestApiQuery(BaseRestApiQuery):
                     if not sub_record or len(sub_record) != len(self._field_names):
                         # skip the record
                         continue
-                    record_dict = copy.deepcopy(record_dict)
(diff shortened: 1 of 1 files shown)


youngyjd:
@allisonsuarez @feng-tao @dorianj

feng-tao (approved):

PR #1059, opened 2021-05-06
feat: ML Features common schema and metadata API definition

allisonsuarez:

### Summary of Changes

Kicking off work for https://github.com/amundsen-io/rfcs/pull/36
Feature APIs just return null right now
- Added feature schemas to Amundsen common
- Moved badges and tags out of table model in common
- Updated CI check to install common from code before running
- Created API stubs for all required requests
- Note: made separate stats, generation code, and sample APIs similar to lineage in order to handle tabs more gracefully in the UI especially when no data is available fro any of these.

### CheckList


metadata/metadata_service/api/feature.py
```diff
@@ -0,0 +1,131 @@
...
+from metadata_service.proxy import get_proxy_client
+
+
+class FeatureDetailAPI(Resource):
+    """
+    FeatureDetail API
+    """
+
+    def __init__(self) -> None:
+        self.client = get_proxy_client()
```
sewardgw, line 17: Is it worth having a base resource to reduce the copy/paste of the `init`s in this file? For example:

```python
class MyResource(Resource):
    def __init__(self) -> None:
        self.client = get_proxy_client()
...
class FeatureDetailAPI(MyResource):
...
```

sewardgw (commented):
overall looks great, main question is really around the scope of a "query" object since that can be a loaded term in reference to data.

common/amundsen_common/models/feature.py
```diff
@@ -0,0 +1,97 @@
...
+
+from amundsen_common.models.user import User
+from amundsen_common.models.badge import Badge
+from amundsen_common.models.tag import Tag
+from amundsen_common.models.table import Column, ProgrammaticDescription, Watermark
+from marshmallow3_annotations.ext.attrs import AttrsSchema
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class Query:
```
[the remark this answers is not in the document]
allisonsuarez: more generic, I tried to follow what we have for dashboards already because we are reusing the same component to show this code snippet or query in frontend

metadata/metadata_service/api/feature.py
allisonsuarez: This is a good idea, I wonder if we should take that abstraction even higher up so Tables and other entities can reuse it as well. Might be a bit out fo the scope of this PR to make the change for all entities.

common/amundsen_common/models/feature.py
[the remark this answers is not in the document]
allisonsuarez: we could rename this something more generic like code snippet or generation code so it goes with the common schema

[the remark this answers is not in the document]
sewardgw: I just tagged you in the RFC thread here: https://github.com/amundsen-io/rfcs/pull/37/files#r632790084

Do you mind if we have the conversation on that thread and provide a synopsis here once resolved?

metadata/metadata_service/api/feature.py
allisonsuarez: I think we should do this as a separate PR @sewardgw so we can apply it to all entities, what do you think?

sewardgw: 👍🏻  sounds good!

sewardgw (approved):
lgtm w/ Query model update!

common/amundsen_common/models/feature.py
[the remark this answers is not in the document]
sewardgw: For reference, this was discussed and resolved in another PR: https://github.com/amundsen-io/rfcs/pull/37#discussion_r633545054

Query was then moved to separate file.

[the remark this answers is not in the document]
allisonsuarez: TLDR: made query file with generic and simple query entity that can be expanded on with query and query composition work

common/amundsen_common/models/feature.py
```diff
@@ -0,0 +1,84 @@
...
+
+from amundsen_common.models.user import User
+from amundsen_common.models.badge import Badge
+from amundsen_common.models.tag import Tag
+from amundsen_common.models.table import Column, ProgrammaticDescription, Watermark
+from marshmallow3_annotations.ext.attrs import AttrsSchema
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class ColumnItem:
```
mgorsk1, line 16: could we reuse `table.Column` class or it doesn't make sense?

metadata/metadata_service/api/feature.py
```diff
@@ -0,0 +1,131 @@
...
+        pass
+
+
+class FeatureSampleAPI(Resource):
+
+    # TODO use DataSample common model
+    def __init__(self) -> None:
+        self.client = get_proxy_client()
+
+    @swag_from('swagger_doc/feature/detail_get.yml')
```
mgorsk1, line 62: FeatureStatsAPI, FeatureGenerationCodeAPI, FeatureSampleAPI have the same file in `@swag_from` - is it correct?

common/amundsen_common/models/feature.py
allisonsuarez: this is for the sample response rather than for the partition Column. So basically just the model for the data we need to display a handfull of rows of data for a feature. Notice that partition column does reuse the table column

mgorsk1 (approved):
👍

PR #1079, opened 2021-05-11
deploy_release: on release, publish to pypi/docker

dorianj:
Given a release with a title of `<module_name> <version_number>`, this will push that release to PyPi and docker.

Note that this does not attempt to create the changelog or any of that stuff. That will need to be done by a separate action. This creates a nice modularity boundary, in that this action doesn't care where the release came from (manual or another automated step)


feng-tao:
cc @youngyjd

youngyjd:
Thank you! @dorianj

Issue #1104, opened 2021-05-17
Feature Proposal: Implement Pandas Profiling column statistics extractor


mgorsk1:
Pandas profiling (https://github.com/pandas-profiling/pandas-profiling) is a tool widely used for calculating reports on tables. It returns parsable json reports with quantile/descriptive statistics on column level. This PR proposes to develop `PandasProfilingColumnStatsExtractor` to populate `TableColumnStats` entities.

## Expected Behavior or Use Case

- new extractor exists parsing pandas-profiling reports to TableColumnStats

## Service or Ingestion ETL

- databuilder

## Possible Implementation

- rely on pandas profiling json report to collect metadata

mgorsk1:
cc @sbrugman

PR #1105, opened 2021-05-17
feat: Pandas-Profiling TableColumnStats Extractor

mgorsk1:
### Summary of Changes

Follows-up on https://github.com/amundsen-io/amundsen/issues/1104 and introduces pandas-profiling based column stats extractor.

### Tests

- added `TestPandasProfilingColumnStatsExtractor`

### Documentation

- extended `README.md` with new extractor description

### CheckList


mgorsk1:
cc @sbrugman

databuilder/README.md
```diff
@@ -1172,6 +1172,103 @@ job = DefaultJob(conf=job_config,
...
+```python
+PandasProfilingColumnStatsExtractor.STAT_MAPPINGS = {'max': ('Maximum', float), 'min': ('Minimum', float)}
+```
+
+Complete set of available metrics is defined as DEFAULT_STAT_MAPPINGS attribute of PandasProfilingColumnStatsExtractor.
+
+#### Common usage patterns
+
+As pandas profiling is executed on top of pandas dataframe, it is up to the user to populate the dataframe before running 
+the report calculation (and subsequently the extractor). While doing so remember that it might not be a good idea to run t
```
dorianj, line 1209: the cut in wordwrap?

databuilder/databuilder/extractor/pandas_profiling_column_stats_extractor.py
```diff
@@ -0,0 +1,191 @@
...
+        report = self._load_report()
+
+        variables = report.get('variables', dict())
+        report_time = self.parse_date(report.get('analysis', dict()).get('date_start'))
+
+        for column_name, column_stats in variables.items():
+            for _stat_name, stat_value in column_stats.items():
+                stat_spec = self.stat_mappings.get(_stat_name)
+
+                if stat_spec:
```
dorianj, line 131: when would `stat_spec` be None? I think it might be nice if we threw if the config was incorrectly typed, just to make the failure happen more clearly/faster

dorianj (approved):
This is dope!

databuilder/databuilder/extractor/pandas_profiling_column_stats_extractor.py
mgorsk1: the idea is that users might config this publisher with subset of `DEFAULT_STAT_MAPPINGS` - for instance when only selected metrics available in pandas-profiling are meant to make way to amundsen (in total it's around 31 different metrics for numerical values and 17 for text values - it might be too much for some users). in such case stat_spec will be None - _stat_name is present in pandas-profiling but not desired by user to be extracted.

databuilder/README.md
mgorsk1: good catch, thanks!

databuilder/databuilder/extractor/pandas_profiling_column_stats_extractor.py
mgorsk1: The idea is that users might want to extract only subset of metrics available through pandas-profiling (and defined within `DEFAULT_STAT_MAPPINGS`). if users don't want to extract the kurtosis, although it's available in the report file, it would be absent in `self.stat_mappings` and hence - `stat_spec` would be `None`. So it wouldn't be consider a failure but more like filtering out. In total there are 31 unique metrics for numerical values and 17 for text - it could be too much for some users.

dorianj: ohh, I'm sorry, I read this code too fast and though it was iterating through the stat_mappings, not the source mappings. totally makes sense.

README.md
```diff
@@ -132,20 +132,21 @@ Please visit [Architecture](./docs/architecture.md#architecture) for Amundsen ar
...
+### Table Column Statistics
+
+- [Pandas Profiling](https://pandas-profiling.github.io/pandas-profiling/docs/master/rtd/)
+
 ### Dashboard Connectors
 
 - [Mode Analytics](https://mode.com/)
 - [Redash](https://redash.io/)
 - [Tableau](https://tableau.com/)
+- [Apache Superset](https://superset.apache.org/)
```
feng-tao, line 144: Make it sort?

feng-tao (approved):

README.md
mgorsk1: done

mgorsk1:
thanks @dorianj @feng-tao 🤝

PR #1108, opened 2021-05-17
feat: Implement Feature Tag, Badge, Description and Owner APIs

allisonsuarez:
### Summary of Changes

- Implemented FeatureTagAPI, FeatureBadgeAPI, FeatureDescriptionAPI and FeatureOwnerAPIs
- Changed neo4j proxy behavior to reuse more general methods for owners

### Tests

- Added unit tests for all APIs
- Updated one test based on new behavior for neo4j
- Also manually tested all changes and new behavior

### CheckList


metadata/metadata_service/api/feature.py
```diff
@@ -83,13 +86,25 @@ class FeatureOwnerAPI(Resource):
...
-    def delete(self, table_uri: str, owner: str) -> Iterable[Union[Mapping, int, None]]:
-        pass
+    @swag_from('swagger_doc/feature/owner_put.yml')
+    def put(self, feature_uri: str, owner: str) -> Iterable[Union[Mapping, int, None]]:
+        try:
+            self.client.add_resource_owner(uri=feature_uri, resource_type=ResourceType.Feature, owner=owner)
+            return {'message': f'The owner {owner} for feature_uri {feature_uri} '
+                               'was added successfully'}, HTTPStatus.OK
+        except Exception:
+            return {'message': f'The owner {owner} for feature_uri {feature_uri} was '
```
dkunitsk, line 96: observation: bit odd there's no logging and err details are not returned. I know there also isn't for table owners.

metadata/metadata_service/proxy/atlas_proxy.py
```diff
@@ -1456,3 +1456,26 @@ def get_lineage(self, *,
 
     def get_feature(self, *, feature_uri: str) -> Feature:
         pass
+
+    def get_resource_description(self, *,
```
dkunitsk, line 1460: I wonder what others think of the generality here. It raises the obvious future question of whether to do `get_resource_description` with Table as `resource_type` or use `get_table_description`.

Keeping with the per-entity boilerplate wagon seems more straightforward, perhaps.

metadata/metadata_service/api/feature.py
sewardgw: +1

metadata/metadata_service/proxy/atlas_proxy.py
sewardgw: One of the benefits we get from using the generic `get_resource_description` is that we can pretty much reuse all of the same code for any resource types with the one caveat being that they must all follow a similar data model within the DB. For instance, tables, dashboards, features, etc. must all have a "Description" and use the same relationship label. This way the # of one-off functions to support N number of resource types stays at 1 per action (e.g. get description, delete desc, etc.)

allisonsuarez: I don't see why we would ever change this relationship for future entities so I think we should work towards making these methods general wherever we can.

metadata/metadata_service/api/feature.py
allisonsuarez: I'll add some!

dkunitsk (approved):
Great!

metadata/metadata_service/api/feature.py
allisonsuarez: done

metadata/metadata_service/proxy/atlas_proxy.py
verdan: +1 to have these functions generic enough to support all kinds of entities/resources, and the implementation can be different based on the `resource_type`

metadata/metadata_service/proxy/neo4j_proxy.py
```diff
@@ -1452,7 +1485,7 @@ def get_dashboard_description(self, *,
         :return:
         """
 
-        return self._get_resource_description(resource_type=ResourceType.Dashboard, uri=id)
+        return self.get_resource_description(resource_type=ResourceType.Dashboard, uri=id)
 
     @timer_with_counter
     def put_dashboard_description(self, *,
```
verdan, line 1491: How about moving calling `put_resource_description` directly from API with the `resource_type=ResourceType.Dashboard`?

verdan (approved):
just a few comments, the rest looks good to me.

metadata/metadata_service/proxy/neo4j_proxy.py
allisonsuarez: This would be backwards incompatible so I decided to not do those changes in this PR

Issue #1071, opened 2021-05-11
Make BigQuery Usage Extractor Support Relational Metadata Store


sahithi03:
Currently usage extractor needs a couple of changes in order to use it with a relational metadata store:
1. GCP console allows running queries using tables from a project different from the one the extractor is used for; e.g We can query tables of Project A from the console of Project B. This will result in "[referenced tables](https://cloud.google.com/logging/docs/reference/audit/bigquery/rest/Shared.Types/AuditData#JobStatistics)" containing tables from a different project. If we use Amundsen with a specific env(google cloud project), the backend relational metadata store will not store tables from a different project leading to a FK constraint violation  between table_usage and table_metadata tables.

2. The bigquery usage extractor job will run after bigquery metadata extractor job because of the foreign key dependency between table_usage and table_metadata tables, however if new tables are created and queried while BQ Metadata extractor job is still running, the logging API response might return a few referenced tables that don’t exist yet in table_metadata tables leading to a FK constraint violation.

3. Whenever we run queries on tables that have table decorators the referenced table name has the decorator included in it.  
e.g `SELECT COUNT(*) FROM [PROJECT_ID:DATASET.TABLE@-3600000]` the referenced tables returned by this query will be TABLE@- 3600000, since this is not originally stored in table_metadata table, this will again result in FK constraint violation

4. Similarly when date sharded tables are queried (Table_20210503), the "referenced tables" contains the table_name along with the suffix (Table_20210503) but since the [metadata extractor](https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/extractor/bigquery_metadata_extractor.py#L50&L51) does not extract metadata of each sharded table, this will lead to a FK constraint violation. 
5. When temporary/ anonymous tables are queried their usage count is also recorded.
6. When we have metadata queries like `select * from my-project.my-dataset.INFORMATION_SCHEMA.TABLES;` or `select * from my-project.my-dataset.__TABLES__ ` the "referenced tables" contains INFORMATION_SCHEMA or `__TABLES__ `as the table referred by the query job.
## Expected Behavior or Use Case

BigQuery usage extractor should only extract usage for tables present in the table_metadata table.
## Service or Ingestion ETL

Databuilder extractors
## Possible Implementation

Possible solutions for each of the issues mentioned above: 
[...]

Issue #1070, opened 2021-05-10
Make BigQuery Watermark Extractor support Relational Metadata store


sahithi03:
When we have a relational DB as metadata store we would have to run bigquery metadata extractor job followed by watermark extractor job in order to follow the foreign key dependency between table_metadata and table_watermark tables.
However, while bigquery metadata extractor job is running, if new tables are added in the google cloud project that is configured with the extractors, then the watermark extractor might extract watermark for tables not yet existing in the table_metadata table, which will lead to a FK constraint violation.

## Expected Behavior or Use Case

Watermark extractor should only extract metadata for tables created before the execution-time of the bigquery metadata extractor in order to avoid fetching tables that might lead to a FK constraint violation.
## Service or Ingestion ETL

Databuilder Extractors
## Possible Implementation

One possible solution is to have a Config cutoff-time which can be set to the actual execution-time of the bigquery metadata task. If not configured, the default for the cutoff-time config can be current time.
The watermark extractor can check if the table creation time (https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table) is less than the cutoff-time and only extract watermark metadata for those tables.
## Context
We are using BigQuery as the DataWarehouse and MySQL as the metadata store.

crazy-2020:
Thanks, I do not have any concern. The cutoff-time would work for both graph db and mysql users and its default value can work for the main graph db users.

feng-tao:
Feel free to create to pr to fix it! thanks

feng-tao:
make sense, FYI, for graph db, if the table only appears after the extractor has run, the watermark node will be a stale node which won't affect the UI. But I could see it could be an issue for rdbms case.

PR #1109, opened 2021-05-18
Fix: Make BigQuery extractors support Relational Metadata store

sahithi03:
### Summary of Changes

Changes made according to the issues: #1070 & #1071
Additional changes made not mentioned in the issue: 
- Skip usage counts for wildcard tables: When there is a query like `select * from project:dataset:tableprefix*`, the `'referenced tables'` in the logging response contains the tableId as tableprefix* resulting in a FK constraint violation between table_usage and table_metadata tables.

### Tests

Added unit tests for the new configs in bigquery_usage_extractor and bigquery_watermark_extractor. 
### Documentation

N/A
### CheckList

### DIFF
--- a/databuilder/databuilder/extractor/bigquery_usage_extractor.py
+++ b/databuilder/databuilder/extractor/bigquery_usage_extractor.py
@@ -4,7 +4,9 @@
 import logging
 import re
 from collections import namedtuple
-from datetime import date, timedelta
+from datetime import (
+    datetime, timedelta, timezone,
+)
 from time import sleep
 from typing import (
     Any, Dict, Iterator, List, Optional, Tuple,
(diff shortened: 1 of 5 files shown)


crazy-2020:
LGTM, we'll see if Amundsen community has any concern.

databuilder/databuilder/extractor/bigquery_usage_extractor.py
```diff
@@ -99,29 +109,54 @@ def _create_records(self, refResources: List[dict], resourcesProcessed: int, ema
...
+                                                schema=datasetId,
+                                                table=tableId,
+                                                column='*',
+                                                email=email)
+                else:
+                    LOGGER.debug(f'Not counting usage for {refResource} since '
+                                 f'{BigQueryTableUsageExtractor.COUNT_TABLES_ONLY_FROM_PROJECT_ID_KEY} is True')
+                    continue
+            else:
+                key = TableColumnUsageTuple(database='bigquery',
```
feng-tao, line 136: I wonder for other non rdbms store (graph db), we should also ignore the tables that are not from the same project?

feng-tao (approved):

databuilder/databuilder/extractor/bigquery_usage_extractor.py
sahithi03: Yes, we could ignore usage counts for tables from other projects even in the case of graph db, but made it configurable in case graph db users want to get the usage of all the referenced tables from their queries, even if they are not from the same project. 
Should I make ‘count_tables_only_from_project_Id_key’ the default behavior?

feng-tao: yeap, I think `count_tables_only_from_project_Id_key` as default makes sense

sahithi03: Updated the PR!

PR #1114, opened 2021-05-18
test: Add Python 3.8 and 3.9 databuilder CI

Fokko:
### Summary of Changes

Ran the databuilder locally using 3.9 and it works fine, would be a pity to not add this as a supported version

### Tests

Added the versions to the `setup.py` and the CI itself.

### CheckList

### DIFF
--- a/.github/workflows/databuilder_pull_request.yml
+++ b/.github/workflows/databuilder_pull_request.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        python-version: ['3.6.x', '3.7.x']
+        python-version: ['3.6.x', '3.7.x', '3.8.x', '3.9.x']
     steps:
       - name: Checkout
         uses: actions/checkout@v1
--- a/.github/workflows/noop_pull_request.yml
+++ b/.github/workflows/noop_pull_request.yml
@@ -30,7 +30,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.6.x", "3.7.x", "3.8.x"]
+        python-version: ["3.6.x", "3.7.x", "3.8.x", "3.9.x"]
     steps:
       - name: Do nothing
         run: exit 0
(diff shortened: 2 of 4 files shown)


dorianj:
LGTM 

Could you also add 3.9 to https://github.com/amundsen-io/amundsen/blob/main/.github/workflows/noop_pull_request.yml ? thanks!

dorianj:
After this lands (including that noop change), we will enforce the check on our side.

Also, are there any places in the docs where this should change?

Fokko:
@dorianj Thanks for looking into this. I've checked the docs, and could not find anything that should be changed. I've added the 3.9 to the noop workflow as well.

dorianj (approved):
Thanks for the submission!!

dorianj:
also, if you're feeling like it, would super appreciate a PR to make the front-end service work with newer versions of Python. there are some real failures there (that the tests catch!), though, so won't be quite as trivial

dorianj:
Uh oh, looks like the build failure is real

Fokko:
@dorianj Happy to bump the front end as well. But let me first dig into this error, looks like something real indeed.

Fokko:
I'm unable to reproduce this locally or in an Ubuntu docker container. I think it has to do with the sequencing. `numpy` is installed in the same run, as it where tries to bind with the headers.

dorianj:
You're welcome to hit me up on the slack!

There might be a discrepancy between the stuff that's in requirements, and that which is in setup.py which might be causing problems

Fokko:
There is an issue with `pyarrow`, which is being brought in by `feast`, which is pinned at `0.8.0`: https://github.com/amundsen-io/amundsen/blob/main/databuilder/setup.py#L76. This pulls in `pyarrow<0.16.0,>=0.15.1`: https://github.com/feast-dev/feast/blob/v0.8.0/sdk/python/setup.py#L45. Which does not support Python 3.8+: https://github.com/apache/arrow/issues/5987

I can bump feast in another PR, to keep the PR's nice and small. WDYT?

PR #1123, opened 2021-05-19
chore: make it a bit friendlier

dorianj:
Given that this doesn't actually check if people filled out the issue template, I think it's a bit unfriendly. I also don't think it will change behavior for people who don't follow the template

Signed-off-by: Dorian Johnson <2020@dorianj.net>

### CheckList


mgorsk1 (approved):
To be honest this setting was same as Airflow config. Now we can claim to be more friendly community ;)

PR #1125, opened 2021-05-19
docs: fix some broken links

dorianj:
I'm not sure when or how these were broken, but the URLs are cleaner now


Golodhros (approved):

markgrover:
I am not familiar with the new syntax that doesn't refer to the name of the file but these links seem to broken now for me when browsing the README on the web. It seems like for other people too: https://amundsenworkspace.slack.com/archives/CGFBVT23V/p1622019021119500

PR #1146, opened 2021-05-21
feat: Added Flask hooks for pre/post event processing

sewardgw:

### Summary of Changes

Created optional hooks for Flask pre / post request processing. These hooks accept an arbitrary function (given that the function matches the signature required for the Flask hook) and allows the implementer to have more control over the pre and post processing of events. Some use-cases for this include:

- Custom auth validation
- Streaming user / resource usage to an external system for monitoring
- Overriding default responses (e.g. GET user returns does not exist, then retrieve user information from 3rd party system then create user)
- Checking database connections before starting app / closing external connections on shut down

### Tests

Added tests for base case where no hooks are created as well as test for when hook is added

### Documentation

Updated configuration docs.

### DIFF
--- a/metadata/docs/configurations.md
+++ b/metadata/docs/configurations.md
@@ -4,7 +4,7 @@ Most of the configurations are set through Flask [Config Class](./../metadata_se
 In order to add a badge to a resource you should first add the combination of badge name and category to the in WHITELIST_BADGES [Config Class](./../metadata_service/config.py).
 
 Example:
-```python 
+```python
 
 WHITELIST_BADGES: List[Badge] = [Badge(badge_name='beta',
                                  category='table_status')]
@@ -16,7 +16,7 @@ Once this is done users will be able to add badge the badges in the whitelist by
 
 #### USER_DETAIL_METHOD `OPTIONAL`
 This is a method that can be used to get the user details from any third-party or custom system.
-This custom function takes user_id as a parameter, and returns a dictionary consisting user details' fields defined in [UserSchema](./../../common/amundsen_common/models/user.py). 
+This custom function takes user_id as a parameter, and returns a dictionary consisting user details' fields defined in [UserSchema](./../../common/amundsen_common/models/user.py).
 
 Example:
 ```python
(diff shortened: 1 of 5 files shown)


feng-tao:
we have a hook interface implementation in frontend (https://www.amundsen.io/amundsen/tutorials/how-to-track-user-metric/) , any reason we add another hook implementation in the backend?

feng-tao (changes requested):
need a RFC as 1) increased numbers of config in the metadata service ; 2) is there a security concern over injecting arbitrary field into header; 3) what is the usecase that is not covered through frontend message event hook

metadata/docs/configurations.md
```diff
@@ -34,6 +34,31 @@ def get_user_details(user_id):
...
+from flask import request
+
+def verify_headers():
+    my_auth = request.headers.get('Authorization')
+    print(my_auth)
+
+
+def streap_resp(resp):
+    resp_data = resp.json
+    stream_data_to_kafka(resp_data)
```
feng-tao, line 53: we have the event action logging framework used at Lyft since 3 years ago (https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/log/action_log.py) . Any reason we need another hook in the metadata layer?

sewardgw: The specific implementation for this is to be able to use neo4j as a "cache" of sorts for user information. First, call the GET User API and, if the response is a 404 then go and reach out to the OIDC provider to get the user information. In addition, we check other attributes (e.g. has it been more than 30 days since we got the user information from the OIDC provider) and determine if we need to get current information from the OIDC provider again.

The examples provided were just intended to be ways that this could be used. While that particular example does overlap heavily with the action logger, this approach allows for more interaction with the Flask objects pre and post execution.

sewardgw:
If an RFC is required in order to add new configs I will open that up, but I am not sure I'm clear on where you see the security concern. Injecting a header is one example of something that _could_ be done but there are many other use-cases that this enables beyond setting a header. Implementers are already able to add their own headers on the fly if they want when they stand up their Amundsen implementation, should they decide to use this functionality it would make it easier, but it's not a declarative approach that implementers must inject a header.

In regards to your point #3 about additional use-cases, see the comment above: https://github.com/amundsen-io/amundsen/pull/1146#discussion_r637192606

Given this, would you say this still requires an RFC?

metadata/docs/configurations.md
feng-tao: I could be wrong, but I think this use case is a bit specific (caching user information other than what we have in user model)  which I am not sure how open source is used in general. Could this feature be done through private fork?

feng-tao: e.g lots of company (including Lyft) has private proxy on top of open source proxy to add some custom changes.

feng-tao:
> If an RFC is required in order to add new configs I will open that up, but I am not sure I'm clear on where you see the security concern. Injecting a header is one example of something that _could_ be done but there are many other use-cases that this enables beyond setting a header. Implementers are already able to add their own headers on the fly if they want when they stand up their Amundsen implementation, should they decide to use this functionality it would make it easier, but it's not a declarative approach that implementers must inject a header.
> 
> In regards to your point #3 about additional use-cases, see the comment above: [#1146 (comment)](https://github.com/amundsen-io/amundsen/pull/1146#discussion_r637192606)
> 
> Given this, would you say this still requires an RFC?

commented

metadata/docs/configurations.md
sewardgw: Yea, I think that's fair. I'll close this out. Should someone else happen to broach this how-to in the future we can continue the topic there.

feng-tao: thanks

PR #1147, opened 2021-05-21
feat: FE ML Feature get_feature_metadata

allisonsuarez:

### CheckList

### DIFF
--- a/frontend/amundsen_application/api/metadata/v0.py
+++ b/frontend/amundsen_application/api/metadata/v0.py
@@ -883,7 +883,7 @@ def _get_feature_metadata(*, feature_key: str, index: int, source: str) -> Dict[
 
     try:
         feature_endpoint = _get_feature_endpoint()
-        url = '{0}/{1}'.format(feature_endpoint, feature_key)
+        url = f'{feature_endpoint}/{feature_key}'
         response = request_metadata(url=url)
     except ValueError as e:
         # envoy client BadResponse is a subclass of ValueError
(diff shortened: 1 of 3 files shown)


frontend/amundsen_application/api/metadata/v0.py
```diff
@@ -842,3 +851,69 @@ def get_column_lineage() -> Response:
...
+def _get_feature_metadata(*, feature_key: str, index: int, source: str) -> Dict[str, Any]:
+
+    results_dict = {
+        'featureData': {},
+        'msg': '',
+    }
+
+    try:
+        feature_endpoint = _get_feature_endpoint()
+        url = '{0}/{1}'.format(feature_endpoint, feature_key)
```
danwom, line 886: nit: use interpolated string

`f'{feature_endpoint}/{feature_key}'`

frontend/amundsen_application/api/metadata/v0.py
```diff
@@ -842,3 +851,69 @@ def get_column_lineage() -> Response:
...
+    if status_code != HTTPStatus.OK:
+        message = 'Encountered error: Metadata request failed'
+        results_dict['msg'] = message
+        logging.error(message)
+        return results_dict
+
+    try:
+        feature_data_raw: dict = response.json()
+
+        feature_data_raw['key'] = feature_key
```
danwom, line 908: Adding the 'key' and the 'marshall' functions in the frontend were added because FE needed some things that the metadata or search service didn't include. I talked about this with @feng-tao a while ago. What do you think about adding these types of fields to the upstream responses?


(Not necessarily asking to change these here, just starting a discussion) 
(Also wondering if FE should just call search/metadata services directly instead of having a FE Flask layer)

danwom (approved):

frontend/amundsen_application/api/metadata/v0.py
allisonsuarez: Honestly I have been thinking the same thing, its all basically just duplicated work which seems unnecessary.

PR #1155, opened 2021-05-25
feat: Common query models

sewardgw:
### Summary of Changes

This change introduces the common models that will be used by the frontend and metadata services per the [rfc](https://github.com/amundsen-io/rfcs/pull/37) to index query compositions.

Two new models are introduced:
- SqlJoin
- SqlWhere

In addition, the `TableSummary` was moved higher up in the file to enable it to be reused. 

### Tests

N/A - changes are limited to making schema models


dorianj (approved):
LGTM -- version bump?

PR #1158, opened 2021-05-25
feat: Databuilder & Metadata Query Components

sewardgw:
### Summary of Changes

This PR implements the databuilder and metadata changes to support[ indexing query components](https://github.com/amundsen-io/rfcs/pull/37).

Major changes:
- New query models in data builder
  - query
  - query execution
  - query where
  - query join
- Generic `QueryMetadataBaseExtractor` created
  - This was implemented with `JsonQueryMetadataExtractor` which expects a pre-defined data structure to be provided
  - Other implementations can extend this and override `_get_next_raw_record` to retrieve their query records (e.g. from a database, a file, etc.) and implement `parse_query_record(...)` to parse the query and generate the required objects to store in Amundsen
- Sample extractor with sample data using aforementioned JSON Query implementation

When capturing query components, users have the option to provide as much or as little data as they would like. For example, the [rfc](https://github.com/amundsen-io/rfcs/pull/37) depicts the following relationship

```text
Where -> Query -> Query Execution
```

This will allow Amundsen to rank order the most common `Wheres` by summing the query executions. Users may choose to make a tradeoff on complexity and only provide the `Where` without providing the `Query` or `Query Execution`. However, doing so will cause the `Where` to count as zero executions and always be ranked last. The same concept applies to `Joins`

### Tests

Added additional tests.

### Documentation

None included in the PR but will add if this feels like an appropriately complex item that should be documented.


databuilder/databuilder/extractor/query/json_query_metadata_extractor.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import (
+    Any, Dict, Iterator, List,
+)
+
+from databuilder.extractor.query.query_metadata_base import QueryMetadataBaseExtractor
+
+LOGGER = logging.getLogger(__name__)
+
+
+class JsonQueryMetadataExtractor(QueryMetadataBaseExtractor):
```
feng-tao, line 14: any reason the input list needs to be in this specific format?

normally you will get a list of queries from query log, i wonder who did the transformation into this format?

feng-tao (changes requested):

databuilder/databuilder/extractor/query/json_query_metadata_extractor.py
sewardgw: The thought here is that the objects in the list are relatively synonymous with the inputs required for each of the corresponding models and, for any implementer that _does_ end up parsing their own query, each of these objects can be extracted from a single query.

For example, the top level objects are:
- sql: straight forward, needs to come from _somewhere_ ( a file, query log, etc.)
- clean_sql (optional): a user defined sql that has been cleaned 
- user (optional): the user that executed the query
- query_tables (optional): corresponds to the model `QueryMetadata`
- query_wheres (optional): corresponds to the model `QueryWhereMetadata`
- query_joins (optional): this corresponds to the model `QueryJoinMetadata`

When an implementer subclasses `QueryMetadataBaseExtractor` they will have to implement `parse_query_record` at which point they _should_ be able to map their parsed query to the Amundsen parsed query.

feng-tao: thanks for sharing, but out of curious : any reason we don't do it in the transformer side which is the typical way on how we transform  / decorate a message into the target form ?

sewardgw: The only reason reason that a transformer wasn't used for the `JsonQueryMetadataExtractor` was because it was easier to set up an example JSON file that matched this exact structure and JSON was just easier than any other format that where there was a lower degree of cardinality between a Query and Joins / Where (which needs 1 to many).

I would expect that a custom implemented extractor would likely use a transformer inside of the `parse_query_record` function to both parse out their query and create the proper structure.

sewardgw: hey @feng-tao just wanted to bump this when you have a moment - any other questions / feedback on this PR that you'd like addressed before approving?

feng-tao: will take a closer look at the pr today, thanks.

databuilder/databuilder/extractor/query/json_query_metadata_extractor.py
```diff
@@ -0,0 +1,96 @@
...
+                        'join_sql': '3'
+                    }
+                ]
+            }
+        ]
+    Implementers should make sure that only VALID and SUCCESSFULLY run queries
+    are included in this list.
+    """
+
+    def __init__(self, query_records: List[Dict]) -> None:
```
feng-tao, line 79: why we need the constructor here instead of passing it through config?

databuilder/databuilder/extractor/query/query_metadata_base.py
```diff
@@ -0,0 +1,354 @@
...
+from databuilder.models.table_metadata import ColumnMetadata, TableMetadata
+from databuilder.models.user import User as UserMetadata
+
+LOGGER = logging.getLogger(__name__)
+
+
+class QueryMetadataBaseExtractor(Extractor):
+    """
+    Provides common funcitons used by query metadata extractors including:
+        - Converting a list of Stemma SQL tables to QueryMetadata,
```
feng-tao, line 25: please remove company mentioned in comment.

databuilder/databuilder/extractor/query/query_metadata_base.py
```diff
@@ -0,0 +1,354 @@
...
+        - Converting stemma where cluases to QueryWhereMetadata
```
feng-tao, line 26: same

databuilder/databuilder/extractor/query/query_metadata_base.py
```diff
@@ -0,0 +1,354 @@
...
+        - Converting stemma joins cluases to QueryJoinMetadata
```
feng-tao, line 27: same

databuilder/databuilder/extractor/query/query_metadata_base.py
feng-tao: Also could you give an example of SQL table in your use case ? cc @markgrover

databuilder/databuilder/extractor/query/query_metadata_base.py
```diff
@@ -0,0 +1,354 @@
...
+    Implementations need to provide the structure that iterates over records,
+    creates the Stemma SQL objects and then passses them to the corresponding
```
feng-tao, line 29: same

feng-tao (commented):
it is a bit long, will take a look at once have more time.

databuilder/databuilder/extractor/query/query_metadata_base.py
sewardgw: Ahh, yep, thanks for catching this. All will be removed in the next commit.

databuilder/databuilder/extractor/query/query_metadata_base.py
sewardgw: The SQL Table referenced here is simply a Python object that is used to serialize a "table" from our proprietary SQL parsing into an Amundsen format. To your point below, all references to this should have been removed.

databuilder/databuilder/extractor/query/json_query_metadata_extractor.py
sewardgw: AFAIK, the pyhocon package doesn't support passing dictionaries through it, it converts all of them into Config objects which turn turn loses many of the native python dict functionality. This made it super difficult to then use those dictionaries later on in the processing.

I _could_ pass in a file location through the config and then have the Class read in the contents of the file, but that would be an unnecessary step in the process for many use-cases that are doing query parsing (IMO). Actually spent a bit of time trying to get this to work w/ the config constructor and I'm open to other suggestions if there is a better way.

feng-tao:
I haven't had a chance to go through everything, but the model which the RFC is based upon looked fine. The concern would be :
- extractor assumes some certain specific input format without any example on how to gather them.  Normal event query audit log only captures : query, user. With the extractor, I am not sure if users would know how to gather all the information. 
- document how a general user would use these for.
- where do we surface these additional table<->query information?

Open for feedback if you think differently.

sewardgw:
Hey @feng-tao - To answer your points in respective order:

- This extractor _does_ make some opinionated decisions about the format of the data, however, I believe _all_ extractors do this, it's just that the complexity for this format is slightly higher. The reason this complexity is needed, IMHO, is because the relationships between queries, query executions and query components (e.g. joins, wheres) _is_ higher. In order to answer questions such as "what are the most relevant joins _now_" we need to have the connections between `joins > query > execution`. If we did not have a complex input format it would require multiple independent extractors (one for query / join, one for query / execution) etc. However, the complexity is mostly optional, users can submit as much or as little as they would like, for example users can use this extractor to only create a `join` and associate it to a table:
```json
{
      "query_joins": [
        {
          "left_table": {
            "schema": "test_schema",
            "table": "test_table1",
            "column": "col1"
          },
          "right_table": {
            "schema": "test_schema",
            "table": "test_view1",
            "column": "col1"
          },
          "join_type": "left outer join",
          "join_operator": "=",
          "join_sql": "test_table1.col1 = test_view1.col1"
        }
      ]
    },
```
or users can associate tables to a query with something like this:
```json
{
  "sql": "select col1, col2 from test_schema.test_view1 where col2 = '2021-05-01'",
  "query_tables": [
    {
      "schema": "test_schema",
      "table": "test_view1"
    }
  ],
}
```
- good call, I will add some documentation for how a general user could use this. There is also an example file + example extractor that can be used as reference.
- This PR is currently only addressing the creation of the relationships between these data models. I have other PRs lined up for metadata & frontend that will begin to surface these relationships for users but as of now I have not worked specifically on the  table<->query information, only on the wheres / joins <-> table information.

sewardgw:
@feng-tao just wanted to bump this for when you have a moment to review the updates

feng-tao:
Sorry for the delay, I'am +1 on the model, the only concern would be the extractor which assume users in the open source community to have some kinds of pre-defined parsing capability to translate the queries log into the data structures without providing much examples how to do the connection (the sample loader already assumes the input is the pre-defined data structure ).  For the context, Amundsen's norm  is to expect the input to be generic, then each could apply its own logic in transformer to do the model conversion.

Curious to see how others think. cc @jinhyukchang @allisonsuarez @mgorsk1

mgorsk1:
I think we already established precedence a couple of days ago with GenericTableUsageExtractor but agree that without any examples on how to reach the shape required by extractor this doesn't bring that much value for end users (which is not the case with extractor mentioned before). Are there any available libs out there that could be used to achieve this?

sewardgw:
There _are_ some, [this being one example](https://pypi.org/project/moz-sql-parser/), but I would not go so far as to recommend any particular library simply because there will be a large effort by the person / team that implements SQL parsing. There isn't a plug-and-play library that does this. 

From my perspective, there are prior examples in Amundsen to open source models and extractors that do not have defined ways in which to extract them. @mgorsk1 mentioned one above, table ownership is another, as is table usage. For this extractor, the underlying data structures apply to _all_ SQL languages (see comments above as well as data in the sample JSON loader file). I think it is reasonable to expect that someone who is spending the time to properly parse queries will also be able to format that object into the proposed structure (or, if someone provides a better approach then this one could also be deprecated).

markgrover:
Thanks for your review @feng-tao and @mgorsk1, and thanks for putting this PR up, @sewardgw 
In the spirit of making progress, my request would be to push this forward.

This PR was put up 3 weeks ago and of course, deep feedback requires time to give and incorporate. This is a key feature that our open-source competitors don't have and we have to keep up the development velocity of the project so those considering Amundsen see it clearly as a winner - with much better experience, integrations and adoption. Would love your help and support in pushing us forward in that direction.

feng-tao:
> ack requires time to give and incorporate. This is a key feature that our open-source competitors don't have and we have to keep up the development velocity of the project so those considering Amundsen see it clearly as a winner - with much better experience, integrations and adoption. Would love your help and support in pushing us forward in that direction.

Sure, I am happy to move forward with the pr if we remove the extractor and have the models alone which matches the what the original RFC defines. What I am not satisfied so far is the limitation of the extractor. For your context, we contribute lots of data models without having an extractors which allow users to ingest their data in their own way. We have the extractors only if the extractor is generic enough with good e2e examples to be used.

feng-tao:
For your further context, Lyft has its own private extractor to deal with all kinds various metadata extractions without upstream those extractors back to OSS. The reason is because the extractors are pretty custom and hard to be used for the OSS community.

sewardgw:
@feng-tao I've updated the PR to remove the extractors, could you take a look when you have a moment?

databuilder/example/scripts/sample_json_query_metadata_loader.py
```diff
@@ -0,0 +1,171 @@
...
+import os
+import sys
+import uuid
+
+from elasticsearch.client import Elasticsearch
+from pyhocon import ConfigFactory
+
+from databuilder.extractor.neo4j_extractor import Neo4jExtractor
+from databuilder.extractor.neo4j_search_data_extractor import Neo4jSearchDataExtractor
+from databuilder.extractor.query.json_query_metadata_extractor import JsonQueryMetadataExtractor
```
feng-tao, line 20: do we still need this sample loader?

feng-tao (commented):
lgtm with 1 question

databuilder/example/scripts/sample_json_query_metadata_loader.py
feng-tao: I think we don't run flake8/mypy on scripts folder, otherwise this file will fail

databuilder/docs/query_metadata_guide.md
```diff
@@ -0,0 +1,175 @@
...
+
+The `QueryExecutionsMetadata`, `QueryJoinMetadata`, `QueryWhereMetadata` and `QueryMetadata` can be seen here:
+
+![Query Assets](./assets/query-assets.png?raw=true "Query Assets")
+
+Amundsen uses `QueryExecutionsMetadata` to determine time-sensitive relevance. As new `QueryExecutionsMetadata` are added and old ones are removed, Amundsen is able to continue to keep the most recent queries and the related joins and wheres relevant.
+
+Since `QueryMetadata` can be related to `QueryExecutionsMetadata`, `QueryJoinMetadata` and `QueryWhereMetadata`, a single extractor is used to ingest these objects into Amundsen.
+
+### Using Extractors
```
feng-tao, line 20: could you update this section as well?

feng-tao (approved):

PR #1163, opened 2021-05-27
refactor: shared dependencies unification

mgorsk1:
### Summary of Changes

Following merging https://github.com/amundsen-io/rfcs/pull/38 this PR unifies dependencies across amundsen services.

#### Assumptions

- global unification (& upgrade to more recent vrersions) of libraries has been conducted (every repo now uses same version for pytest, flake8 and so on). if there were mulitple versions of library X across different projects (for example - we had 3 different versions of mypy - `0.660`, `0.782` and `0.812` across frontend,metadata and search) - the most recent one was selected for unification (in case of mypy - `0.812`)
- any linting issues that has arisen because of the previous point are solved by excluding certain types of errors (fixing linting in whole repo would make this PR a major pain to review and resolve)
- databuilder is excluded from requirements-common.txt unification 

### Tests

- all existing linter rules passing
- all existing unit tests passing

#### TODO

- verify correctness of building docker images

### Documentation

- updated appropriate README.md and installation.md files

### CheckList


frontend/Makefile
```diff
@@ -25,11 +25,11 @@ test: test_unit lint mypy
 
 .PHONY: image
 image:
-	docker build -f public.Dockerfile -t ${IMAGE}:latest .
+	cd .. && docker build -f frontend.public.Dockerfile -t ${IMAGE}:latest . && cd frontend
```
bbsbb, line 28: Shouldn't this file refer to Dockerfile.frontend.public ?

mgorsk1: it should, nice catch!

Dockerfile.frontend.local
```diff
@@ -0,0 +1,29 @@
...
+RUN npm rebuild node-sass
+RUN npm run dev-build
+
+COPY ./frontend /app
+
+FROM python:3.7-slim
+WORKDIR /app
+
+COPY ./frontend /app
+COPY requirements-dev.txt /app/requirements-dev.txt
```
sewardgw, line 21: Are there dev requirements needed for the container?

search/Makefile
```diff
@@ -49,3 +49,7 @@ push-image:
 .PHONY: build-push-image
 build-push-image-latest: image push-image
 build-push-image-version: image-version push-image-version
+
+.PHONY: install_deps
+install_deps:
+	pip3 install -e ".[all]"
```
sewardgw, line 55: for each of the `install_deps` in the make files, is the `-e` strictly required when doing pip install? 

I'm trying to think if there is a developer experience where the dev may do `pip install .[all]` and if it could fail.

metadata/requirements-common.txt
```diff
@@ -0,0 +1 @@
+../requirements-common.txt
```
sewardgw, line 1: nice! Didn't know you could just reference files like this

sewardgw (approved):
Thanks for the work here! clearly a big effort for a much needed change. A few comments / questions inline

search/Makefile
mgorsk1: it will fail if you skip `-e`. Without `-e` it makes a copy of a source dir to /tmp and conducts installation there. When this copy operation is conducted symlinks are not resolved so requirements-common.txt (or requirements-dev.txt) are not copied too. Adding -e flag changes the process in such way that files are not copied to /tmp beforehand but installation is conducted in source dir.

metadata/requirements-common.txt
mgorsk1: we use it in several places in `docs` subdirectory :)

Dockerfile.frontend.local
mgorsk1: without this file `pip install -e .[all]` fails cause `requirements-dev.txt` is missing (as `setup.py` fails on missing file when reading in to `requirements_dev` variable). 
I can either:
- remove this file from `Dockerfile` and introduce handling of missing files in `setup.py`  (load empty array to `requirements_dev` variable if the file is missing)
- keep this file in `Dockerfile` (they are not installed actually, just the file needs to be there `setup.py` to execute correctly) and not amend `setup.py` 

my preference is the latter one but if you feel differently I can make the change.

sewardgw: Thanks for the context. I agree with this approach. This PR was also about unifying the requirements, which appears to have been done. Should this be an issue in the future it can be addressed then.

search/Makefile
sewardgw: Got it. Trying to think through a scenario where someone makes changes / installs from source but forgets to add the `-e` flag. Is it worth updating the readme / docs somewhere to mention this?

sewardgw (approved):

sewardgw:
One last thought - worth squashing some of the commits?

mgorsk1:
> One last thought - worth squashing some of the commits?

I always `Squash and Merge` so main stays decluttered but good point, it's a messy one!

Dockerfile.frontend.local
mgorsk1: I've updated README.md for search and metadata + `docs/installation.md` for frontend. Should I add it also somewhere else?

mgorsk1:
I think this requires approval of codeowner of all repos to merge, are you up for review @feng-tao ?

databuilder/setup.py
```diff
@@ -1,25 +1,18 @@
...
 
 from setuptools import find_packages, setup
 
 __version__ = '4.4.1'
 
+requirements_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'requirements.txt')
+with open(requirements_path) as requirements_file:
+    requirements = requirements_file.readlines()
 
-requirements = [
```
dorianj, line 10: I thought for databuilder, we didn't want to specify exact deps (since it may be installed alongside other stuff) -- I think we should either update the requirements.txt to be less-specific, or remove this. WDYT?

dorianj (approved):

requirements-dev.txt
```diff
@@ -0,0 +1,16 @@
+# Copyright Contributors to the Amundsen project.
+# SPDX-License-Identifier: Apache-2.0
+
+# Common dependencies for code quality control (testing, linting, static checks, etc.) ---------------------------------
+
+flake8==3.9.2
+flake8-tidy-imports==4.3.0
+isort[colors]~=5.8.0
+mock==4.0.3
+mypy==0.812
```
feng-tao, line 10: I think it could be problematic for databuilder given it is used for library not a service like other modules, we should at least make it as range instead of pin down to certain versions.

databuilder/setup.cfg
```diff
@@ -13,7 +13,7 @@ exclude =
   databuilder/sql_parser/usage/presto/antlr_generated
 max-complexity = 10
 max-line-length = 120
-ignore = NONE
+ignore = E117,E252,E741,F541,F841,W504,W605
```
feng-tao, line 16: could you give some context what we ignore here?

databuilder/setup.py
feng-tao: +1 on dorian's comment

databuilder/setup.py
```diff
@@ -85,14 +78,14 @@
     'mysqlclient>=1.3.6,<3'
 ]
 
-all_deps = requirements + kafka + cassandra + glue + snowflake + athena + \
+all_deps = requirements + requirements_dev + kafka + cassandra + glue + snowflake + athena + \
```
feng-tao, line 81: also why not put common into databuilder given eventually we should align model definition between common and databuilder?

feng-tao (changes requested):

databuilder/setup.py
mgorsk1: requirements_common contain mostly shared dependencies for Flask services (+ obviously amundsen-common). during rfc we decided to exclude databuilder from sharing those (which would make sense if we decide not to pin any deps in databuilder, while in requirements_common flask and other deps are need to be pinned).

requirements-dev.txt
mgorsk1: ok, so `>=` for all requirements dev ?

databuilder/setup.py
mgorsk1: it doesn't really change the approach we had before but if you think that with this PR we want to stop pinning deps in databuilder (at least those generic ones) then I will remove version pinning.

databuilder/setup.cfg
mgorsk1: I think that with bumping flake8 it just became more strict with checks. `make lint` on `main` branch passes while here (after bumping from 3.5.0 to 3.9.2) it yields

```
(.venv) root@mariusz-XPS-13-7390-2-in-1:/amundsen/databuilder# make lint
flake8 .
./databuilder/task/task.py:72: [F541] f-string is missing placeholders
./databuilder/publisher/elasticsearch_publisher.py:69: [W504] line break after binary operator
./databuilder/publisher/elasticsearch_publisher.py:80: [E741] ambiguous variable name 'l'
./databuilder/publisher/neo4j_csv_publisher.py:460: [F541] f-string is missing placeholders
./databuilder/publisher/atlas_csv_publisher.py:89: [F541] f-string is missing placeholders
./databuilder/publisher/atlas_csv_publisher.py:104: [F541] f-string is missing placeholders
./databuilder/publisher/atlas_csv_publisher.py:217: [F541] f-string is missing placeholders
./databuilder/models/atlas_serializable.py:52: [F541] f-string is missing placeholders
./databuilder/models/schema/schema.py:29: [E252] missing whitespace around parameter equals
./databuilder/models/schema/schema.py:29: [E252] missing whitespace around parameter equals
./databuilder/models/schema/schema.py:30: [E252] missing whitespace around parameter equals
./databuilder/models/schema/schema.py:30: [E252] missing whitespace around parameter equals
./databuilder/extractor/dbt_extractor.py:124: [E117] over-indented
./databuilder/extractor/hive_table_last_updated_extractor.py:228: [F541] f-string is missing placeholders
./databuilder/extractor/delta_lake_metadata_extractor.py:42: [W504] line break after binary operator
./databuilder/extractor/delta_lake_metadata_extractor.py:43: [W504] line break after binary operator
./databuilder/extractor/delta_lake_metadata_extractor.py:44: [W504] line break after binary operator
./databuilder/extractor/delta_lake_metadata_extractor.py:45: [W504] line break after binary operator
./databuilder/extractor/delta_lake_metadata_extractor.py:46: [W504] line break after binary operator
./databuilder/extractor/atlas_search_data_extractor.py:237: [F841] local variable 'e' is assigned to but never used
./databuilder/extractor/atlas_search_data_extractor.py:354: [F541] f-string is missing placeholders
./databuilder/extractor/dashboard/apache_superset/apache_superset_chart_extractor.py:57: [F541] f-string is missing placeholders
./databuilder/extractor/dashboard/apache_superset/apache_superset_metadata_extractor.py:59: [F541] f-string is missing placeholders
./databuilder/extractor/dashboard/tableau/tableau_dashboard_last_modified_extractor.py:43: [W504] line break after binary operator
./databuilder/extractor/dashboard/tableau/tableau_dashboard_extractor.py:49: [W504] line break after binary operator
./example/scripts/sample_dbt_loader.py:68: [F541] f-string is missing placeholders
./example/scripts/sample_deltalake_metadata.py:52: [F541] f-string is missing placeholders
./example/scripts/sample_pandas_profiling_column_stats_loader.py:16: [F541] f-string is missing placeholders
./example/scripts/sample_feast_loader.py:54: [F541] f-string is missing placeholders
./tests/unit/publisher/test_mysql_csv_publisher.py:22: [F541] f-string is missing placeholders
./tests/unit/publisher/test_neo4j_csv_publisher.py:23: [F541] f-string is missing placeholders
./tests/unit/extractor/test_csv_extractor.py:24: [F541] f-string is missing placeholders
./tests/unit/extractor/test_kafka_source_extractor.py:18: [F541] f-string is missing placeholders

```

requirements-dev.txt
feng-tao: let's make it range .

databuilder/setup.cfg
feng-tao: could you file a ticket and see if we could get some community member to help fixing these?

mgorsk1: sure, will do after merging. sounds like a nice idea for `good-first-issue`

feng-tao (approved):

databuilder/setup.py
mgorsk1: I have started to work on requirements.txt in databuilder per your suggestion. I didn't find a lot of deps there that are directly mentioned in any of the `extras_require` section. There were few I've moved (to atlas and rds) and my approach for the remaining ones would be not to remove pinning deps completely but instead replace `==` with `>=` . This way we at least we don't fall into regression issues if some library requires `lower` version than we've previously supported. Let me know your thoughts on that. @feng-tao @dorianj

PR #1168, opened 2021-05-28
fix: failing tests after introducing queries into table model

mgorsk1:
### Summary of Changes

added missing attributes to expected data in tests

### Tests

- fixed

### Documentation

n/a

### CheckList


mgorsk1:
cc @sewardgw - I wonder if it would make sense to always run tests on all packages (now it's run only if some files in given subdir changes). wdyt @dorianj

PR #1187, opened 2021-06-03
chore: Updates the mypy version to `0.782`

verdan:
### Summary of Changes

This PR updates the mypy version, as the older version does not work with Python 3.8, and we are already using the latest version in metadata repository.

### CheckList

### DIFF
--- a/search/requirements.txt
+++ b/search/requirements.txt
@@ -18,7 +18,7 @@ jsonschema>=3.0.1,<4.0
 marshmallow>=3.0,<=3.6
 marshmallow3-annotations>=1.0.0
 mock==2.0.0
-mypy==0.660
+mypy==0.782
 pytest==3.5.1
 pytest-cov==2.5.1
 pytest-env==0.6.2


mgorsk1:
I've already bumped it to `0.812` here https://github.com/amundsen-io/amundsen/pull/1163 (alongside couple of different qa tools with more recent versions). not sure if makes sense to do it one by one, wdyt?

verdan:
> I've already bumped it to `0.812` here #1163 (alongside couple of different qa tools with more recent versions) across all projects. not sure if makes sense to do it one by one, wdyt?

@mgorsk1  sure thing, I'm happy to drop this in favor of your PR. 
One remark though: would that make sense to only consolidate the requirements now, and use the version we are already using in any service? Updates to the latest versions go to a new PR? thoughts?

mgorsk1:
agree, however when centralizing the versions of libraries in #1163 I have not introduced completely new versions. If 3 different projects used mypy with different versions, for `requirements-dev.txt` I went with the latest out of those 3, not something entirely new. Not sure what would be better way to do this.

verdan:
ah, I see. this is perfect then. I also meant the same to use the latest of what we are already using in any service. 👍

mgorsk1:
cool, I've updated the #1163 to be more explicit about approach I went with!

PR #1192, opened 2021-06-03
feat: Generic Usage Extractor with Scala custom sql script for Snowflake

alldoami:
### Summary of Changes

Implemented a Generic Usage Extractor which utilizes the TableColumnUsage to fill in the popular tables and frequent users feature. CZI uses Snowflake very heavily and because Snowflake does not have table usage built in like BigQuery, I've created a custom sql parser script in Scala that I run in Databricks from an Airflow DAG. The way I calculate popularity is by reading the query logs, parsing what tables are being used within the query, insert a row into the Snowflake popularity table with a unique user, and add one to its score.

### Tests

I would love help to write tests for this! I have this deployed in 5 different Airflow deployments right now and it works great.

### Documentation

A lot of this has come up from a conversation through Slack: https://amundsenworkspace.slack.com/archives/C01857S0V1V/p1619635080092300

Would love to credit @lexisantoro for providing the extractor code! It mainly looks like the snowflake metadata extractor besides the `_get_extract_iter` function.

### CheckList


databuilder/databuilder/extractor/snowflake_usage_extractor.py
```diff
@@ -0,0 +1,106 @@
...
+    def extract(self):
+        if not self._extract_iter:
+            self._extract_iter = self._get_extract_iter()
+        try:
+            return next(self._extract_iter)
+        except StopIteration:
+            return None
+
+    def get_scope(self) -> str:
+        return "extractor.snowflake"
```
sewardgw, line 80: this scope should be unique to the extractor, I believe this one matches the snowflake metadata scope.

databuilder/README.md
```diff
@@ -356,6 +356,39 @@ job = DefaultJob(
 job.launch()
 ```
 
+#### [SnowflakeUsageExtractor](https://github.com/amundsen-io/amundsendatabuilder/blob/master/databuilder/extractor/snowflake_usage_extractor.py "SnowflakeUsageExtractor")
+An extractor that extracts table popularity metadata from a custom created Snowflake table (created by a script that may look like [this scala script](https://github.com/amundsen-io/amundsendatabuilder/blob/master/databuilder/example/scripts/sample_snowflake_table_usage.scala "sample_snowflake_table_usage")). You can create a DAG using the [Databricks Operator](https://github.com/apache/airflow/blob/main/airflow/providers/databricks/operators/databricks.py) and run this script within Databricks.
```
sewardgw, line 360: Is there an alternative or more generic method than using this databricks example? Just thinking through how this could be used by those who don't have databricks

databuilder/databuilder/extractor/snowflake_usage_extractor.py
```diff
@@ -0,0 +1,106 @@
...
+    # SELECT statement from snowflake table that contains usage data
+    SQL_STATEMENT = """
+        SELECT
+            database,
+            schema,
+            name,
+            user_email,
+            read_count
+        FROM
+            {database}.{schema}.{table}
```
sewardgw, line 33: Given that this table will not be available by default in snowflake and that some pre-processing needs to occur to create this table, should this executor include the curation of this table _and then_ the extraction of this metadata?

databuilder/databuilder/extractor/snowflake_usage_extractor.py
```diff
@@ -0,0 +1,106 @@
...
+        Using raw level iterator, it groups to table and yields TableColumnUsage
+        :return:
+        """
+        for row in self._get_raw_extract_iter():
+            col_readers = []
+            col_readers.append(ColumnReader(database="snowflake",
+                                            cluster=row["database"],
+                                            schema=row["schema"],
+                                            table=row["name"],
+                                            column="*",
```
sewardgw, line 93: It looks like `ColumnReader` does currently persist the column in the database (although why, I'm not sure..). Is it worth capturing the column here in the event that `ColumnReader` is updated in the future to do something with that field?

databuilder/example/scripts/sample_snowflake_table_usage.scala
```diff
@@ -0,0 +1,132 @@
+"""
+This is a example script that parses table names from Snowflake's query logs and calculates their popularity based on
+how many times that same user queries that specific table. You can run this in a Databricks notebook as a job within
```
sewardgw, line 3: If this can be executed against any snowflake account, can the references to databricks be removed?

sewardgw (commented):
Great start! Added comments inline

databuilder/README.md
alldoami: You can run this anywhere you have the ability to run scala! This is databricks specific, it's just the environment we use to run scala jobs.

databuilder/databuilder/extractor/snowflake_usage_extractor.py
alldoami: That's a great suggestion. I do think that wherever you're writing the data to though should include the table creation. You'd still need info about the database and schema, and then the table name would be variable. So pre-processing would still need to happen if you want to place this in a specific separate db/schema. Does that sound reasonable?

databuilder/databuilder/extractor/snowflake_usage_extractor.py
alldoami: Ya I was seeing this: https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/models/table_column_usage.py#L63-L64 so didn't think it'd be worth putting in info about the column because it might be confusing... we could put a note though...

feng-tao (commented):
amundsen-kube-helm/templates/helm/charts/elasticsearch-1.32.5.tgz  is checked incorrectly btw

README.md
```diff
@@ -221,13 +221,13 @@ Notes from all past meetings are available [here](https://docs.google.com/docume
...
 
 Currently **officially** using Amundsen:
-
 1. [Asana](https://asana.com/)
 1. [Bagelcode](https://site.bagelcode.com/)
 1. [Bang & Olufsen](https://www.bang-olufsen.com/en)
 1. [Brex](https://www.brex.com/)
 1. [Cameo](https://www.cameo.com)
 1. [Cimpress Technology](https://cimpress.com)
+1. [Chan Zuckerberg Initiative](https://chanzuckerberg.com)
```
feng-tao, line 230: could you make it sorted alphabetically? thanks , great to see your team onboard to Amundsen!

databuilder/example/scripts/sample_snowflake_table_usage.scala
```diff
@@ -0,0 +1,131 @@
...
+
+val keySpec = new PKCS8EncodedKeySpec(pkcs8EncodedBytes)
+val kf = KeyFactory.getInstance("RSA")
+val privateKey = kf.generatePrivate(keySpec)
+
+val url = "jdbc:snowflake://snowflake_account_name.snowflakecomputing.com";
+val prop = new Properties();
+
+// TODO: change these values to your snowflake creds
+val user = "user@email.com"
```
feng-tao, line 51: interesting, to make sure I understand correctly: you run a databricks scala job to read and calculate snowflake table usage?

README.md
alldoami: lol i don't know how to alphabetize 🤦‍♀️

databuilder/example/scripts/sample_snowflake_table_usage.scala
alldoami: yup! exactly. We also store our creds (user and private key) in databricks secrets 😄 it's super cool. Scala isn't my language of choice, but it had the best sql parser package.

mgorsk1: nice! we are doing very similar thing but with Apache Ranger audit logs, which makes it easier cause they are already json serialized.

databuilder/databuilder/extractor/snowflake_usage_extractor.py
mgorsk1: I think the point here also would be - what makes this extractor `Snowflake` specific? We have a very similar table in hive only with parsed Apache Ranger Logs. The columns are much less identical (the names vary slightly). For me Snowflake specific part is one step before (where you create this table using scala parser) - maybe we could treat this extractor in a more generic fashion?

databuilder/databuilder/extractor/snowflake_usage_extractor.py
```diff
@@ -0,0 +1,106 @@
...
+        {where_clause_suffix};
+    """
+    # CONFIG KEYS
+    WHERE_CLAUSE_SUFFIX_KEY = 'where_clause_suffix'
+    POPULARITY_TABLE_DATABASE = 'popularity_table_database'
+    POPULARTIY_TABLE_SCHEMA = 'popularity_table_schema'
+    POPULARITY_TABLE_NAME = 'popularity_table_name'
+
+    DEFAULT_CONFIG = ConfigFactory.from_dict(
+        {WHERE_CLAUSE_SUFFIX_KEY: ' ',
```
mgorsk1, line 43: a side-note: sql query could as well have hardcoded `WHERE` close and default condition would then be 1=1. With current approach users also need to provide `WHERE` keyword here, am I right?

mgorsk1 (commented):
Thanks for awesome contribution. I've added minor nits - mostly around the fact this extractor looks like a very generic one - only creating the input table is snowflake specific really. Extraction part could be based on any kind of table. Let me know if you think it makes sense to generalize it a little bit.

databuilder/databuilder/extractor/snowflake_usage_extractor.py
```diff
@@ -0,0 +1,106 @@
...
+        return "extractor.snowflake_usage"
+
+    def _get_extract_iter(self) -> Iterator[TableColumnUsage]:
+        """
+        Using raw level iterator, it groups to table and yields TableColumnUsage
+        :return:
+        """
+        for row in self._get_raw_extract_iter():
+            col_readers = []
+            col_readers.append(ColumnReader(database="snowflake",
```
mgorsk1, line 89: this too could be a parameter for the extractor

databuilder/databuilder/extractor/snowflake_usage_extractor.py
sewardgw: +1 to this. The structure of the table you've defined seems quite reasonable and something that I think is a good idea to standardize on:
```
SELECT
        database,
        schema,
        name,
        user_email,
        read_count
    FROM
{database}.{schema}.{table}
```
Can this extractor take a reference to the database driver (e.g. pg, snowflake, etc.) so we can reuse it?

databuilder/databuilder/extractor/snowflake_usage_extractor.py
sewardgw: yea, makes sense. Column level parsing (such as in your scala script) is also much more complex

databuilder/tests/unit/extractor/test_snowflake_usage_extractor.py
```diff
@@ -0,0 +1,99 @@
...
+            self.assertEqual(results, None)
+
+    def test_extraction(self) -> None:
+        with patch.object(SQLAlchemyExtractor, '_get_connection') as mock_connection:
+            connection = MagicMock()
+            mock_connection.return_value = connection
+            sql_execute = MagicMock()
+            connection.execute = sql_execute
+
+            sql_execute.return_value = [{
```
sewardgw, line 45: 👍🏻

databuilder/databuilder/extractor/snowflake_usage_extractor.py
mgorsk1: If we standardize then this should be `user_id` as different companies use different attributes for id.

alldoami: Ya I kept the table the same as the one that `TableColumnUsage` reads in: https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/models/table_column_usage.py#L25-L31, would we want to deviate from this?

databuilder/databuilder/extractor/snowflake_usage_extractor.py
alldoami: yes users have to provide the `WHERE` keyword. I used the same convention as the `snowflake_metadata_extractor`: https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/extractor/snowflake_metadata_extractor.py#L70

databuilder/databuilder/extractor/snowflake_usage_extractor.py
alldoami: I also didn't think about making this a generic extractor because I'm not familiar with that many other database drivers. Would the only information you'd need is the type of database? (like just making the `database` variable a variable instead of me hardcoding it as `snowflake`?)

alldoami: https://github.com/amundsen-io/amundsen/pull/1192#discussion_r645361211

databuilder/databuilder/extractor/generic_usage_extractor.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class GenericUsageExtractor(Extractor):
+    # SELECT statement from table that contains usage data
+    SQL_STATEMENT = """
+        SELECT
+            database,
+            schema,
+            name,
+            user_email,
```
mgorsk1, line 31: can we rename it to user_id? I think some projects might not rely on email for user id and this could be misleading

databuilder/databuilder/extractor/generic_usage_extractor.py
```diff
@@ -0,0 +1,111 @@
...
+            read_count
+        FROM
+            {database}.{schema}.{table}
+        {where_clause_suffix};
+    """
+    # CONFIG KEYS
+    DATABASE_DRIVER = ''
```
mgorsk1, line 38: should that remain empty? if that's the property for sql alchemy connection the name could be misleading too

mgorsk1 (commented):
few nits

databuilder/databuilder/extractor/generic_usage_extractor.py
alldoami: Oops sorry! Replaced this with `DATABASE_KEY`

databuilder/databuilder/extractor/generic_usage_extractor.py
alldoami: hmm the ColumnReader is `user_email` though... I would want to keep it exactly as the ColumnReader is reading it... https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/models/table_column_usage.py#L30

mgorsk1: ok, that makes sense then!

PR #1194, opened 2021-06-04
chore: update requirement file in metadata

feng-tao:
Signed-off-by: feng-tao <fengtao04@gmail.com>

### Summary of Changes

Put additional proxy requirement into setup.py extra section.

Some contexts: https://amundsenworkspace.slack.com/archives/CHQERT0D7/p1622780101143900, no RC of the issue yet

### CheckList

### DIFF
--- a/metadata/setup.py
+++ b/metadata/setup.py
@@ -25,7 +25,12 @@ setup(
     install_requires=requirements,
     extras_require={
         'oidc': ['flaskoidc==0.2.3'],
-        'atlas': ['apache-atlas==0.0.11']
+        'atlas': ['apache-atlas==0.0.11',
+                  'amundsenatlastypes==1.2.0'],
+        'rds': ['amundsen-rds==0.0.5',
+                'mysqlclient>=1.3.6,<3',
+                'sqlalchemy>=1.3.6,<1.4',
+                'alembic>=1.2,<2.0']
     },
     python_requires=">=3.6",
     classifiers=[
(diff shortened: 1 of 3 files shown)


feng-tao:
cc @mgorsk1 @crazy-2020

crazy-2020 (approved):

mgorsk1:
this makes it more difficult to use default docker images with atlas and rds (for neo4j you don't need to do anything else, for atlas and rds you will need to prebuild i'd additionally). maybe we should think about publishing separate docker images for each backend so the users are not affected by this. wdyt?

metadata/setup.py
```diff
@@ -25,7 +25,12 @@
     install_requires=requirements,
     extras_require={
         'oidc': ['flaskoidc==0.2.3'],
-        'atlas': ['apache-atlas==0.0.11']
+        'atlas': ['apache-atlas==0.0.11',
+                  'amundsenatlastypes==1.2.0'],
```
mgorsk1, line 29: most recent is 1.2.2

feng-tao:
> this makes it more difficult to use official docker images with atlas and rds (for neo4j you don't need to do anything else, for atlas and rds you will need to prebuild additionally). Dockerfile should still build with all deps. long term maybe we should think about publishing separate docker images for each backend so the users are not affected by this. wdyt?

+1 on separate docker image for different proxy

metadata/setup.py
feng-tao: update

mgorsk1:
only thing then is to figure out if we have N * 2 images (N - number of proxies times two cause there is one with and one wihtout oidc right now)

allisonsuarez (approved):
LGTM

PR #1196, opened 2021-06-05
feat: Elasticsearch Metadata Extractors

mgorsk1:
### Summary of Changes

Introduces extractors for Elasticsearch metadata:
- Index extractor utilizing TableMetadata to feed Amundsen index metadata from ES
- Column Stats extractor utilizing ES Aggregation to provide stats on numerical properties

### Tests

- added tests for each extractor

### Documentation

- added docs for each extractor

### CheckList


feng-tao (commented):
hey @mgorsk1 , I maybe missing something fundamental, but ES shouldn't be the source of truth as the index is always rebuilt from the metadata store. Why we want to extract from ES instead of metadata?

mgorsk1:
Do you mean that ES Indexes are always populated from some other DBs like postgres or snowflake and we should only ingest metadata from those systems ?

I can think of several use cases where ES index is the only source of truth (like in logging use-cases, ingesting OCR data). On the other hand, if the index indeed was created by extracting table from (for example) postgresql then it's also valuable to have this information for the lineage purposes.

feng-tao:
> you mean that ES Indexes are always populated from some other DBs like postgres or snowflake and we should only ingest metadata from those systems ?

yeah, what I said is only from Amundsen context that ES is always used for search purpose but not for metadata source of truth. ES certainly could be source of truth for other use cases like you said.

mgorsk1:
Ok, maybe it's worth to update the docs though if that was unclear. The purpose of the extractor is indeed for users relying on ES for something else than Amundsen (like use-cases I mentioned above).

databuilder/databuilder/extractor/es_base_extractor.py
```diff
@@ -0,0 +1,86 @@
...
+from databuilder.extractor.base_extractor import Extractor
+
+
+class ElasticsearchBaseExtractor(Extractor):
+    """
+    Extractor to extract index metadata from Elasticsearch
+    """
+
+    ELASTICSEARCH_CLIENT_CONFIG_KEY = 'client'
+    ELASTICSEARCH_EXTRACT_TECHNICAL_DETAILS = 'extract_technical_details'
```
feng-tao, line 19: could you give a comment on config value ?

feng-tao (commented):
Also if the extractor is using for the metadata stored in ES, have you looked into es dbapi (e.g https://github.com/preset-io/elasticsearch-dbapi ) given we have a dbapi extractor?

databuilder/databuilder/extractor/es_base_extractor.py
mgorsk1: it's included it in docs.

mgorsk1: As for dbapi - from the docs I can see that `This library supports Elasticsearch 7.X versions.` - I tried it with 6.8.13 and failed. Also - I am not sure the column statistics part would be achieved using this library.

mgorsk1:
ping @feng-tao for re-review

feng-tao:
@mgorsk1 sorry for the delay, but I kinda packed this week and may not have time to go through the code. Maybe other folks (cc @verdan @allisonsuarez ) to take a look...

verdan:
Sure thing. Will be able to review tomorrow.

databuilder/README.md
```diff
@@ -1270,6 +1270,114 @@ job = DefaultJob(conf=job_config,
 job.launch()
 ```
 
+### [ElasticsearchMetadataExtractor](./databuilder/extractor/es_metadata_extractor.py)
+
+The included `ElasticsearchMetadataExtractor` provides support for extracting basic metadata for Elasticsearch indexes.
+
+It extracts index metadata into `TableMetadata` model so the results are retrievable the same way as table metadata.
```
verdan, line 1310: any reason we are using TableMetadata model here? Shouldn't this be a new model for this use case?

databuilder/README.md
```diff
@@ -1303,6 +1303,114 @@ job = DefaultJob(conf=job_config,
...
+node_files_folder = f'{tmp_folder}/nodes'
+relationship_files_folder = f'{tmp_folder}/relationships'
+
+dict_config = {
+    f'loader.filesystem_csv_neo4j.{FsNeo4jCSVLoader.NODE_DIR_PATH}': node_files_folder,
+    f'loader.filesystem_csv_neo4j.{FsNeo4jCSVLoader.RELATION_DIR_PATH}': relationship_files_folder,
+    f'loader.filesystem_csv_neo4j.{FsNeo4jCSVLoader.SHOULD_DELETE_CREATED_DIR}': True,
+    f'extractor.es_indexes.{ElasticsearchMetadataExtractor.CLUSTER}': 'demo',
+    f'extractor.es_indexes.{ElasticsearchMetadataExtractor.SCHEMA}': 'dev',
+    f'extractor.es_indexes.{ElasticsearchMetadataExtractor.ELASTICSEARCH_CLIENT_CONFIG_KEY}': Elasticsearch()
```
verdan, line 1350:
```suggestion
    f'extractor.es_metadata.{ElasticsearchMetadataExtractor.CLUSTER}': 'demo',
    f'extractor.es_metadata.{ElasticsearchMetadataExtractor.SCHEMA}': 'dev',
    f'extractor.es_metadata.{ElasticsearchMetadataExtractor.ELASTICSEARCH_CLIENT_CONFIG_KEY}': Elasticsearch()
```

databuilder/README.md
```diff
@@ -1303,6 +1303,114 @@ job = DefaultJob(conf=job_config,
...
+node_files_folder = f'{tmp_folder}/nodes'
+relationship_files_folder = f'{tmp_folder}/relationships'
+
+dict_config = {
+    f'loader.filesystem_csv_neo4j.{FsNeo4jCSVLoader.NODE_DIR_PATH}': node_files_folder,
+    f'loader.filesystem_csv_neo4j.{FsNeo4jCSVLoader.RELATION_DIR_PATH}': relationship_files_folder,
+    f'loader.filesystem_csv_neo4j.{FsNeo4jCSVLoader.SHOULD_DELETE_CREATED_DIR}': True,
+    f'extractor.es_indexes.{ElasticsearchColumnStatsExtractor.CLUSTER}': 'demo',
+    f'extractor.es_indexes.{ElasticsearchColumnStatsExtractor.SCHEMA}': 'dev',
+    f'extractor.es_indexes.{ElasticsearchColumnStatsExtractor.ELASTICSEARCH_CLIENT_CONFIG_KEY}': Elasticsearch()
```
verdan, line 1403:
```suggestion
    f'extractor.es_metadata.{ElasticsearchColumnStatsExtractor.CLUSTER}': 'demo',
    f'extractor.es_metadata.{ElasticsearchColumnStatsExtractor.SCHEMA}': 'dev',
    f'extractor.es_metadata.{ElasticsearchColumnStatsExtractor.ELASTICSEARCH_CLIENT_CONFIG_KEY}': Elasticsearch()
```

verdan (changes requested):

databuilder/README.md
mgorsk1: Hm I think TableMetadata is a very diverse concept and can be applied on NoSQL DBs as well. Unless we would want a different type of resource in UI for those but I am not sure how would it differ from regular table view and whether it makes sense to make such distinction.

verdan (approved):
LGTM

mgorsk1:
@feng-tao need an ok from codeowner 🙏

feng-tao (approved):

PR #1216, opened 2021-06-10
fix: Fix db key constraints

youngyjd:
### Summary of Changes

We at Brex have databases named with numbers and underscores in it. There were exceptions when we tried to use Badges, as Badges checked key format https://github.com/amundsen-io/amundsen/blob/54f5ed6bb4dbcb0ed59182e485bd109374d549a9/databuilder/databuilder/models/badge.py#L52-#L67

Gladly, those checks haven been removed recently https://github.com/amundsen-io/amundsen/commit/aa67a0c14d191501b23ea488f1a4b9d4bbe20230#diff-1eb09b5d29f66bcea3cd4a4a23befe3c93d8e577a33de1e9d903f3fb9c1faffdL52

This PR is to change another place that checks a key should match a pattern.

### Tests

Added unit tests

### CheckList

### DIFF
--- a/databuilder/databuilder/models/schema/schema_constant.py
+++ b/databuilder/databuilder/models/schema/schema_constant.py
@@ -11,4 +11,4 @@ SCHEMA_REVERSE_RELATION_TYPE = 'SCHEMA_OF'
 DATABASE_SCHEMA_KEY_FORMAT = '{db}://{cluster}.{schema}'
 
 # pattern used to match a schema key, e.g., hive://gold.test_schema
-SCHEMA_KEY_PATTERN_REGEX = '([a-z]+://[a-zA-Z0-9_-]+).[a-zA-Z0-9_.-]+'
+SCHEMA_KEY_PATTERN_REGEX = '([a-zA-Z0-9_]+://[a-zA-Z0-9_-]+).[a-zA-Z0-9_.-]+'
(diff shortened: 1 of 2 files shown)


youngyjd:
A small one. mind taking a look? @feng-tao @allisonsuarez @dorianj

databuilder/databuilder/models/schema/schema_constant.py
```diff
@@ -11,4 +11,4 @@
 DATABASE_SCHEMA_KEY_FORMAT = '{db}://{cluster}.{schema}'
 
 # pattern used to match a schema key, e.g., hive://gold.test_schema
-SCHEMA_KEY_PATTERN_REGEX = '([a-z]+://[a-zA-Z0-9_-]+).[a-zA-Z0-9_.-]+'
+SCHEMA_KEY_PATTERN_REGEX = '([a-zA-Z0-9_]+://[a-zA-Z0-9_-]+).[a-zA-Z0-9_.-]+'
```
feng-tao, line 14: the name uri is db://cluster.schema.table, db normally is hive, snowflake, presto etc. why the db will be sth else?

feng-tao (approved):
fine with me...

databuilder/databuilder/models/schema/schema_constant.py
youngyjd: I guess the reason is that we only have snowflake but not any other data warehouse, so initially we didn't include `snowflake` into keys. Instead of constructing keys like`{db=snowflake}://{cluster=snowflake_database}.{schema=schema_name}/{table=table_name}`, we are doing `{db=snowflake_database}://{cluster=production/staging, etc.}.{schema=schema_name}/{table=table_name}`.

PR #1218, opened 2021-06-10
feat: Change Popular Table Cache as Metadata Config Variable

alldoami:
Signed-off-by: Allison Doami <adoami@chanzuckerberg.com>

### Summary of Changes

Make it an option to update the popular table cache variable to be a variable in the metadata config so that those who want a longer/shorter cache can change it.

### CheckList

### DIFF
--- a/metadata/metadata_service/config.py
+++ b/metadata/metadata_service/config.py
@@ -75,6 +75,9 @@ class Config:
     # On User detail method, these keys will be added into amundsen_common.models.user.User.other_key_values
     USER_OTHER_KEYS = {'mode_user_id'}  # type: Set[str]
 
+    # Number of seconds until popular table cache refreshes
+    POPULAR_TABLE_CACHE_EXPIRY_SEC = 11 * 60 * 60 + randint(0, 3600)
+
     # Number of minimum reader count to qualify for popular table
     POPULAR_TABLE_MINIMUM_READER_COUNT = 10  # type: int
 
--- a/metadata/metadata_service/config.py
+++ b/metadata/metadata_service/config.py
@@ -3,6 +3,7 @@
 
 import distutils.util
 import os
+from random import randint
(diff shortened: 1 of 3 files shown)


alldoami:
@dorianj Thanks for pointing out this variable!

alldoami:
Ah hm... @dorianj any suggestions on how to fix the tests? https://github.com/amundsen-io/amundsen/pull/1218/checks?check_run_id=2798523136

In order to keep it outside of the functions, I'll need the app context...

dorianj:
I'm not sure what the right way to resolve this is. One thing we could do is attach the cache instance to the object in the constructor rather than at the  file, which would require doing the caching using inline calls rather than generators.

alldoami:
@dorianj I'm not exactly sure what that means 😅 Does that mean to put the `@_CACHE.cache(` line above the `__init__`?

mgorsk1:
I think we could wrap some functions into our custom decorator, which by default would be an pass-through noop decorator and could be configured to accept beaker cache manager in config class, sth like that:

```
from beaker.cache import CacheManager
from beaker.util import parse_cache_config_options


class NoCacheConfig:
    @staticmethod
    def my_decorator(func):
        def wrapper(*args, **kwargs):
            print("Something is happening before the function is called.")
            func(*args, **kwargs)
            print("Something is happening after the function is called.")

        return wrapper

    wrapper = my_decorator


class CacheConfig:
    cache_opts = {
        'cache.type': 'file',
        'cache.data_dir': '/tmp/cache/data',
        'cache.lock_dir': '/tmp/cache/lock'
    }

    cache = CacheManager(**parse_cache_config_options(cache_opts))

    wrapper = cache.cache('my_search_func', expire=3600)

cache_on = False

if cache_on:
    config = CacheConfig
else:
    config = NoCacheConfig


class Test:
    @config.wrapper
    def test(self, x):
        pass


t = Test()
t.test(1)
```

PR #1230, opened 2021-06-15
import multiprocessing's TimeoutError explicitly

mcohen01:
The catch on L277 is actually looking for Python's builtin `TimeoutError`, but `multiprocessing.context.TimeoutError`s are still thrown from L274.


dorianj:
Please fix DCO

Should we import just multiprocessing.context and refer to it will full qualification, so as not to shadow the builtin?

Issue #1231, opened 2021-06-15
Visible focus on table column expand and collapse buttons


SamMade:
## Expected Behavior or Use Case
Visible focus should be shown when tabbing through the interactive elements of the table detail expand and collapse buttons. The buttons do get focus but with no visible indicator, the user get's lost to where they are.

## Service
frontend/amundsen_application/static/js/components/Table/styles.scss

## Possible Implementation
Not familiar as to exactly why there's a sass rule to explicitly hide focus which a comment that says TODO but removing that rule will show the focus.

## Context
This helps with accessibility

SamMade:
PR made: https://github.com/amundsen-io/amundsen/pull/1233

PR #1233, opened 2021-06-15
style: Adding back focus on table detail button

SamMade:
making it accessible so focus does appear on the expand and collapse buttons

### Summary of Changes

Removed explicit rule that hide the focus indicator

addresses issue: https://github.com/amundsen-io/amundsen/issues/1231

### Tests

no test necessary as just a style change


Golodhros:
Hi @SamMade, thanks for creating this PR!

The way we want to fix this, and why it was marked as something to fix instead or removing that, is by using https://www.npmjs.com/package/focus-visible. 

The idea is that the outline will show only when it is being accessed using a11y ways, but not on the regular operation as it will do if we merge your changes.

Is this something you want to contribute to the project?

PR #1235, opened 2021-06-15
refactor: bw-incompatible cleanup of usage models

dkunitsk:

### Summary of Changes

For the ongoing feature work, we need a feature usage model. I created a general `Usage` model (similarly to generic [Owner](https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/models/owner.py#L23)) which serves as a superclass for `ColumnReader` and `DashboardUsage`. For features, we will use `Usage` directly. 

I found that table usage had two nearly identical models: `ColumnUsageModel` and `ColumnReader` are identical except for argument names (`table_name`/`column_name` vs `table`/`column`). Naturally, I picked the model which had tests and deleted the other. `ColumnReader` also seems to predate and be in wider use than the deleted model. 

Finally, table and dashboard usage models previously created users with empty-string defaults (e.g. `'github_username': ''`). I can't see what value that'd be to anyone, so I've replaced it with the more natural form.

Nothing has been done about "Column" usage not supporting usage at the column level... Although I did remove the error raised when passing a column other than `"*"` to minimize impact since the model which was deleted used to ignored the column name instead of failing.

### Tests

New test of new `Usage` model. Updated tests of modified models. 

### CheckList


dorianj:
cc @sewardgw

dorianj (commented):
Seems super sensible to me 👍 

Can you draft the message describing to community what the bw-incompat changes are, and what they'll need to do to upgrade?

dkunitsk:
Thanks @dorianj! See draft message below:

Backwards incompatible change alert:

Redundant databuilder class [ColumnUsageModel](https://github.com/amundsen-io/amundsen/blob/3bbc4669a53526e273b5b717d2ecf028fc5233a3/databuilder/databuilder/models/column_usage_model.py#L20) was deleted in favor of [ColumnReader](https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/models/table_column_usage.py) which is in wider use and has test coverage. The only difference between them is arguments `table_name` and `column_name` are called `table` and `column`. 

This is done to remove redundancy and refactor usage models to subclass [Usage](https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/models/usage/usage.py), which implements common behavior for table and dashboard usage. A similar refactor was done for [Owner](https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/models/owner.py#L23) - e.g. see [TableOwner](https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/models/table_owner.py#L10) which subclasses it.

Note: For tables, there is still no usage at the column level, despite the misleading model names. The names were retained to reduce the impact radius.

feng-tao:
@dkunitsk Will look at tonight, but before I go through, does it work for gremlin/ neptune and Mysql store? cc @AndrewCiambrone @crazy-2020

feng-tao:
the columnReader could be the one we initially used to do query parsing with antlar to get column usage before we remove all those antlar classes.

dkunitsk:
@feng-tao thanks!! Yes, `Usage` is table and graph serializable and in general no tests have been deleted so all that good stuff should continue to work. 

Re the Lyft context: thank you! You're right there is a custom transformer that I didn't know about. Luckily it's using TableColumnUsage (which is yet a 3rd table usage model...) and that one has remained backwards compatible. 

Looking forward to your review.

databuilder/databuilder/models/feature/feature_watermark.py
```diff
@@ -18,6 +18,8 @@ class FeatureWatermark(GraphSerializable):
     """
     NODE_LABEL = 'Feature_Watermark'
 
+    TYPE_ATTR = 'watermark_type'
```
dkunitsk, line 21: this is unrelated to the `Usage` stuff. Just a minor improvement in the ongoing feature work.

databuilder/databuilder/models/dashboard/dashboard_usage.py
```diff
@@ -51,73 +41,29 @@ def __init__(self,
...
-        )
-        yield dashboard_usage_record
+            return super().create_next_node()
+        return None
 
-    def __repr__(self) -> str:
-        return f'DashboardUsage({self._dashboard_group_id!r}, {self._dashboard_id!r}, ' \
-               f'{self._email!r}, {self._view_count!r}, {self._should_create_user_node!r}, ' \
-               f'{self._product!r}, {self._cluster!r})'
+    # override superclass for customized _should_create_user_node behavior
```
crazy-2020, line 64: nit: what do you think if we move `_should_create_user_node`(defaults to true) to Usage constructor? Thanks

databuilder/databuilder/models/table_column_usage.py
```diff
@@ -6,64 +6,51 @@
...
 from databuilder.models.graph_relationship import GraphRelationship
 from databuilder.models.graph_serializable import GraphSerializable
 from databuilder.models.table_metadata import TableMetadata
 from databuilder.models.table_serializable import TableSerializable
-from databuilder.models.user import User
+from databuilder.models.usage.usage import Usage
 
 
-class ColumnReader(object):
+class ColumnReader(Usage):
```
mgorsk1, line 18: Is there a reason why we don't have explicit `ColumnReader` and `TableReader`. This is one thing that is still confusing for me - if we don't support per-column usage then why have ColumnReader at all?

databuilder/databuilder/models/table_column_usage.py
```diff
@@ -6,64 +6,51 @@
...
+                db=database,
+                cluster=cluster,
+                schema=schema,
+                tbl=table),
+            user_email=user_email,
+            read_count=read_count,
+        )
 
 
 class TableColumnUsage(GraphSerializable, TableSerializable):
```
mgorsk1, line 46: Again, the naming is confusing and I am not sure why it's not just `TableReader`

mgorsk1 (commented):
awesome idea, thanks for doing this. two small nits regarding naming, which I think is slightly ambiguous

databuilder/databuilder/models/column_usage_model.py
```diff
@@ -1,118 +0,0 @@
-# Copyright Contributors to the Amundsen project.
-# SPDX-License-Identifier: Apache-2.0
-
-from typing import Iterator, Union
```
feng-tao, line 4: make sense to delete the duplicate one... I kinda recalled now which the model is purely for csv column usage support (e.g https://github.com/amundsen-io/amundsendatabuilder/pull/173)

databuilder/databuilder/models/table_column_usage.py
feng-tao: which one is for tableReader? i think the initial implementation is to implement both column usage and table usage in one model with query parser extractor with antlar. Then we found the query parser antlar has many production issues (e.g memory OOM) which we remove the extractor and leave the models here.

databuilder/databuilder/models/table_column_usage.py
feng-tao: We plan to support in the first place (even before the project is open source)... but we found that the antlar usage extractor we have have some performance issues...

feng-tao (commented):
looks reasonable. The deleted one are quick start use case with csv extractor. Not sure if you have the chance to test the quick start flow which the column usage still shows up in the UI(it is fine if not though)...

feng-tao (approved):
lgtm from my side with a few questions (e.g will it break the quick start script if we remove the duplicated column usage model)?

databuilder/databuilder/models/table_column_usage.py
dkunitsk: I myself can't speak to the history, but going forward there's not many options except:

- rename the current classes, which would be a larger incompatible change
- add new duplicate classes doing the same thing but named more correctly. Not sure it would get much adoption and may add confusion
- wait until we do implement per-column usage inside the existing classes. In that case having `ColumnReader` also do table usage when column is not specified does make sense. 

If and when the last option happens, everything will make sense. Until then, we either have to accept the weirdness or be willing to make a bigger incompatible change.

databuilder/databuilder/models/dashboard/dashboard_usage.py
dkunitsk: This is a good idea, and would also be helpful in the generic Owner class, which has to do the same thing. If you don't mind, I will do this in a follow up.

dkunitsk:
@feng-tao have run quick start guide:
- started local neo4j (via docker-amundsen.yml)
- ran `sample_data_loader` from this branch (i modified it in this PR to stop using the redundant model)
- checked local neo4j. The usages are showing up as specified in `sample_column_usage.csv` between entities and users. 

I believe that means we're 💯

PR #1249, opened 2021-06-18
feat: Adds support of `USER_DETAIL_METHOD` to Neo4j

verdan:

### Summary of Changes

We have been using the [USER_DETAIL_METHOD](https://www.amundsen.io/amundsen/metadata/docs/configurations/#user_detail_method-optional) for some time now in atlas_proxy. And it works nicely if you are using OIDC or any 3rd party tool to get your user information, instead of ingesting this in Neo4j. 

This PR adds the same support to Neo4j Proxy. 

### Documentation

NA

### CheckList


verdan:
@feng-tao can you please help with the review?

mgorsk1:
There are significant downsides to having this approach vs syncing data with databuilder and I think if we are introducing this method for neo4j it should be well documented that caching it is suggested and might slightly improve it but imo it's not as smooth as advertised.

verdan:
Those who do not want to (or have a way to) ingest users via databuilder, and want to create users on runtime, OR add/modify any attributes of a user profile coming from the graph, need this functionality. I am not sure if I'd call it the significant downsides, but indeed there are some trade-offs of this approach and the one implementing this should aware of this. 
This pretty much also depends on the implementation of this method, as to what you are doing inside this method and how you are retrieving the user details inside this method.

mgorsk1:
What I meant is that in comparison to retrieval directly from neo4j, this is not scaling as well. If you for example have many readers connected to the table, the latency for table retrieval is then bound to the external system performance (which would most likely be handled by different team than Amundsen) and number of users connected to the table itself. I am not saying it's not needed, just that it's tricky, should be considered last resort and would use good documentation on how to avoid most common bottlenecks.

dorianj:
+1 to documenting this potential footgun if we start pointing users to it, but I don't think those docs need to block landing it because at present it's fairly undiscoverable

dorianj (approved):

PR #1254, opened 2021-06-21
feat: Adds dynamic table notices from Atlas

stswn:
### Summary of Changes

In follow up to [#957](https://github.com/amundsen-io/amundsenfrontendlibrary/pull/957) functionality of table notices is extended to allow for defining multiple dynamic notices for each table in Apache Atlas. Notices coming from Atlas are displayed together with config notices in the same fashion.

### Tests

Tests were adjusted to accommodate new data model.

### Documentation

--

### DIFF
--- a/frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
+++ b/frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
@@ -18,6 +18,7 @@ import { GetTableLineageRequest } from 'ducks/lineage/types';
 import { OpenRequestAction } from 'ducks/notification/types';
 import { UpdateSearchStateRequest } from 'ducks/search/types';
 
+import { NoticeSeverity } from 'config/config-types';
 import {
   getDescriptionSourceDisplayName,
   getMaxLength,
@@ -28,6 +29,7 @@ import {
   issueTrackingEnabled,
   isTableListLineageEnabled,
   notificationsEnabled,
+  sortNotices,
 } from 'config/config-utils';
 
 import BadgeList from 'features/BadgeList';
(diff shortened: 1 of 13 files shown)


stswn:
@mgorsk1

mgorsk1 (changes requested):
backend-wise lgtm with just one nit - we might need to release and bump amundsen-common in separate MR

can you please extend unit tests and fill out DCO https://github.com/apps/dco?

cc @Golodhros for frontend stuff

frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -401,12 +410,12 @@ export class TableDetail extends React.Component<
           </header>
           <div className="column-layout-1">
             <aside className="left-panel">
-              {!!tableNotice && (
+              {notices.map((notice) => (
```
Golodhros, line 413: What if notices is empty?

At this point, I would extract a component here, something like <NoticeList /> and move all the logic inside.

frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -351,6 +353,13 @@ export class TableDetail extends React.Component<
         ResourceType.table,
         `${data.cluster}.${data.database}.${data.schema}.${data.name}`
       );
+      const dataNotices = data.notices.map((n) => ({
+        severity: n.severity as NoticeSeverity,
+        messageHtml: n.message_html,
+      }));
```
Golodhros, line 359: This doesn't apply to dashboards? Only for tables?

allisonsuarez:
I think is a good idea to make notices dynamic, however I believe we should make sure this model is reusable for other entities and that we add the metadata proxy method as an interface in base proxy for others to use. Could we maybe open an RFC for this so it can be opened up for discussion?

PR #1260, opened 2021-06-23
fix: Fixes the Amundsen Builds

verdan:

### Summary of Changes

We are using the master branch of the GitHub action responsible for the docs deployment. 
The latest version of that action breaks the document's build. This PR tries to check if changing the version actually
fixes the build. 

### CheckList

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-mkdocs==1.1
-mkdocs-material==4.6.3
-mkdocs-redirects==1.0.0
+mkdocs>=1.1
+mkdocs-material>=4.6.3
+mkdocs-redirects>=1.0.0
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,3 +1,3 @@
-mkdocs==1.1
-mkdocs-material==4.6.3
-mkdocs-redirects==1.0.0
+mkdocs>=1.1
```
mgorsk1, line 1: if those versions are not working anymore should we use `>=` but with higher (currently working) versions?

verdan: Indeed will use the current versions. This was only to verify the issue.

mgorsk1 (approved):
lgtm

PR #1264, opened 2021-06-23
refactor: remove hard pins in requirements-common.txt

dkunitsk:
Signed-off-by: Dmitriy Kunitskiy <dkunitskiy@lyft.com>

### Summary of Changes
The current setup introduces hard pins for `amundsen-common` and the hard pins may also be inconvenient for custom setups of the amundsen microservices. We are attempting to roll forward with this removal 
After discussing with @szczeles and @feng-tao, we've decided to try removing hard pins completely and see what kind of trouble we run into. 

I've also deleted some dependencies which appear to no longer be used. 

### Tests

Rebuilt containers (following https://github.com/amundsen-io/amundsen/blob/master/docs/developer_guide.md#building-local-changes) and ran the quick start guide with sample data. Don't see any issues so far! 

### CheckList
Make sure you have checked **all** steps below to ensure a timely review.
- [ ] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).
- [ ] PR includes a summary of changes.
- [ ] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [ ] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public functions and the classes in the PR contain docstrings that explain what it does

PR #1267, opened 2021-06-24
refactor: undo the great unpinning

dkunitsk:

### Summary of Changes

This formally ends [The Great Unpinning](https://github.com/amundsen-io/amundsen/pull/1264). In effect, I realized it does not solve the problem of `amundsen-common` as it still results in this simple library requiring many unrelated dependencies. Additionally, the loss of reproducible builds introduced by unpinned service dependencies is possibly not worth the benefit.

Instead, I propose to keep amundsen-common separate with its small requirements set. Why is this okay? Because, when `amundsen-common` is used inside databuilder/frontend/metadata/search, the `install_requires` of `amundsen-common` will be completely irrelevant. In other words, the pain of desynchronized dependencies will not return and the benefits of @mgorsk1's efforts in https://github.com/amundsen-io/amundsen/pull/1163 are maintained. 

One element of the earlier PR I have kept is the removal of outdated dependencies:
- itsdangerous
- six
- MarkupSafe
- aniso8601

### CheckList

### DIFF
--- a/common/setup.py
+++ b/common/setup.py
@@ -44,7 +44,7 @@ setup(
         'marshmallow3-annotations>=1.0.0'
     ],
     extras_require={
-        'dev': requirements_dev
+        'all': requirements_dev
     },
     python_requires=">=3.6",
     package_data={'amundsen_common': ['py.typed']},
--- a/common/setup.py
+++ b/common/setup.py
@@ -4,20 +4,13 @@ import os
 
 from setuptools import find_packages, setup
 
-__version__ = '0.16.0'
+__version__ = '0.17.0'
 
-requirements_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'requirements-common.txt')
-with open(requirements_path) as requirements_file:
-    requirements_common = requirements_file.readlines()
 
 requirements_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'requirements-dev.txt')
 with open(requirements_path) as requirements_file:
     requirements_dev = requirements_file.readlines()
 
-# avoid circular references
-requirements_common = [r for r in requirements_common if not r.startswith('amundsen-common')]
-
-all_deps = requirements_common + requirements_dev
 
 setup(
     name='amundsen-common',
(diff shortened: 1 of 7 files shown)


common/setup.py
```diff
@@ -29,9 +22,28 @@
...
+        # e.g.
+        # Python FooBar package for foobaring
+        # pyfoobar>=1.0, <2.0
+        #
+        # This will allow for any consuming projects to use this library as
+        # long as they have a version of pyfoobar equal to or greater than 1.x
+        # and less than 2.x installed.
+        'Flask>=1.0.2',
+        'attrs>=19.0.0',
+        'marshmallow>=3.0',
```
mgorsk1, line 43: As we discussed I really wouldn't want to get back to this situation. One of the most notorious things that happened and triggered dependency unification is marshmallow mismatch and it was brought up by community quite often. And the way it's defined in requirements-common.txt is pretty much ok (it's not hard pin but a fair range).

Right now it's perfectly in sync but it can easily become inconsistent. I would rather filter out marshmallow from requirements-common.txt which I agree is uglier solution but at least keeps things consistent across the board.

mgorsk1 (changes requested):

common/setup.py
dkunitsk: Ah I understand. However I think the point I had made in the PR description was missed. This list in common/setup.py does *not* need to be in sync with `requirements-common.txt`. It is okay for the requirements listed here to be a looser set. When `amundsen-common` is used in search or frontend or etc, the looser requirements here will not matter. 

Since I suspect the dependencies in requirements-common.txt will be advanced upwards, the looser requirements here will not have an effect on which versions are used in services.

dkunitsk: In other words, the actual requirements listed here in `common/setup.py` are the library's own requirements which only matter when the library is used outside of amundsen, for example in Airflow etc.

common/setup.py
```diff
@@ -29,9 +22,28 @@
...
+        # Generally packages listed here are pinned to a major version range.
+        #
+        # e.g.
+        # Python FooBar package for foobaring
+        # pyfoobar>=1.0, <2.0
+        #
+        # This will allow for any consuming projects to use this library as
+        # long as they have a version of pyfoobar equal to or greater than 1.x
+        # and less than 2.x installed.
+        'Flask>=1.0.2',
```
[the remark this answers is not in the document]
dkunitsk: Great point. Let me look into this. Would love to get rid of Flask here, or at the very least move it into an extra_requires.

common/setup.py
```diff
@@ -29,9 +22,28 @@
     maintainer='Amundsen TSC',
     maintainer_email='amundsen-tsc@lists.lfai.foundation',
     packages=find_packages(exclude=['tests*']),
-    install_requires=requirements_common,
+    install_requires=[
+        # Packages in here should rarely be pinned. This is because these
```
dkunitsk, line 26: note, I didn't author this comment, I took it from the earlier version of `amundsen-common`

common/setup.py
```diff
@@ -29,9 +22,28 @@
...
+        # This will allow for any consuming projects to use this library as
+        # long as they have a version of pyfoobar equal to or greater than 1.x
+        # and less than 2.x installed.
+        'Flask>=1.0.2',
+        'attrs>=19.0.0',
+        'marshmallow>=3.0',
+        'marshmallow3-annotations>=1.0.0'
+    ],
     extras_require={
-        'all': all_deps,
```
mgorsk1, line 34: I guess for consistency we could still keep `all` where all would be dev + install_requires

common/setup.py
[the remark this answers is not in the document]
dkunitsk: Sadly, although it's not used by amundsen itself, Flask is used to add logging. We use it internally at Lyft and I imagine so do others. It would be a backwards incompatible change to add it as an extra_requires. I am happy to do that and I agree it makes sense, but I suggest we take that change out separately.

allisonsuarez (approved):
My understanding is there's enough consensus on this route to try it out. we can always go back if not, but I think this is the safest way out of the situation.

mgorsk1 (approved):

PR #1274, opened 2021-06-24
feat: Making notices more general

elenaboal:
### Summary of Changes

Added wildcard functionality to notices (ie if we wanted to add a notice to every table in the shadow schema, could do that with *.*.shadow.*) and dynamic HTML messages (ie if we wanted to have each of those notices have a different message depending on which table it was)

### Why
Part of the follow up work for the Notices feature detailed on the RFC: https://github.com/amundsen-io/rfcs/blob/master/rfcs/029-resource-notices.md#future-possibilities
Mentioned on the original PR: https://github.com/amundsen-io/amundsenfrontendlibrary/pull/957#issuecomment-802943900

### Tests

Added notices tests to test the functionality of the wildcard and the dynamic message.


frontend/amundsen_application/static/js/config/config-types.ts
```diff
@@ -151,7 +151,7 @@ export enum NoticeSeverity {
 }
 export interface NoticeType {
   severity: NoticeSeverity;
-  messageHtml: string;
+  messageHtml: string | ((resourceName) => string);
```
Golodhros, line 154: I think this would work better:
```suggestion
  messageHtml: string | (string) => string;
```

frontend/amundsen_application/static/js/config/index.spec.ts
```diff
@@ -68,42 +68,248 @@ describe('getDisplayNameByResource', () => {
...
+            messageHtml: 'testMessage',
+          },
+        };
+        const expected = 'testMessage';
+        const notice = ConfigUtils.getResourceNotices(
+          ResourceType.table,
+          'testName'
+        );
+        const actual = notice && notice.messageHtml;
+        expect(actual).toEqual(expected);
```
Golodhros, line 86: Nit: space before expect

frontend/amundsen_application/static/js/config/index.spec.ts
```diff
@@ -68,42 +68,248 @@ describe('getDisplayNameByResource', () => {
...
+    });
 
-    it('returns the notice', () => {
-      const expected = 'testMessage';
-      const notice = ConfigUtils.getResourceNotices(
-        ResourceType.table,
-        'testName'
-      );
-      const actual = notice && notice.messageHtml;
+    describe('when there is a simple wildcard in gold position', () => {
```
Golodhros, line 90: 'cluster' instead of 'gold'

Golodhros: I would group all these under a 
`describe('when there are wildcards', ()...`

frontend/amundsen_application/static/js/config/index.spec.ts
```diff
@@ -68,42 +68,248 @@ describe('getDisplayNameByResource', () => {
...
+        const expected = 'testMessage';
+        const notice = ConfigUtils.getResourceNotices(
+          ResourceType.table,
+          'gold.hive.core.fact_rides'
+        );
+        const actual = notice && notice.messageHtml;
 
-      expect(actual).toEqual(expected);
+        expect(actual).toEqual(expected);
+      });
```
Golodhros, line 106: Let's add a test for a case when it doesn't match too.

frontend/amundsen_application/static/js/config/index.spec.ts
```diff
@@ -68,42 +68,248 @@ describe('getDisplayNameByResource', () => {
...
 
-    it('returns false', () => {
-      const expected = false;
-      const actual = ConfigUtils.getResourceNotices(
-        ResourceType.table,
-        'testNameNoThere'
-      );
+        expect(actual).toEqual(expected);
+      });
+    });
```
Golodhros, line 126: Same for non-matching notice

frontend/amundsen_application/static/js/config/index.spec.ts
```diff
@@ -68,42 +68,248 @@ describe('getDisplayNameByResource', () => {
...
-  });
 
-  describe('when there is no notice', () => {
-    AppConfig.resourceConfig[ResourceType.table].notices = {
-      testName: {
-        severity: NoticeSeverity.WARNING,
-        messageHtml: 'testMessage',
-      },
-    };
+    describe('when there is a simple wildcard in hive position', () => {
```
Golodhros, line 109: datasource instead of 'hive'

frontend/amundsen_application/static/js/config/index.spec.ts
```diff
@@ -68,42 +68,248 @@ describe('getDisplayNameByResource', () => {
...
+          'gold.hive.core.fact_rides'
+        );
+        const actual = notice && notice.messageHtml;
+
+        expect(actual).toEqual(expected);
+      });
+    });
+
+    describe('when there are 2 notices that match', () => {
+      it('returns notice', () => {
```
Golodhros, line 186:
```suggestion
      it('returns the last matched notice', () => {
```

frontend/amundsen_application/static/js/config/index.spec.ts
```diff
@@ -68,42 +68,248 @@ describe('getDisplayNameByResource', () => {
...
+          'gold.hive.core.fact_rides'
+        );
+        const actual = notice && notice.messageHtml;
+
+        expect(actual).toEqual(expected);
+      });
+    });
+
+    describe('when there is a notice with a dynamic message', () => {
+      it('returns notice', () => {
```
Golodhros, line 251:
```suggestion
      it('returns notice with dynamic message', () => {
```

frontend/amundsen_application/static/js/config/index.spec.ts
```diff
@@ -68,42 +68,248 @@ describe('getDisplayNameByResource', () => {
...
+          },
+        };
+        const expected = false;
+        const actual = ConfigUtils.getResourceNotices(
+          ResourceType.table,
+          'testNameNoThere'
+        );
+        expect(actual).toEqual(expected);
+      });
     });
```
Golodhros, line 313: Would all this work for dashboards too?
Could you add some extra tests?

Golodhros (changes requested):
Overall looks good!

Let's update our docs on https://github.com/amundsen-io/amundsen/blob/main/frontend/docs/application_config.md#notices

frontend/amundsen_application/static/js/config/config-utils.ts
```diff
@@ -77,7 +80,42 @@ export function getResourceNotices(
   const { notices } = AppConfig.resourceConfig[resourceType];
 
   if (notices && notices[resourceName]) {
-    return notices[resourceName];
+    const thisNotice = notices[resourceName];
+    if (typeof thisNotice.messageHtml !== 'string') {
```
Golodhros, line 84: I think it would be better do do a positive check like:
```suggestion
    if (typeof thisNotice.messageHtml === 'function') {
```
I will be more safe and explicit

frontend/amundsen_application/static/js/config/config-utils.ts
```diff
@@ -77,7 +80,42 @@ export function getResourceNotices(
...
+          if (i === decomposedKey.length - 1) {
+            noticeFromWildcard = notices[key];
+          }
+          continue;
+        }
+        break;
+      }
+    });
+    if (noticeFromWildcard) {
+      if (typeof noticeFromWildcard.messageHtml !== 'string') {
```
Golodhros, line 112: same here

frontend/amundsen_application/static/js/config/config-utils.ts
```diff
@@ -77,7 +80,45 @@ export function getResourceNotices(
   const { notices } = AppConfig.resourceConfig[resourceType];
 
   if (notices && notices[resourceName]) {
-    return notices[resourceName];
+    const thisNotice = notices[resourceName];
+    if (typeof thisNotice.messageHtml === 'function') {
+      thisNotice.messageHtml = thisNotice.messageHtml(resourceName);
+    }
+    return thisNotice;
```
Golodhros, line 87: Sorry for the late comment, but I think we could do something like:
```
const withComputedMessage = (notice, resourceName) => {
if (typeof notice.messageHtml === 'function') {
      notice.messageHtml = notice.messageHtml(resourceName);
    }
    return notice;
}
```
And use it on line 115 as well.

frontend/docs/application_config.md
```diff
@@ -183,7 +183,7 @@ A notice is a small box with an icon and a message containing HTML markup (like
   <img src='img/notices-alert-table.png' width='50%' />
 </figure>
 
-To set them up, we'll use the current configuration objects for the resources. For example, if company X wants to deprecate the use of one table or dashboard, they can opt to add new notices in their configuration file:
+To set them up, we'll use the current configuration objects for the resources. In the event that we want to add the same notice to every table that follows a particular pattern, we use a wildcard character, *, for pattern matching. In addition, we can have dynamic HTML messages to allow for notices to change their message based on what table it is. For example, if company X wants to deprecate the use of one table or dashboard, they can opt to add new notices in their configuration file:
```
Golodhros, line 186:
```suggestion
To set them up, we'll use the current configuration objects for the resources. In the event that we want to add the same notice to every table that follows a particular pattern, we use a wildcard character, *, for pattern matching. In addition, we can have dynamic HTML messages to allow for notices to change their message based on what table it is. 

For example, if company X wants to deprecate the use of one table or dashboard, they can opt to add new notices in their configuration file:
```

frontend/docs/application_config.md
```diff
@@ -211,6 +211,64 @@ To set them up, we'll use the current configuration objects for the resources. F
...
+            severity: NoticeSeverity.WARNING,
+            messageHtml: `This dashboard is deprecated`,
+          },
+      },
+    },
+
+  },
+```
+
+The above code will show a notice with a red exclamation icon whenever a final user visits any table within the specified cluster, database, and schema or any dashboard within the specified product, cluster, and groupname.
```
Golodhros, line 238: What we were doing of introducing the code after showing it might work with one example, but now it seems we would better introduce the code **before** we show it. Try something like:

If you want to target several tables at once, you can use wildcards as showed below:
//code
This will show a notice with a red...

frontend/docs/application_config.md
```diff
@@ -211,6 +211,64 @@ To set them up, we'll use the current configuration objects for the resources. F
...
+```
+  resourceConfig: {
+    [ResourceType.table]: {
+      ... //Table Resource Configuration
+      notices: {
+          "<CLUSTER>.<DATABASE>.<SCHEMA>.*": {
+            severity: NoticeSeverity.ALERT,
+            messageHtml: (resourceName) => {
+              const [cluster, datasource, schema, table] = resourceName.split('.');
+              return `This schema is deprecated, please use <a href="https://amundsen.lyft.net/table_detail/${cluster}/${datasource}/SCHEMA/${table}">this table instead</a>`;
```
Golodhros, line 249:
```suggestion
              return `This schema is deprecated, please use <a href="https://amundsen.<company>.net/table_detail/${cluster}/${datasource}/SCHEMA/${table}">this table instead</a>`;
```

elenaboal:
remade pr here: https://github.com/amundsen-io/amundsen/pull/1280

PR #1280, opened 2021-06-28
feat: Making notices more general (real)

elenaboal:
### Summary of Changes

Added wildcard functionality to notices (ie if we wanted to add a notice to every table in the shadow schema, could do that with ..shadow.*) and dynamic HTML messages (ie if we wanted to have each of those notices have a different message depending on which table it was)

### Tests

Added notices tests to test the functionality of the wildcard and the dynamic message.

### Documentation

Added documentation in frontend/docs/application_config.md

### DIFF
--- a/frontend/amundsen_application/static/js/config/config-utils.ts
+++ b/frontend/amundsen_application/static/js/config/config-utils.ts
@@ -114,7 +114,8 @@ export function getResourceNotices(
       }
     });
     if (hasNotice) {
-      const noticeFromWildcard: NoticeType = wildcardNoticesArray[0];
+      // const noticeFromWildcard: NoticeType = wildcardNoticesArray[0];
+      const [noticeFromWildcard] = wildcardNoticesArray;
       return withComputedMessage(noticeFromWildcard, resourceName);
     }
   }
--- a/frontend/amundsen_application/static/js/config/config-utils.ts
+++ b/frontend/amundsen_application/static/js/config/config-utils.ts
@@ -13,7 +13,16 @@ import { ResourceType } from '../interfaces';
 
 export const DEFAULT_DATABASE_ICON_CLASS = 'icon-database icon-color';
 export const DEFAULT_DASHBOARD_ICON_CLASS = 'icon-dashboard icon-color';
+const WILDCARD_SIGN = '*';
+const RESOURCE_SEPARATOR = '.';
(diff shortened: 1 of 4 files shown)


Golodhros (approved):

Golodhros:
@dorianj check this out!

This was something you suggested on my first PR and that we found it was really useful. We would love to hear your thoughts on the specific implementation.

dorianj:
This is cool! Definitely great to have these be more flexible.

Rather than using strings with wildcards, I believe that we should allow the use of regexes. Then the eval code can branch on type, rather than the more complicated inspection of the string contents.

The trouble is that you can't use them in object literals in JS. You _can_ however use them in a Map. I'm not sure how to square this circle -- any thoughts?

Less importantly: i appreciate the completeness of tests. However, it feels a bit too much to test every permutation of position of wildcards. This might go away if regexes were implemented, since the test would be more generic.

Golodhros:
> This is cool! Definitely great to have these be more flexible.

Yeah, we needed to do some batch notices and the current solution wasn't scaling.

> Rather than using strings with wildcards, I believe that we should allow the use of regexes. Then the eval code can branch on type, rather than the more complicated inspection of the string contents.
> 
> The trouble is that you can't use them in object literals in JS. You _can_ however use them in a Map. I'm not sure how to square this circle -- any thoughts?

I agree the string matching logic is not great, luckily we only have four elements to match. I don't see a way of creating a map the way we have it set up though.
 
> Less importantly: i appreciate the completeness of tests. However, it feels a bit too much to test every permutation of position of wildcards. This might go away if regexes were implemented, since the test would be more generic.

Maybe it would be an opportunity to do some parametrized tests. 

Here we prioritized not making a breaking change, but that said, we didn't see many other options anyways. Would you be OK with us going with this solution at this time? Can you live with this?

frontend/amundsen_application/static/js/config/config-utils.ts
```diff
@@ -77,7 +86,37 @@ export function getResourceNotices(
...
+            wildcardNoticesArray[0] = notices[key];
+            hasNotice = true;
+          }
+          continue;
+        }
+        break;
+      }
+    });
+    if (hasNotice) {
+      const noticeFromWildcard: NoticeType = wildcardNoticesArray[0];
```
Golodhros, line 117: Doing this will fix the betterer error:
```
      const [noticeFromWildcard] = wildcardNoticesArray;
```

PR #1283, opened 2021-06-29
!refactor: Updates the Amudnsen Helm chart to make it work with latest flaskoidc

verdan:
### CheckList


feng-tao:
cc @mgorsk1 hey @mgorsk1 could you take a look? I didn't follow the latest OIDC config change.

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -30,6 +30,43 @@ tolerations: []
...
+  #  flaskApp.class -- 'FlaskOIDC' if you want to enable OIDC
+  class: ""
+
+# To enable OIDC,
+#     - please make sure to fill in the mandatory variables under 'envVars' for each service.
+#     - please make sure set frontend.config.class: amundsen_application.oidc_config.OidcConfig
+# Everything under oidc.configs will be parsed as environment variables for each service.
+# Following variables must be set under `envVars` for each service.
+# - FLASK_OIDC_CLIENT_ID
+# - FLASK_OIDC_CLIENT_SECRET
```
mgorsk1, line 46: would it make sense to pull client ID and/or secret from k8s secrets? otherwise users will need to provide sensitive values in values.yaml which might be tricky if you are trying to store your chart/deployment in git

something like:
```
containers:
  - name: frontend
    envFrom:
      - secretRef:
          name: frontend-oidc-client-secret
```

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -83,6 +124,12 @@ search:
   # search.podAnnotations -- Search pod specific annotations.
   podAnnotations: {}
 
+  # Everything set under "envVars" during helm install/upgrade will be parsed as environment variable for this service.e
+  envVars:
+    # No need to explicitly add the variables in values file or in configmap. You can pass these variables runtime
+    FLASK_OIDC_CLIENT_ID: ""
```
mgorsk1, line 130: it doesn't make sense to set those at all if users are not using OIDC

mgorsk1 (commented):
few comments

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -30,6 +30,43 @@ tolerations: []
 # podAnnotations -- amundsen application wide configuration of podAnnotations. This applies to search, metadata, frontend and neo4j. Elasticsearch has it's own configuation properties for this. [ref](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
 podAnnotations: {}
 
+# Optional Flask App wrapper to wrap your Amundsen Microservices
+flaskApp:
+  #  flaskApp.module -- 'flaskoidc' if you want to enable OIDC
+  module: ""
```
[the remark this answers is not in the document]
verdan: We do need this separately. We don't always use these for the OIDC, and sometimes use other wrappers for the Amundsen services. I'd like to keep this as is.

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -30,6 +30,43 @@ tolerations: []
...
+  #  flaskApp.class -- 'FlaskOIDC' if you want to enable OIDC
+  class: ""
+
+# To enable OIDC,
```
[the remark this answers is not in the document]
verdan: makes sense. I will update this.

amundsen-kube-helm/templates/helm/values.yaml
verdan: The reason I did put it here is to document it properly as to where/how to set these variables. 
I'll be moving these OIDC related stuff under oidc property, and that will remove the confusion.

amundsen-kube-helm/templates/helm/values.yaml
verdan: excuse my lack of knowledge of k8s, but I am not sure how that would work with the pipelines. If we are injecting these variables via CI/CD or GitHub actions, would we still be able to create those secrets in K8s containers?

amundsen-kube-helm/templates/helm/values.yaml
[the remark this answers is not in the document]
verdan: @mgorsk1 fixed.

amundsen-kube-helm/templates/helm/values.yaml
verdan: I think this is more on the case-to-case scenario. For example, we used the Github secrets at one time to inject these values for one of our instances. thoughts?

amundsen-kube-helm/templates/helm/values.yaml
verdan: fixed.

amundsen-kube-helm/templates/helm/values.yaml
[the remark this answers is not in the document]
sewardgw: Would it make sense to use an if statement to set some overridable defaults if OIDC is enabled?

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -31,23 +31,38 @@ tolerations: []
...
+  enabled: true
+  frontend:
+    client_id: ""
+    client_secret: ""
+  metadata:
+    client_id: ""
+    client_secret: ""
+  search:
+    client_id: ""
+    client_secret: ""
```
sewardgw, line 63: Is there a scenario where these services have different values (if not, can these be consolidated?)?

amundsen-kube-helm/templates/helm/values.yaml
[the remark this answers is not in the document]
verdan: @sewardgw you mean this one: https://github.com/amundsen-io/amundsen/pull/1283/files#diff-5b9eab6d0443b0adc6218aa7c294d6fda6fc2d22f33c68d8a47437b3b2ed1013R26

amundsen-kube-helm/templates/helm/values.yaml
verdan: It is a good practice to keep secrets separate for each service i.e., to create a new application for each service under oidc provider.

amundsen-kube-helm/templates/helm/values.yaml
mgorsk1: indeed it's a case to case scenario, we can leave it be for now

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -30,6 +30,58 @@ tolerations: []
...
+  #   - everything under oidc.configs will be parsed
+  #   - flaskApp.module will be set as 'flaskoidc'
+  #   - flaskApp.class will be set as 'FlaskOIDC'
+  enabled: true
+  frontend:
+    client_id: ""
+    client_secret: ""
+  metadata:
+    client_id: ""
+    client_secret: ""
```
mgorsk1, line 60: I would only leave comment not to paste secret directly here but pass it from env variables when running `helm` like 
```
`--set oidc.metadata.client_secret=$OIDC_METADATA_CLIENT_SECRET`
```

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -131,6 +202,9 @@ metadata:
   # metadata.podAnnotations -- Metadata pod specific annotations.
   podAnnotations: {}
 
+  # Everything set under "envVars" during helm install/upgrade will be parsed as environment variable for this service.e
```
mgorsk1, line 205: typo in `service.e` ?

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -131,6 +202,9 @@ metadata:
...
+  envVars:
```
mgorsk1, line 206: shouldn't this be `envVars: {}` ?

amundsen-kube-helm/templates/helm/values.yaml
```diff
@@ -203,6 +260,9 @@ frontEnd:
   # frontEnd.podAnnotations -- Frontend pod specific annotations.
   podAnnotations: {}
 
+  # Everything set under "envVars" during helm install/upgrade will be parsed as environment variable for this service.
+  envVars:
```
mgorsk1, line 264: shouldn't it be `envVars: {}` ?

mgorsk1 (approved):

PR #1289, opened 2021-06-30
fix: fixes the popular dashboard issue

verdan:

### Summary of Changes

This makes the popular resources endpoint backward compatible. so in case if the function is not implemented by any proxy, this will call the popular tables function instead. 

### CheckList

### DIFF
--- a/metadata/metadata_service/proxy/atlas_proxy.py
+++ b/metadata/metadata_service/proxy/atlas_proxy.py
@@ -1674,4 +1674,4 @@ class AtlasProxy(BaseProxy):
                               num_entries: int,
                               resource_types: List[str],
                               user_id: Optional[str] = None) -> Dict[str, List]:
-        pass
+        raise NotImplementedError
--- a/metadata/metadata_service/api/popular_resources.py
+++ b/metadata/metadata_service/api/popular_resources.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from http import HTTPStatus
-from typing import Dict, Iterable, List, Mapping, Optional, Union
+from typing import Dict, Iterable, Mapping, Optional, Union
 
 from amundsen_common.entity.resource_type import ResourceType
 from amundsen_common.models.dashboard import DashboardSummarySchema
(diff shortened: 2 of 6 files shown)


verdan:
@mgorsk1 FYI

metadata/metadata_service/proxy/atlas_proxy.py
```diff
@@ -1674,4 +1674,4 @@ def get_popular_resources(self, *,
                               num_entries: int,
                               resource_types: List[str],
                               user_id: Optional[str] = None) -> Dict[str, List]:
-        pass
+        raise NotImplementedError
```
mgorsk1, line 1677: If we want to base the logic on this exception then we need to:
1. remove `@abstractmethod` decorator
2. remove dummy implementation of the method from any proxy that doesn't support this (like AtlasProxy)

Otherwise:
1. Class won't be instantiated if method decorated with `@abstractmethod` is not overriden (`TypeError: Can't instantiate abstract class Actual with abstract methods` will be raised) 

Example:
```
import abc

class Test(abc.ABC):
    @abc.abstractmethod
    def x(self):
        raise NotImplementedError

class Actual(Test):
    def y(self):
        pass

x = Actual()
```

mgorsk1 (changes requested):

metadata/metadata_service/proxy/atlas_proxy.py
verdan: @mgorsk1  Sure, this is another way of implementing things, but the way I implemented is not breaking anything, and I don't think this is wrong in any way. 

I want to make sure that each proxy has the method visible so as not to miss, and visible all the times that this needs to be implemented. 

```python
import abc

class Test(abc.ABC):
    @abc.abstractmethod
    def x(self):
        raise NotImplementedError

class Actual(Test):
    def y(self):
        pass

    def x(self):
        raise NotImplementedError

o = Actual()
```

mgorsk1: Sure, that makes sense as long as this PR also implements this method as not an abstract one for each of the proxies. Otherwise it's not really fixing anything in terms of backwards compatibility.

verdan: Sorry, I don't get it what you said. We already override this method in each proxy.

mgorsk1 (approved):

PR #1304, opened 2021-07-02
docs: databuilder/README: add es version

jroof88:
### Documentation

databuilder will only work with elasticsearch version 6.x. I tried using es 7.10 and it threw an error because es 7 removes mapping types (https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html) and amundsen uses these in the index declaration templates (ex: https://github.com/amundsen-io/amundsen/blob/main/common/amundsen_common/models/index_map.py\#L13-L114). This commit adds a single line to the databuilder README that says the elasticsearch requirements. This line is already in the search_service README so it makes sense to have in the databuilder README as well.

### CheckList

### DIFF
--- a/databuilder/README.md
+++ b/databuilder/README.md
@@ -12,6 +12,7 @@ For information about Amundsen and our other services, visit the [main repositor
 
 ## Requirements
 - Python >= 3.6.x
+- elasticsearch 6.x (currently it doesn't support 7.x)
 
 ## Doc
 - https://www.amundsen.io/amundsen/


feng-tao (approved):
Thanks for the contribution. Though we should consider moving to 7.x cc @allisonsuarez

feng-tao (approved):

jroof88:
@feng-tao I bet it would pretty easy to support both with different `INDEX_MAPPING` strings for `6.x` vs `7.x`

mgorsk1:
+1 to compatibility for both ES 6 and 7. 6 is still maintained and extended by Elastic. It is popular version amongst users

PR #1305, opened 2021-07-02
feat: search_service/config: fail on bad AWS ES config

jroof88:
When trying to set up the AWS Elasticsearch Config for the search service, if you don't have all the nessescary configuration values populated (host, region, credentials) the service will still set up properly but the client will be None. This commit changes the flow AwsSearchConfig to fail if all the values aren't populated which will result in the service not starting up and throwing this error properly that is a lot easier to debug. Without this, it is a bit of a rabbit chase to figure out why you can't connect to AWS ES.

### Summary of Changes

When trying to set up the AWS Elasticsearch Config for the search service, if you don't have all the nessescary configuration values populated (host, region, credentials) the service will still set up properly but the client will be None. This commit changes the flow `AwsSearchConfig` to fail if all the values aren't populated which will result in the service not starting up and throwing this error properly that is a lot easier to debug. Without this, it is a bit of a rabbit chase to figure out why you can't connect to AWS ES.

### Tests

I did not add any tests since this is a small control flow change and there isn't a `config_test.py`. If contributors feel like I should add one let me know and I can!

### Documentation

I added a blurb to the comment for `AwsSearchConfig` noting that you must have all of the configuration values populated for the start up will fail.

### CheckList

### DIFF
--- a/search/search_service/config.py
+++ b/search/search_service/config.py
@@ -95,6 +95,10 @@ class AwsSearchConfig(LocalConfig):
 
     If you run Amundsen on Kubernetes use IAM roles for service accounts
     (https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).
+
+    You MUST have the environment variables AWS_REGION and PROXY_ENDPOINT populated and boto3 must be able to get
+    credentials out of whatever environment you are running in. If any of these values turn out to be empty the config
+    will fail and the service will not start up.
     """
     import boto3
     from elasticsearch import Elasticsearch, RequestsHttpConnection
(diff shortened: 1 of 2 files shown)


feng-tao (approved):

feng-tao:
@jroof88 could you fix the DCO (`git commit --amend --no-edit --signoff`) ?

jroof88:
> @jroof88 could you fix the DCO (`git commit --amend --no-edit --signoff`) ?

Neat! Done!

jroof88:
Looks like there actually is a test out there for this. Fixing.

jroof88:
@feng-tao Tests are all cleaned up. Had to add an ENV var to `pytest` configuration. This is ready for another pass.

feng-tao:
thanks @jroof88 !

bbsbb:
@jroof88 - Can you elaborate on the intended behavior here? As far as I can see(and test), with this implementation any import of the config module would raise an exception if AWS_REGION is not set or credentials are missing, even if there is no intention to use AWS at all?

madison-ookla:
> @jroof88 - Can you elaborate on the intended behavior here? As far as I can see(and test), with this implementation any import of the config module would raise an exception if AWS_REGION is not set or credentials are missing, even if there is no intention to use AWS at all?

Running into this exact issue myself. This seems like problematic behavior, as any non-AWS based setup is now broken by default.

verdan:
> @jroof88 - Can you elaborate on the intended behavior here? As far as I can see(and test), with this implementation any import of the config module would raise an exception if AWS_REGION is not set or credentials are missing, even if there is no intention to use AWS at all?

@jroof88 , I am having this issue as well. Can you please help fix this?

mgorsk1:
I propose to just revert this PR. I don't think that there is much that can be improved over previous iteration - when using AWS config you `need` to make sure that you have set up every env variable (hence `all(...)` condition). 

The way Python classes work is that if the code is not within class function it's always executed. That's why we cannot build such  strict/demanding flows in config classes.

This doc:
```
  You MUST have the environment variables AWS_REGION and PROXY_ENDPOINT populated and boto3 must be able to get
  credentials out of whatever environment you are running in. If any of these values turn out to be empty the config
  will fail and the service will not start up.
```
is true not just for the config but for amundsen in general.

feng-tao:
sorry didn't catch this issue during review, let's revert the commit to unblock.

feng-tao:
also the change should handle if you use AwsSearchConfig or no?

PR #1316, opened 2021-07-06
fix: search-oidc-config

petehanssens:
### Summary of Changes

OIDC config added to Dockerfile.search.public

### Tests

I manually built the image and tested it in my own environment

### Documentation

None but will be writing a blog post on okta OIDC setup

### CheckList


Dockerfile.search.public
```diff
@@ -8,3 +8,21 @@ COPY requirements-common.txt /app/requirements-common.txt
...
+
+ENV FLASK_APP_MODULE_NAME flaskoidc
+ENV FLASK_APP_CLASS_NAME FlaskOIDC
+ENV FLASK_OIDC_WHITELISTED_ENDPOINTS status,healthcheck,health
+ENV SQLALCHEMY_DATABASE_URI sqlite:///sessions.db
+
+# You will need to set these environment variables in order to use the oidc image
+# FLASK_OIDC_CLIENT_SECRETS - a path to a client_secrets.json file
+# FLASK_OIDC_SECRET_KEY - A secret key from your oidc provider
+# You will also need to mount a volume for the clients_secrets.json file.
```
verdan, line 25: This is not true.

Dockerfile.search.public
```diff
@@ -8,3 +8,21 @@ COPY requirements-common.txt /app/requirements-common.txt
...
+
+RUN  pip3 install -e .&& \
+     pip3 install -e .[oidc]
+
+ENV FLASK_APP_MODULE_NAME flaskoidc
+ENV FLASK_APP_CLASS_NAME FlaskOIDC
+ENV FLASK_OIDC_WHITELISTED_ENDPOINTS status,healthcheck,health
+ENV SQLALCHEMY_DATABASE_URI sqlite:///sessions.db
+
+# You will need to set these environment variables in order to use the oidc image
```
verdan, line 22: You also need to add `FLASK_OIDC_CONFIG_URL`
Would be nice to add this link for ref: https://github.com/verdan/flaskoidc#readme

This change should go across all repos.

verdan (changes requested):

Dockerfile.search.public
petehanssens: @verdan let me know if this is resolved

Dockerfile.search.public
petehanssens: @verdan i've removed

mgorsk1 (changes requested):
thanks for contribution @petehanssens! I think it would also be useful to extend https://github.com/amundsen-io/amundsen/blob/main/search/Makefile to build both regular and oidc images (just like we do for metadata and frontend, example here: https://github.com/amundsen-io/amundsen/blob/main/frontend/Makefile#L52-L54).

This way oidc image will be build by GH Action and pushed to dockerhub on release.

Also - remember to complete DCO so our checks are happy. Thanks!

petehanssens:
Hi @verdan I think this branch has gotten a bit stale so it might be worthwhile closing

PR #1323, opened 2021-07-07
feat databuilder: delta table owners

jroof88:
THIS IS A WORK IN PROGRESS. The aim here to implement delta lake owners from the metadata into TableOwner metadata objects in the extractor.


jroof88:
Closing as I don't plan to persue this. Owners isn't really a concrete concept in delta. I was able to get around this issue by using a custom transformer that tacks on our owners based on a JSON in S3.

PR #1326, opened 2021-07-08
feat: add complex delta type rendering on frontend

jroof88:
feat: add complex delta type rendering on frontend

This PR adds support to render complex delta types as nested columns on the UI. This is modeled after the implementation already done for `hive` and `presto` types.

### Summary of Changes
The change here is fairly simple in that it just adds a new `DatabaseId` to the `parser.ts` enum in order to render nested fields nicely on the UI. I also linked the docs for Delta Types above the `SUPPORTED_TYPES` const similar to `hive` and `presto`.

### Tests
I added tests to `parser.spec.ts` to specifically test for delta complex types and make sure they are turn into the proper column rendering.

I also added some sample delta tables to `sample_data/sample_table.csv ` and sample columns to `sample_data/sample_col.csv` with some nested fields and ran the application locally to verify it works:
Screen Shot 2021-07-09 at 11 43 59 AM

I decided to commit the sample data as well since then folks who are testing can see what nested fields look like!

### Documentation
I added a link to delta complex types in the code I wrote to follow the convention of `hive` & `presto`

### CheckList

### DIFF
--- a/databuilder/example/sample_data/sample_col.csv
+++ b/databuilder/example/sample_data/sample_col.csv
@@ -10,3 +10,5 @@ col3,"col3 description","string",3,dynamo,gold,test_schema,test_table2,
 col4,"col4 description","int",4,dynamo,gold,test_schema,test_table2,
 col1,"view col description","int",1,hive,gold,test_schema,test_view1,
 col1,"col1 description","int",1,hive,gold,test_schema,test_table3,
+col1,"col1 description","bigint",1,delta,gold,test_schema,delta_test_table
+col2,"col2 nested delta column","struct<col3:string,col4:bigint>",2,delta,gold,test_schema,delta_test_table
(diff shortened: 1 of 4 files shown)


sewardgw (approved):
LGTM, squash commits on merge?

As additional food for thought, since you just created this - do you see the value in having explicitly defined parsing processes per data store? When I looked at this a while ago it wasn't completely obvious why Amundsen parses a struct from Hive but wouldn't parse a struct from BigQuery or Delta, in this example. Curious on your thoughts around the value of this data store validation as well vs treating a map, struct, array, etc. as global standards

jroof88:
> do you see the value in having explicitly defined parsing processes per data store?

I'm kind of indifferent honestly. 

On one side, I feel like this should just happen automatically for all complex/nested data types. It's a little frustrating that all it needed was 2 lines of code but now I need to get this merged and a new frontend image cut for it to show up in my production build. 

However on the other side, let's say we onboard a new data store with some fancy nested type syntax and the parser doesn't know what to do with it then the UI could break. I think that's the only benefit of it being opt-in per data store is that we can verify the parser works properly before allowing that data store to have nicely rendered nested/complex types.

feng-tao:
nested types are different or slightly different for each data store which is the reason we have the parsers per store initially. cc @ttannis who built that.

feng-tao (approved):

feng-tao:
thanks @jroof88 , great that you continue adding support for delta!

jroof88:
> thanks @jroof88 , great that you continue adding support for delta!

No problem! Thanks for the swift reviews. Also more PRs to come. I would like to get nested delta types indexed in the search similar to the BigQuery extractor. Stay tuned 👀

Issue #939, opened 2021-03-02
Installation Datasets : Field May not be null


mclose90:
I was following the installation instructions and everything seemed to be working until I get to step Verify Setup part 2.
When I navigate to http://localhost:5000/table_detail/gold/hive/test_schema/test_table1. I get something went wrong.

In the browser logs and docker logs I see.

{"msg":"Encountered exception: {'columns': {0: {'col_type': ['Field may not be null.'], 'stats': {0: {'stat_type': ['Field may not be null.']}, 1: {'stat_type': ['Field may not be null.']}, 2: {'stat_type': ['Field may not be null.']}, 3: {'stat_type': ['Field may not be null.']}, 4: {'stat_type': ['Field may not be null.']}}}, 1: {'col_type': ['Field may not be null.']}, 2: {'col_type': ['Field may not be null.']}, 3: {'col_type': ['Field may not be null.']}, 4: {'col_type': ['Field may not be null.'], 'stats': {0: {'stat_type': ['Field may not be null.']}, 1: {'stat_type': ['Field may not be null.']}, 2: {'stat_type': ['Field may not be null.']}}}}}","status_code":500,"tableData":{}}

## Expected Behavior

Table information to be displayed 

## Current Behavior

"Something went wrong"
Docker logs show Field may not be null

## Steps to Reproduce

1.) I used a fresh linux mint Ubuntu focal based installation.
2.) Installed all need docker components
3.) follow the installation.md line for line w/ Neo4j
4.) Go all the way to Verify setup step 2

## Screenshots (if appropriate)
![image](https://user-images.githubusercontent.com/41022806/109713497-4b4f3300-7b6f-11eb-857b-1928948c893e.png)

## Context

This issue has affected me by not being able to verify setup. 

## Your Environment

OS: Linux Mint 20.1 x86_64 
Host: VirtualBox 1.2 
Kernel: 5.4.0-66-generic

* Amunsen version used: commit 8836cd90b338f0672753a2efb846d24dd8929a38

aselelliott:
I am facing the same issue. 
OS: Linux 3.10.0 x86_64

Also not being to reach neo4j UI on localhost:7474
![image](https://user-images.githubusercontent.com/80000497/109854289-7ba0db00-7c1c-11eb-8870-678e7fab4b97.png)

Coqueiro:
Facing the same issue with the basic installation

Coqueiro:
[...]

PR #1335, opened 2021-07-12
fix: Makes sure to work well with older versions of databuilder 

verdan:

### Summary of Changes

Related to this issue: https://github.com/amundsen-io/amundsen/issues/939
For Amundsen installation that ingested data using older version of databuilder, they still have a number of columns with the attribute `type` and not the `col_type`. This PR makes the code work with both the older and new version of databuilder and does not breaks the code. 

### CheckList

### DIFF
--- a/metadata/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata/metadata_service/proxy/neo4j_proxy.py
@@ -170,7 +170,7 @@ class Neo4jProxy(BaseProxy):
             last_neo4j_record = tbl_col_neo4j_record
             col = Column(name=tbl_col_neo4j_record['col']['name'],
                          description=self._safe_get(tbl_col_neo4j_record, 'col_dscrpt', 'description'),
-                         col_type=tbl_col_neo4j_record['col']['col_type'],
+                         col_type=tbl_col_neo4j_record['col']['col_type'] or tbl_col_neo4j_record['col']['type'],
                          sort_order=int(tbl_col_neo4j_record['col']['sort_order']),
                          stats=col_stats,
                          badges=column_badges)


metadata/metadata_service/proxy/neo4j_proxy.py
```diff
@@ -170,7 +170,7 @@ def _exec_col_query(self, table_uri: str) -> Tuple:
             last_neo4j_record = tbl_col_neo4j_record
             col = Column(name=tbl_col_neo4j_record['col']['name'],
                          description=self._safe_get(tbl_col_neo4j_record, 'col_dscrpt', 'description'),
-                         col_type=tbl_col_neo4j_record['col']['col_type'],
+                         col_type=tbl_col_neo4j_record['col']['col_type'] or tbl_col_neo4j_record['col']['type'],
```
feng-tao, line 173: I am not sure if that's long term fix. Given we have bump the major version for databuilder, we should update the doc to make it more explicit.

verdan: What is the migration plan to migrate existing data to newer version? People are using older version of databuilder just because of this issue (this has been asked a number of times on Slack and users are eventually downgrading to older version). 

I think this is the efficient fix to support existing data and at the same time migrate to newer version.

feng-tao: It doesn't solve the issue when we have breaking changes.

feng-tao: We should advise the users to upgrade to a newer version. Or what is the downside of doing so?

dorianj: I think the comment is that there isn't a published migration guide to migrate the data from the older version into the newer format?

feng-tao: We add sth similar before (https://github.com/amundsen-io/amundsen/blob/main/docs/faq.md#how-to-migrate-from-amundsen-1x---2x) for upgrade/migration notice.

Issue #1212, opened 2021-06-10
Stop suppressing linting issues in databuilder


mgorsk1:
## Expected Behavior or Use Case

Remove suppressing of flake8 exceptions in databuilder project.

It's a perfect fit for anyone looking for an easy start to contribute to Amundsen. 

## Service or Ingestion ETL

- databuilder

## Possible Implementation

- change line https://github.com/amundsen-io/amundsen/blob/main/databuilder/setup.cfg#L16 to NONE and fix issues in code that came up after running `make lint`

## Context

After unifying and upgrading quality libraries we had to suppress a number of new exceptions.

The full list is available here: https://github.com/amundsen-io/amundsen/blob/main/databuilder/setup.cfg#L16

The subject of this issue/feature request is to remove suppressing those exceptions and solve the issues raised by flake8 directly in the code.

danlfels:
I took a look at this topic, and I don't think there's any way to implement the exact fix you describe. Flake8 has warnings for line breaks before (W503) AND after (W504) binary operators, so there's no good way to keep it happy. There are reports of similar issues (https://github.com/mila-iqia/cookiecutter-pyml/issues/10) (https://github.com/pypa/packaging/issues/144) and the fix I saw there was just to ignore one or the other. 


Currently we ignore W504 warnings (so we allow line breaks only after binary operators) but the current style recommendation is to prefer them before the binary operator, as seen here: (https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).

Personally, I'm in favor of just deleting the ignore line (line 16) from setup.cfg, since it will ignore both W503 and W504, and allow developers to choose their own styles.

mgorsk1:
tThanks @danlfels for taking an interest. At the time this issue was created, there were several different flake8 warnings suppressed by databuilder setup.cfg but it's no longer the case - someone already fixed them. I think this issue can be closed.

PR #1341, opened 2021-07-13
Fix: Delete ignore warnings

danlfels:
Hopefully should solve linting woes.


danlfels:
Should resolve [#1212 ](https://github.com/amundsen-io/amundsen/issues/1212)

PR #1342, opened 2021-07-13
feat: Update feature tags in search on user action

dkunitsk:

### Summary of Changes

This PR changes the method that updates tags with the search service after a user adds/deletes them in the UI. The method is made more generic to work with other entity types (ML Features, specifically). 

I also simplified this slightly by removing the validation in the search service which disallowed the `:` and `/` characters. These characters simply needed to be escaped. This validation was added here: https://github.com/amundsen-io/amundsensearchlibrary/pull/100. After this validation was removed, we can filter more simply by `key`.

I have also changed the `request_search` call from passing `json` to passing `data`. The `data` arg is what all other API calls from frontend use. The `json` arg is a bit of syntactic sugar which saves the caller from doing `json.dumps` and manually adding the `content-type`. However, if a user configures their own search client as [here](https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/api/utils/request_utils.py#L100) it may not support this so the `data` arg is more bulletproof. Only the standard `requests` client supports the `json` arg. 

NOTE: this is technically backwards incompatible, as a deploy of FE without redeploying Search will cause a breakage when updating tags. But this is pretty mild as the only thing that would break is the realtime update in Search.

### Tests

- Updated the relevant search tests and frontend tests. 
- Have also manually tested the new _update_search_tag for PUTs and DELETEs.

### DIFF
--- a/frontend/amundsen_application/api/metadata/v0.py
+++ b/frontend/amundsen_application/api/metadata/v0.py
@@ -12,6 +12,8 @@ from flask import current_app as app
 from flask.blueprints import Blueprint
 
 from amundsen_common.entity.resource_type import ResourceType, to_label
+
+from amundsen_application.api.utils.search_utils import generate_query_json
 from amundsen_application.log.action_log import action_logging
 
 from amundsen_application.models.user import load_user, dump_user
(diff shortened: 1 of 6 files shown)


frontend/amundsen_application/api/metadata/v0.py
```diff
@@ -411,67 +413,66 @@ def _update_metadata_tag(table_key: str, method: str, tag: str) -> int:
...
+        LOGGER.info(f'Fail to get entity from serviceservice, http status code: {search_response.status_code}')
+        LOGGER.info(search_response.text)
+        return search_response.status_code
+
+    raw_data_map = json.loads(search_response.text)
+    # key should uniquely identify this resource
     if len(raw_data_map['results']) > 1:
-        LOGGER.error(f'Error! Duplicate table key: {table_key}')
-    table = raw_data_map['results'][0]
+        LOGGER.error(f'Error! Duplicate table key: {key}')
```
allisonsuarez, line 454: duplicate entity key right?

frontend/amundsen_application/api/metadata/v0.py
```diff
@@ -411,67 +413,66 @@ def _update_metadata_tag(table_key: str, method: str, tag: str) -> int:
...
     new_tags_list = [item for item in old_tags_list if item['tag_name'] != tag]
     if method != 'DELETE':
         new_tags_list.append({'tag_name': tag})
-    table['tags'] = new_tags_list
+    resource['tags'] = new_tags_list
 
     # remove None values
-    pruned_table = {k: v for k, v in table.items() if v is not None}
-
+    pruned_table = {k: v for k, v in resource.items() if v is not None}
```
dkunitsk, line 465: also not "table"

allisonsuarez (approved):
looks great to me, thanks!

PR #1353, opened 2021-07-16
feat: blanket uneditable schema option

jroof88:
### Summary of Changes

This commit adds a new configuration value for the concept of uneditable schemas. The two options right now UNEDITABLE_SCHEMAS and UNEDITABLE_TABLE_DESCRIPTION_MATCH_RULES are not conducive to folks who are deploying amundsen from the docker containers in dockerhub. Here we add a new configuration value ALL_UNEDITABLE_SCHEMAS that pulls it's value from an environment variable andand is a blanket on/off switch to allow schema editing on the UI.

Within my organization, we want don't want users updating anything on the UI. We want to handle all of the metadata during ETL and not allow users to mess with it. I suspect other people using amundsen will want to do the same.

I added some comments in the code about these config values and also updated the documentation for Uneditable Table Descriptions (https://www.amundsen.io/amundsen/frontend/docs/flask_config/#uneditable-table-descriptions) to note this change.

Lastly, I added a value to the helm chart to set this environment variable for those who are deploying to K8s using helm...which I am doing :)

### Tests
I added a new test to the metadata util test suite to verify this config works. I also had added this option to existing MockConfigs in existing tests to verify functionality did not change.

### Documentation
I added some comments in `frontend/config.py` as to what the editeable schema values are and I updated the frontend developer config guide with the new options (https://www.amundsen.io/amundsen/frontend/docs/flask_config/#uneditable-table-descriptions).

### CheckList
Make sure you have checked all steps below to ensure a timely review.

[X] PR title addresses the issue accurately and concisely.
[X] PR includes a summary of changes.
[X] PR adds unit tests, updates existing unit tests, OR documents why no test additions or modifications are needed.
[X] In case of new functionality, my PR adds documentation that describes how to use it.

### DIFF
--- a/frontend/amundsen_application/config.py
+++ b/frontend/amundsen_application/config.py
@@ -32,8 +32,13 @@ class Config:
 
     COLUMN_STAT_ORDER = None  # type: Dict[str, int]
 
+    # The following three variables control whether table descriptions can be edited via the UI
+    # ALL_UNEDITABLE_SCHEMAS: set environment variable to 'true' if you don't want any schemas to be able to be edited
+    # UNEDITABLE_SCHEMAS: a set of schema names whose tables will not be editable
+    # UNEDITABLE_TABLE_DESCRIPTION_MATCH_RULES: a list of regex rules for schema name, table name, or both
+    # See https://www.amundsen.io/amundsen/frontend/docs/flask_config/#uneditable-table-descriptions for more info!
+    ALL_UNEDITABLE_SCHEMAS = os.getenv('ALL_UNEDITABLE_SCHEMAS', 'false') == 'true'  # type: bool
     UNEDITABLE_SCHEMAS = set()  # type: Set[str]
-
     UNEDITABLE_TABLE_DESCRIPTION_MATCH_RULES = []  # type: List[MatchRuleObject]
 
     # DEPRECATED (since version 3.9.0): Please use `POPULAR_RESOURCES_COUNT`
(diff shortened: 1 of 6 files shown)


frontend/amundsen_application/config.py
```diff
@@ -32,8 +32,13 @@ class Config:
 
     COLUMN_STAT_ORDER = None  # type: Dict[str, int]
 
+    # The following three variables control whether table descriptions can be edited via the UI
+    # ALL_UNEDITABLE_SCHEMAS: set environment variable to 'true' if you don't want any schemas to be able to be edited
+    # UNEDITABLE_SCHEMAS: a set of schema names whose tables will not be editable
+    # UNEDITABLE_TABLE_DESCRIPTION_MATCH_RULES: a list of regex rules for schema name, table name, or both
+    # See https://www.amundsen.io/amundsen/frontend/docs/flask_config/#uneditable-table-descriptions for more info!
+    ALL_UNEDITABLE_SCHEMAS = os.getenv('ALL_UNEDITABLE_SCHEMAS', 'false') == 'true'  # type: bool
```
sewardgw, line 40: nit - we may want to do a little bit of processing on the envvar just to allow for True, true, 1 (or 0) etc. that are common values that are used as booleans

sewardgw (approved):

frontend/amundsen_application/config.py
jroof88: Yeah I thought about this but I noticed a similar pattern on [Line 77-78](https://github.com/amundsen-io/amundsen/pull/1353/files/f4c6f039b7bb73f3964af8c76f45e1df2a37eb4e#diff-e6f76b9d9261ca6c32f284a0a78c47c70d36fe887a9dace1de9ae1bb672d8d78R77-R78):
```python
# Settings for Preview Client integration
    PREVIEW_CLIENT_ENABLED = os.getenv('PREVIEW_CLIENT_ENABLED') == 'true'  # type: bool
```

I think a utility function might be nice here but seeing as I documented the exact value that the env var needs to be I think this is ok for now. What do you think?

PR #1355, opened 2021-07-17
feat: Healthier health checks

sewardgw:

Updated health checks to check the health of the services they interact with. This change is 100% backward compatible with the prior health checks.

### Summary of Changes

- Moved default health check response to base proxy classes
- Implemented health checks that validate elastic search and neo4j health
  - Other search / metadata proxies return with the default OK / 200 response
  - Enables additional checks for CPU utilization, memory consumption etc. to be added in the future
- Frontend health checks propagate downward into search and metadata with logic to verify both services are health
- bumped versions

Examples of the frontend health check response:

**Both services fail**
```json
{
  "status": "fail",
  "checks": {
    "search_health": {
      "search_service": {
        "status": "Unable to connect."
      }
    },
    "metadata_health": {
      "metadata_service": {
        "status": "Could not get health from metadata"
      }
    }
  }
}
```

**Both services pass**
```json
{
  "status": "ok",
  "checks": {
    "search_health": {
      "ElasticsearchProxy:connection": {
        "cluster_name": "docker-cluster",
        "status": "yellow",
        "timed_out": false,
        ....
        "number_of_in_flight_fetch": 0,
        "task_max_waiting_in_queue_millis": 0,
        "active_shards_percent_as_number": 51.61290322580645
      }
    },
    "metadata_health": {
      "Neo4jProxy:connection": {
        "overview_enabled": false
      }
    }
  }
}

```

### Tests

Added tests

### Documentation

Updated swagger docs

### CheckList


frontend/Makefile
```diff
@@ -18,7 +18,7 @@ lint:
 
 .PHONY: mypy
 mypy:
-	mypy --ignore-missing-imports --follow-imports=skip --strict-optional --warn-no-return .
+	mypy --ignore-missing-imports --follow-imports=skip --strict-optional --warn-no-return --exclude venv .
```
mgorsk1, line 21: I would move mypy config to setup.cfg instead of complicating the call here:

https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude

wdyt?

frontend/amundsen_application/api/__init__.py
```diff
@@ -1,20 +1,22 @@
...
 
-from typing import Any, Tuple
+from typing import Any
 import logging
 
 from flask import Flask, render_template
 import jinja2
 import os
 
+from .healthcheck import run_healthcheck
```
mgorsk1, line 11: any reason this is not absolute import?

frontend/amundsen_application/api/healthcheck.py
```diff
@@ -0,0 +1,47 @@
...
+def run_healthcheck() -> Tuple[str, int]:
+    search_url = app.config['SEARCHSERVICE_BASE'] + '/healthcheck'
+    metadata_url = app.config['METADATASERVICE_BASE'] + '/healthcheck'
+
+    try:
+        search_health_resp = request_search(url=search_url, method='GET')
+        if search_health_resp.status_code == HTTPStatus.OK:
+            search_health = health_check.HealthCheck(**search_health_resp.json())
+        else:
+            search_checks = {'search_service': {'status': 'Could not get health from search'}}
```
mgorsk1, line 23: This might be confusing message since we catch Connection and ReadTimeout errors and assign very similar `Unable to connect` message

frontend/amundsen_application/api/healthcheck.py
```diff
@@ -0,0 +1,47 @@
...
+            search_health = health_check.HealthCheck(status=health_check.FAIL, checks=search_checks)
+    except (ConnectionError, ReadTimeout):
+        search_checks = {'search_service': {'status': 'Unable to connect.'}}
+        search_health = health_check.HealthCheck(status=health_check.FAIL, checks=search_checks)
+
+    try:
```
mgorsk1, line 29: this block is 99% the same as the one for search. do you think that in the spirit of DRY it could be rewritten as a function?

mgorsk1 (commented):
looks nice, few nits/questions

frontend/Makefile
sewardgw: agreed, will move that over in the next commit

frontend/amundsen_application/api/healthcheck.py
sewardgw: yep, that will work better if/when we add additional checks

frontend/amundsen_application/api/healthcheck.py
sewardgw: As an outside perspective, any thoughts on what would make this more clear?

frontend/amundsen_application/api/__init__.py
sewardgw: not particularly, I'll update this to conform

frontend/Makefile
```diff
@@ -18,7 +18,7 @@ lint:
 
 .PHONY: mypy
 mypy:
-	mypy .
+	mypy --ignore-missing-imports --follow-imports=skip --strict-optional --warn-no-return .
```
feng-tao: any reason for adding this?

metadata/Makefile
```diff
@@ -17,7 +17,7 @@ lint:
 
 .PHONY: mypy
 mypy:
-	mypy .
+	mypy --ignore-missing-imports --follow-imports=skip --strict-optional --warn-no-return .
```
feng-tao: any reason for adding this?

metadata/metadata_service/proxy/neo4j_proxy.py
```diff
@@ -88,13 +90,26 @@ def __init__(self, *,
...
+        checks = {}
+        try:
+            # dbms.cluster.overview() is only available for enterprise neo4j users
+            cluster_overview = self._execute_cypher_query(statement='CALL dbms.cluster.overview()', param_dict={})
+            checks = dict(cluster_overview.single())
+            checks['overview_enabled'] = True
+            status = health_check.OK
+        except neobolt.exceptions.ClientError:
+            checks = {'overview_enabled': False}
+            status = health_check.OK  # Can connect to database but plugin is not available
```
feng-tao, line 108: what plugin is not available?

metadata/metadata_service/proxy/neo4j_proxy.py
```diff
@@ -88,13 +90,26 @@ def __init__(self, *,
...
                                             trust=trust)  # type: Driver
 
-    def is_healthy(self) -> None:
-        # throws if cluster unhealthy or can't connect.  An alternative would be to use one of
-        # the HTTP status endpoints, which might be more specific, but don't implicitly test
-        # our configuration.
-        with self._driver.session() as session:
-            session.read_transaction(self._execute_cypher_query,
-                                     statement='CALL dbms.cluster.overview()', param_dict={})
+    def health(self) -> health_check.HealthCheck:
```
feng-tao, line 93: any reason we rename the method? now other proxies will still have the is_healthy method for healthcheck while neo4j uses a new one?

frontend/Makefile
sewardgw: I actually didn't add this (as seen here https://github.com/amundsen-io/amundsen/blob/main/frontend/Makefile#L21), as part of this PR I had moved the logic under the mypy section of the setup.cfg file but for some reason the `make mypy` command wasn't picking up all of the configs so I reverted it back to what it was before.

metadata/Makefile
sewardgw: same as above

metadata/metadata_service/proxy/neo4j_proxy.py
sewardgw: `dbms.cluster.overview()` a few lines above. In the neo4j docs (https://neo4j.com/docs/operations-manual/current/reference/procedures/) it is shown as an enterprise only feature. Executing this procedure using the OSS version raises this client error

metadata/metadata_service/proxy/neo4j_proxy.py
sewardgw: this function isn't actually being used within the proxies, there is no entrypoint to access the `health` function from what I can tell. This was more so _removed_ rather than renamed. The only reason I didn't change the other proxies is because I didn't implement a proxy-specific health check for them.

dorianj (approved):

PR #1365, opened 2021-07-21
refactor: Updates Postgres extractor to not use information_schema

verdan:

### Summary of Changes

In PostgreSQL information_schema is limited to only tables you have access to. However, the pg_catalog tables are not.
This means that if we want to do a metadata-only integration via a User that has read access to zero tables, at present they'd have to do a materialized view of information_schema. 

This PR makes use of pg_catalog instead properly, to make sure the data is ingested for all users, even with no access. 

### CheckList

### DIFF
--- a/databuilder/databuilder/extractor/base_postgres_metadata_extractor.py
+++ b/databuilder/databuilder/extractor/base_postgres_metadata_extractor.py
@@ -36,7 +36,7 @@ class BasePostgresMetadataExtractor(Extractor):
     DEFAULT_CLUSTER_NAME = 'master'
 
     DEFAULT_CONFIG = ConfigFactory.from_dict(
-        {WHERE_CLAUSE_SUFFIX_KEY: ' ', CLUSTER_KEY: DEFAULT_CLUSTER_NAME, USE_CATALOG_AS_CLUSTER_NAME: True}
+        {WHERE_CLAUSE_SUFFIX_KEY: 'true', CLUSTER_KEY: DEFAULT_CLUSTER_NAME, USE_CATALOG_AS_CLUSTER_NAME: True}
     )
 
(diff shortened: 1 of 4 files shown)


databuilder/databuilder/extractor/base_postgres_metadata_extractor.py
```diff
@@ -36,7 +36,7 @@ class BasePostgresMetadataExtractor(Extractor):
     DEFAULT_CLUSTER_NAME = 'master'
 
     DEFAULT_CONFIG = ConfigFactory.from_dict(
-        {WHERE_CLAUSE_SUFFIX_KEY: ' ', CLUSTER_KEY: DEFAULT_CLUSTER_NAME, USE_CATALOG_AS_CLUSTER_NAME: True}
+        {WHERE_CLAUSE_SUFFIX_KEY: 'true', CLUSTER_KEY: DEFAULT_CLUSTER_NAME, USE_CATALOG_AS_CLUSTER_NAME: True}
```
sewardgw, line 39: What does `WHERE_CLAUSE_SUFFIX_KEY: 'true'` do? Looking at the code it seems like this would get passed into this part of the query:
```sql 
WHERE att.attnum >=0 and {where_clause_suffix}
```

mgorsk1: it's a common placeholder for any additional (you can see also 1=1) conditions in sql query. This way you don't have to add `and` conditionally to where clause

mgorsk1 (approved):

sewardgw (approved):

feng-tao:
@verdan could we delete the local branch once it is merged? otherwise the local branch will show up in every repo that forked by others. thx

PR #1368, opened 2021-07-21
fix: key generation in databuilder query models.

jackolantern:
### Summary of Changes

Currently key generation transforms a query into a key by making it
lowercase, removing all whitespace, and removing all ';' characters.

This fix:
- performs no transformations of the query inside of quoted regions (', ", `)
- collapses runs of whitespace characters to a single ' '
- only removes ';' from the end of the query

This ensures that the queries such as `select 'X; Y; Z'` and
`select 'xyz'` receive different keys.

While this does mean that a query will now most likely now be assigned a
different key than before, this is likely preferable to the current situation.

### Tests

Existing tests have been updated with the new keys and hashes as appropriate, and additional tests have been added which document and verify the new behavior and will help to prevent regressions. 

### Documentation

The docstrings provide additional details about the normalization procedure and its tradeoffs.

### CheckList

### DIFF
--- a/databuilder/databuilder/models/query/base.py
+++ b/databuilder/databuilder/models/query/base.py
@@ -1,0 +1,69 @@
+from typing import Iterator
+
+from databuilder.models.graph_serializable import GraphSerializable
+
+
+class QueryBase(GraphSerializable):
+    @staticmethod
+    def _normalize(sql: str) -> str:
+        """
+        Normalizes a SQL query or SQL expression.
+
+        No checks are made to ensure that the input is valid SQL.
+        This is not a full normalization.  The following operations are preformed:
+
+        - Any run of whitespace characters outside of a quoted region is replaces by a single ' ' character.
+        - Characters outside of quoted regions are made lower case.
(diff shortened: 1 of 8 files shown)


databuilder/databuilder/models/query/base.py
```diff
@@ -0,0 +1,69 @@
...
+        - Characters outside of quoted regions are made lower case.
+        - If present, a trailing ';' is removed from the query.
+
+        Note:
+        Making characters outside quoted regions lower case does not in general result in an equivalent SQL statement.
+        For example, with MySQL the case sensitivity of table names is operating system dependant.
+        In practice, modern systems rarely rely on case sensitivity, and since making the non-quoted regions of the
+        query lowercase is very helpful in identifying queries, we go ahead and do so.
+
+        Also, this method fails to identify expressions such as `1 + 2` and `1+2`.
```
sewardgw, line 25: I dont think this is something Amundsen needs to handle. The standardization below seems to be more than good enough for the overwhelming amount of use-cases

sewardgw (approved):

PR #1389, opened 2021-07-26
fix: return empty list from exceptions in fetch_columns on deltalake extractor

justaparth:
### Summary of Changes
For the deltalake metadata extractor, when fetching columns the code `describe`s the table, and then tries to look at schema by doing `spark.table('<table>').schema`; this is wrapped in a try-catch to fail gracefully.

We ran into an issue recently where running `spark.table('<table>').schema` raised an exception, causing this method to drop into the except and log the error and return `raw_columns`. However, the type of `raw_columns` is not `List[ScrapedColumnMetadata]`, it's a list of spark Row objects, e.g.

```
[Row(col_name='date', data_type='date', comment=''),
 Row(col_name='hour_start', data_type='timestamp', comment=None),
...
```

which caused the eventual user of the result (`create_table_metadata`) to encounter an exception because the metadata was not the type it was expecting.

This commit fixes this by returning an empty list instead of `raw_columns`, and adds a test to assert that this is the behavior if an exception is returned in `fetch_columns`.

### Tests
I added a test to simulate the behavior of `fetch_columns` with a non-existent table, which will cause an exception in the method.

### Documentation
N/A

### CheckList

### DIFF
--- a/databuilder/databuilder/extractor/delta_lake_metadata_extractor.py
+++ b/databuilder/databuilder/extractor/delta_lake_metadata_extractor.py
@@ -299,7 +299,7 @@ class DeltaLakeMetadataExtractor(Extractor):
                 field_dict[field.name] = field
         except AnalysisException as e:
             LOGGER.error(e)
-            return raw_columns
+            return []
         parsed_columns: Dict[str, ScrapedColumnMetadata] = {}
         partition_cols = False
         sort_order = 0
--- a/databuilder/tests/unit/extractor/test_deltalake_extractor.py
+++ b/databuilder/tests/unit/extractor/test_deltalake_extractor.py
@@ -158,6 +158,10 @@ class TestDeltaLakeExtractor(unittest.TestCase):
         for a, b in zip(actual, expected):
             self.assertEqual(a, b)
 
+    def test_fetch_delta_columns_failure(self) -> None:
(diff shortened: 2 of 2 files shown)


databuilder/databuilder/extractor/delta_lake_metadata_extractor.py
```diff
@@ -299,7 +299,7 @@ def fetch_columns(self, schema: str, table: str) -> List[ScrapedColumnMetadata]:
                 field_dict[field.name] = field
         except AnalysisException as e:
             LOGGER.error(e)
-            return raw_columns
+            return None
```
justaparth, line 302: another option would be to return `[]` here; i thought `None` felt a bit more explicit about failure versus an empty list, but i'm open to whatever would be more consistent with the codebase broadly

sewardgw:
Can you fix the signoff?
```
git commit --amend --no-edit --signoff
```

justaparth:
@sewardgw done!

feng-tao:
cc @jroof88 given you work on delta extractor recently

jroof88:
> cc @jroof88 given you work on delta extractor recently

@justaparth and I work together 😎 and he identified this bug! PR LGTM

justaparth:
- update to return `[]` instead of `None`; i think it should fix the CI bug and is actually closer i think to the original code's intent

feng-tao (approved):

PR #1391, opened 2021-07-26
fix: deployment-frontend env nested & quote issue

jroof88:
### Summary of Changes

This commit fixes the deployment-frontend.yaml from a slightly bad change that was introduced in #1353. Namely:

- The logic to render the env var for `ALL_UNEDITABLE_SCHEMAS` was improperly nested in the OIDC config so if you wanted to use it but didn't populate OIDC config then the logic would not work
- Use the quote function to properly make an environment variable.

### Tests

Unforuntely there are no integration tests to make sure the helm chart compiles for different sets of values. I was only able to expose this issue when I tried to deploy #1353 in my production build. It would be nice to have that though? Maybe helm has something? Unsure at the moment.

### Documentation

N/A

### CheckList


feng-tao (approved):
lgtm

PR #1396, opened 2021-07-29
Add github action to create helm chart index

mhumeSF:
Add github action to create helm chart index published on gh-pages
branch. One could then install amundsen helm chart like so:

```
helm repo add amundsen https://amundsen.io/amundsen
helm install amundsen
```

### CheckList


mhumeSF:
Anything else needed for this?

feng-tao:
this looks correct. @mgorsk1 could you take another look in case I miss anything?

feng-tao (approved):

.github/workflows/helm_release.yaml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.2.1
```
feng-tao, line 25: so this publish the chart to github helm repo?

mgorsk1:
Does it mean we need to reserve gh-pages branch solely for helm index?

mgorsk1 (changes requested):
added some questions I think should be answered before merging

mhumeSF:
> does it mean we need to reserve gh-pages branch solely for helm index?

No, just a place for `index.yaml` to be resolvable. For example https://mhumesf.github.io/amundsen/ is still working, as well as https://mhumesf.github.io/amundsen/index.yaml

> maybe we should have separate on release action (like we do for other microservices) instead of running this with every merge to main? Otherwise we have misalignment in release logic

This bit below will make it so only take action when files are modified in the helm directory. Chart-releaser does publish a release title amudsen-<chart-version>. You can see that here https://github.com/mhumeSF/amundsen/releases/tag/amundsen-2.1.1. Maybe it'd be better to publish the release as `admundsen-helm-2.1.1` with other releases. But again, the github workflow will only trigger when files in the path `'amundsen-kube-helm/templates/**'` are updated and will only release when the chart version in `Chart.yaml` is bumped higher.
```
on:
  push:
    branches:
      - main
    paths:
      - 'amundsen-kube-helm/templates/**'
```

> I assume the version of the chart is taken from Chart.yaml? What happens if version is not bumped but new merge to main happens? Does the same version gets overwritten with different chart contents?

In the case where files in the helm chart are modified, but no chart version is bumped, the github action is triggered but will error because it will try to upload using a tag that already exists. 

> I would also document it so users are aware of it cause it's not obvious that we provide helm index.
I'll add to PR updates to documentation.

---

After some investigation, It look likes mkdocs will remove any files published on gh-pages not related to `docs` dir. This means the published `index.yaml` file gets removed when docs are republished. I'll have to look into mkdocs to see if there is a way to skip/overlook index.yaml from getting destroyed.

mhumeSF:
@mgorsk1 Haven't found anything. I'm going to close for now. I'll open a new PR if I find a simple way to do this.

PR #1398, opened 2021-07-30
feat: generalize application model (+ add databricks application support)

jroof88:
### Summary of Changes

Currently the Application model isn't very usable to anyone who doesn't use Airflow. This commit generalize the model while also freezing current Airflow logic into place for backwards compatibility. I update docs and tests as well to make sure it works properly.

I also add a databricks logo for rendering databricks applications nicely on the frontend :)

### Tests

I refactored the tests in `test_application.py` to allow for multiple test cases and added proper testing for the databricks model type

### Documentation

I updated the model documentation to remove the concept Airflow only support

### CheckList


mgorsk1:
do you think there is a way to refactor this class so no coding is required to have different app?

jroof88:
> Do you think there is a way to refactor this class so no coding is required to introduce another app?
> 
> I think only thing that would need to be done is figuring out how to render keys but apart from that it would be much user friendly and scalable.
> 
> Btw shall we introduce new icon for databricks app?

@mgorsk1 Yeah so this was kind of my dilemma while doing this. I did introduce a new logo for Databricks in this PR ([HERE](https://github.com/amundsen-io/amundsen/blob/7cd8f04a79079a2f651d740b1f07575de6066af1/frontend/amundsen_application/static/images/icons/logo-databricks.png)). I don't think it would be very hard to make this open to any kind of app but then the UI rendering would just show up with no icon? Is that ok? Are there any other instances of this happening? I think what I would do is leave all of the Airflow specific things for backwards compatibility and then have a new template for any other type of app. 

I would definitely vote for this route as long as folks are ok with developer essentially being able to put in any kind of app. We could also maybe log a warning message or something in their app doesn't have an icon.

mgorsk1:
We can have a default icon for when the app doesn't match supported ones. The same happens for tables if the database name doesn't have matching icon.

+1 for making generic design for anything except Airflow.

jroof88:
> We can have a default icon for when the app doesn't match supported ones. The same happens for tables if the database name doesn't have matching icon.

@mgorsk1 Cool this is super helpful. I did not know this. I will refactor this PR to be generic but freeze the Airflow constructs already in place for backwards compatibility 👍. Will reach out for a review when it is ready.

feng-tao:
make sense to make the application more generic to support other workflow type including databricks job/prefect etc.

@jroof88 I am curious how you link the databricks job with the table. I did something similar for databricks which requires some internal usage tracking(not visible external).

jroof88:
> make sense to make the application more generic to support other workflow type including databricks job/prefect etc.
> 
> @jroof88 I am curious how you link the databricks job with the table. I did something similar for databricks which requires some internal usage tracking(not visible external).

@feng-tao My plan is to link the amundsen table to the databricks ETL job that writes to the table so it's is easy for folks to go from table to job without have to naviagate/search the databricks jobs ui

feng-tao:
Is it through manual mapping?

jroof88:
> Is it through manual mapping?

We have code that parses out databricks jobs and then can map the links to tables yes

jroof88:
@feng-tao @mgorsk1 This is ready for a review!

jroof88:
Hi @feng-tao & @mgorsk1 friendly ping for a review here! All refactored to be generalized while also adding the dbx logo for folks to use databricks as the application

databuilder/databuilder/models/application.py
```diff
@@ -46,15 +54,54 @@ def __init__(self,
...
+        self.application_type = application_type
+        self.task = task_id
+
+        application_id_format = Application.APPLICATION_ID_FORMAT
+        application_key_format = Application.APPLICATION_KEY_FORMAT
+        application_description_format = Application.APPLICATION_DESCRIPTION_FORMAT
+
+        # The Application model was originally designed to only be compatible with Airflow
+        # If the type is Airflow we must use the hardcoded Airflow constants for backwards compatibility
+        if self.application_type == 'Airflow':
```
mgorsk1, line 75: just to be 200% sure can we do `if self.application_type.lower() == 'airflow'` ?

mgorsk1 (approved):
one nit but lgtm

databuilder/databuilder/models/application.py
jroof88: Yup! Done ✅

PR #1406, opened 2021-08-02
feat: Add frontend configuration to render edit button in TableDetailPage

ozandogrultan:
### Summary of Changes

This PR updates the frontend configuration for tables and adds a new property, `renderEditButton`, to the `supportedDescriptionSources` object. The purpose of this change is to better control rendering of edit buttons in the table detail page and add the ability to remove them completely when tables and column descriptions are `readOnly`. This is particularly useful when metadata and description sources are different.

Here is a screenshot of a table detail page after `ALL_UNEDITABLE_SCHEMAS` env variable is configured as `true` and `renderEditButton` is configured as `false` for `github` source:

Screenshot 2021-08-02 at 16 32 42

### Tests

Added a frontend unit test for the `getDescriptionEditUrl` function.

### Documentation

I've updated the `Supported Description Sources` section of the frontend documentation and added the new configuration property.

### CheckList


jkulzick:
My understanding of the current implementation is that with `ALL_UNEDITABLE_SCHEMAS` set to `true` then the `EditableSection` for every table should have a property `readOnly` set to `true`. Based on [here](https://github.com/amundsen-io/amundsen/blob/44837168f4a3b0380c9eb93295514dfcf641dc44/frontend/amundsen_application/static/js/components/EditableSection/index.tsx#L134) a `renderReadOnlyButton` should be rendered.

Is `renderReadOnlyButton` currently rendering a disabled edit button? If it is not currently disabled perhaps we can just update the code to disable the button. Would that be sufficient for your use case? It seems reasonable to me from a UX perspective to disable buttons that will not do anything.

IMHO we should keep the rendering or disabling of the edit button separate from the `editUrl` because it feels like we are combining concerns (the url we hit for editing vs should someone be able to edit).

ozandogrultan:
@jkulzick The current issue is removal of edit button completely and the value of `editUrl` concerns are already combined since with [this](https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/static/js/components/EditableSection/index.tsx#L88), no edit button is rendered when `editUrl` is null, undefined or empty string, which is the desired behaviour for our use case. But when `editUrl` is populated, which is the default behaviour, a disabled button with a popover is rendered like this:

Screenshot 2021-08-03 at 18 54 03

But for our use case, this creates an unintuitive UI since table metadata source (Github), and description sources are not the same.

Please let me know if you have any other suggestions to overcome this issue, I would be happy to update the PR accordingly.

jkulzick:
@ozandogrultan that makes sense and agree with you on the current UX. it isn't clear to me if the current [check](https://github.com/amundsen-io/amundsen/blob/c78575e19e115f78cdf647ff52465b850c2db415/frontend/amundsen_application/static/js/components/EditableSection/index.tsx#L88) on `editUrl` is a guard or to prevent rending of the edit button. that is perhaps an implementation detail.

If at all possible it would be ideal to do improve the UX without introducing the need for additional configuration since that just makes installing and maintaining the install more cumbersome.

If we can come to some consensus that the current UX for how the edit button is rendered for non-editable tables is counterintuitive then we could just change the current behavior without the need for additional configuration or modifying how the `editUrl` is populated. I could see a number of small tweaks to how the edit button is rendered that could improve the UX but the key here is to arrive at some consensus.

@Golodhros would you or another one of the frontend code owners have some suggestions?

jkulzick:
@ozandogrultan I did not fully understand how the `editUrl` is used, but I think I understand your use case now.

1. `!readOnly` (table `is_editable`) --> an edit button is displayed and on click the user can edit the field in the Amundsen UI
2. `readOnly` (table is not editable) + `editUrl` --> an edit button is displayed on hover. on click the user is is redirected to the `editUrl` (or source of the table metadata or description) outside of Amundsen UI
3. `readOnly` + no `editUrl` --> no edit button is displayed.
4. missing - `readOnly` + `editUrl` (table has a source) + user's cannot edit the source

Now I understand the requested configuration change. I'm definitely not familiar enough with the intention behind the existing `editUrl` feature or the use of `TableSource` to comment so I'm going to sit back and let the experts opine 😅 . Sorry for any confusion!

ozandogrultan:
@jkulzick Thanks a lot for taking time to explain your thought process! The points that you've mentioned are totally correct and since the default behaviour is `2` when the table is `readOnly`, this creates a counter-intuitive UX for our use case.

I've created this PR to solve this UX issue for teams that might be in a similar position as us, but I'm definitely open to any other suggestions by the frontend owners of the project.

MrwanBaghdad:
Hey team it'd be great if we can get some feedback on this PR

frontend/amundsen_application/static/js/config/config-utils.ts
```diff
@@ -344,6 +344,26 @@ export function getDescriptionSourceIconPath(sourceId: string): string {
...
+ * is returned.
+ */
+export function getDescriptionEditUrl(dataSource: TableSource): string | null {
+  const config = AppConfig.resourceConfig[ResourceType.table];
+  const sourceId = dataSource.source_type;
+  if (
+    config &&
+    config.supportedDescriptionSources &&
+    config.supportedDescriptionSources[sourceId] &&
+    config.supportedDescriptionSources[sourceId].renderEditButton
```
Golodhros, line 359: If I understand well, this would default to false on all the description sources added by users that aren't dremio or github, right?

ozandogrultan: The current behavior would be that the users would need to explicitly set `renderEditButton` for their custom description sources as it's not an optional attribute. I can change it to be optional and if we do that this would default to `undefined`. I intentionally made it required so that the users won't have any unintentional behavior, but I'm open to any suggestions on the whole implementation.

Golodhros: I think being backwards compatible outweighs any other consideration at this moment. This is a fairly small change, and we don't want to annoy our users with required changes on our next version just for this change.

Another thing would be if we want to wait to ship this on our next major version, whenever that happens.

WDYT @dorianj ?

ozandogrultan: I am very willing to make the changes for the backwards compatibility, as long as we can reach a consensus.

@dorianj I would be happy to receive your input as well to decide on the approach here. We are trying to solve a relatively small UI issue but I believe a big number of people might potentially have the same issue.

dorianj: Sorry y'all, was on vacation when this happened and missed this.

I don't think this should break back compat... to match current behavior, `undefined` would equate to `true`, which is a bit unintuitive but I think that's OK

ozandogrultan: Thank you @dorianj for checking this. I've made the suggested change so it would be great if you can take a look again. also I've signed-off my last commit but DCO task is still failing for some reason.. if you can let me know how to fix it I can do it

ozandogrultan: I've fixed the checks and implemented the changes, a review would be much appreciated
@dorianj @Golodhros

danwom:
Sorry for jumping in so late in the discussion. If I understand everything correctly:

1. When `is_editable: true`, the edit button allows users to update the descriptions
2. When `is_editable: false`, and `source` is present,  the edit button links the users to the `source`
3. When `is_editable: false`, and `source` is absent,  the edit button is not shown
4. **[new] When `is_editable: false`, and `source` is present, and a config is set, the edit button is not shown**

Could we modify the `source` field in the table metadata response to indicate whether the source is editable or not?

ozandogrultan:
> Sorry for jumping in so late in the discussion. If I understand everything correctly:
> 
> 1. When `is_editable: true`, the edit button allows users to update the descriptions
> 2. When `is_editable: false`, and `source` is present,  the edit button links the users to the `source`
> 3. When `is_editable: false`, and `source` is absent,  the edit button is not shown
> 4. **[new] When `is_editable: false`, and `source` is present, and a config is set, the edit button is not shown**
> 
> Could we modify the `source` field in the table metadata response to indicate whether the source is editable or not?

@danwom The status quo is exactly as you've listed. The changes I'm proposing is not exactly about making the `source` editable or not though. So in our case we would like to make table descriptions uneditable, without an edit button, but table source can be editable or non-editable independently. And the user can access the source with the button at the top-right of the page. Could you please elaborate on why are you suggesting to add a field on `source` in this case? I would be happy to make the necessary changes

PR #1408, opened 2021-08-02
feat: more general notices

elenaboal:
### Summary of Changes

Update notices pattern matching to be more general so that `*.*.*.event_*` would put a notice on all tables that had a table name starting with event_ (for example). Right now, pattern matching on notices are only implemented so that you can match whole clusters, datasources, schemas, or tables, not part of tables.  One use case for this would be adding notices to all event tables in the case of a SEV.

### Tests

Added relevant test cases to existing tests for notices


allisonsuarez (approved):
This looks good to me, thanks Elena!

Golodhros (commented):
Sorry, late to this as coming back from vacations!

I miss an update on the docs for this new feature!

PR #1435, opened 2021-08-11
fix: Removing hardcoded K8s Cluster Domain

nirav-chotai:
### Summary of Changes

`cluster.local` was hardcoded in the ConfigMap which breaks the setup where the Kubernetes Cluster Domain is not the default. And, most of the production clusters running with custom domain names. It's good to have a variable to set the Cluster Domain. 

### Tests

- Deployed helm chart in Kubernetes cluster with different cluster domains
- Deployment completed 
- None of the apps were working as all apps were using services DNS with cluster.local as it was hardcoded in the ConfigMap

### Documentation

- Added `clusterDomain` variable and updated README.md

### CheckList

### DIFF
--- a/amundsen-kube-helm/README.md
+++ b/amundsen-kube-helm/README.md
@@ -29,7 +29,7 @@ This is setup templates for deploying [amundsen](https://github.com/amundsen-io/
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| `clusterDomain`          | Cluster domain                                                                                            | `cluster.local` |
+| clusterDomain | string  | `cluster.local` | Kubernetes Cluster Domain |
 | LONG_RANDOM_STRING | int | `1234` | A long random string. You should probably provide your own. This is needed for OIDC. |
 | affinity | object | `{}` | amundsen application wide configuration of affinity. This applies to search, metadata, frontend and neo4j. Elasticsearch has it's own configuation properties for this. [ref](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) |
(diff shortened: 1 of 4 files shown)


amundsen-kube-helm/README.md
```diff
@@ -29,6 +29,7 @@ This is setup templates for deploying [amundsen](https://github.com/amundsen-io/
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| `clusterDomain`          | Cluster domain                                                                                            | `cluster.local` |
```
mgorsk1, line 32: Can you adjust the docs taking into consideration below remarks? 

1. we shouldn't put key names in `` (see below) as it stands out from rest of them
2. should type be `Cluster domain` or `string`?
3. perhaps `Cluster domain` was supposed to be a description, not type?

image

mgorsk1 (changes requested):

mgorsk1:
Thanks for the contribution @nirav-chotai! Do you think you could also add `rakuten` to Amundsen users here https://github.com/amundsen-io/amundsen/blob/main/README.md#who-uses-amundsen ? (The list is sorted alphabetically btw.)

amundsen-kube-helm/README.md
nirav-chotai: Address above comments.

nirav-chotai:
> Thanks for the contribution @nirav-chotai! Do you think you could also add `rakuten` to Amundsen users here https://github.com/amundsen-io/amundsen/blob/main/README.md#who-uses-amundsen ? (The list is sorted alphabetically btw.)

It's too early, we're evaluating.

mgorsk1 (approved):

PR #1453, opened 2021-08-23
docs: Include example of how to populate a source table using the source model

alldoami:
Signed-off-by: Allison Doami <adoami@chanzuckerberg.com>

### Summary of Changes

Included example of how to populate a source table in Snowflake from an Airflow repo that uses the DAG ID as the file name.


alldoami:
Would love to get this merged so others can benefit!

markgrover:
Thanks!
Looks reasonable to me, but theres' a bunch of whitespace changes. Mind taking them out?

alldoami:
@markgrover is there a problem removing the end line spaces? Feel like it's better to remove them 😅

PR #1459, opened 2021-08-25
feat: Allow dangerous html based on config variable

MrwanBaghdad:
### Summary of Changes

Added configuration variable to allow rendering of dangerous HTML in the programmatic description for table detail pages

### Documentation

The use case is that we'd like to enable embedding in the programmatic descriptions. As the programmatic descriptions are populated by an ingestion pipeline that can validate its safety and is not editable by any web user; a requirement missing in the table description were editing may be allowed and it won't be safe to enable such functionality of rendering dangerous HTML

### CheckList


dorianj:
Do we even need to use `EditableText` for this, given that editable is false? Can we include `ReactMarkdown` directly? I think that would assuage the concern of turning `EditableText` into a footgun?

Golodhros:
> Do we even need to use `EditableText` for this, given that editable is false? Can we include `ReactMarkdown` directly? I think that would assuage the concern of turning `EditableText` into a footgun?

That makes sense to me!

frontend/amundsen_application/static/js/config/config-utils.ts
```diff
@@ -474,6 +474,13 @@ export function isTableQualityCheckEnabled() {
   return AppConfig.tableQualityChecks.isEnabled;
 }
 
+/**
+ * Returns whether dangerous html should be allowed in programmatic descriptions
+ */
+export function isProgrammaticDescAllowDangerousHtml(): boolean {
+  return AppConfig.programmaticDescriptionAllowDangerousHtml;
+}
+
```
Golodhros, line 483: We need a basic tests for this

frontend/amundsen_application/static/js/config/config-utils.ts
```diff
@@ -474,6 +474,13 @@ export function isTableQualityCheckEnabled() {
   return AppConfig.tableQualityChecks.isEnabled;
 }
 
+/**
+ * Returns whether dangerous html should be allowed in programmatic descriptions
+ */
+export function isProgrammaticDescAllowDangerousHtml(): boolean {
```
Golodhros, line 480: this is a mouthful of name. It is tricky, but I think 'hasHTMLDescription' would work better. Same for the config object field.

frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -232,7 +233,15 @@ export class TableDetail extends React.Component<
 
     return descriptions.map((d) => (
       <EditableSection key={`prog_desc:${d.source}`} title={d.source} readOnly>
-        <EditableText maxLength={999999} value={d.text} editable={false} />
+        <div className="editable-text">
```
Golodhros, line 236: I am not sure what are we doing here.
You modified <EditableText> to allow for the 'allowDangerousHtml' prop, why using ReactMarkdown here directly?

Golodhros (changes requested):

ozandogrultan:
@Golodhros I have changed the approach slightly; instead of checking a frontend config variable, now **only programmatic descriptions** always allow dangerous html. This would allow the usage of tags like `iframe` in the programmatic descriptions.

Golodhros:
We talked about this before @dorianj, what do you think about this change?

ozandogrultan:
@dorianj @Golodhros Any updates on this pull request? It is a small change which would help our team (and potentially teams which require similar functionality) a lot and would be great to see it merged.

ozandogrultan:
@dorianj @Golodhros Just bumping this up, our team is still looking forward to seeing this PR getting merged!

Golodhros (approved):
LGTM

PR #1464, opened 2021-08-31
feat: Adding ability to include conditions along with node and relation types in the neo4j staleness removal task

kristenarmes:

### Summary of Changes

Currently, the neo4j staleness removal task can only remove stale nodes and/or relations based on specified types. This new functionality allows users to specify a condition along with the type. To do this, you can specify the inputs to be a list of `TargetWithCondition` objects that each define a target type and a condition. Only stale nodes or relations of that type and that meet the condition will be removed when using this type of input.

For example, you could add a condition that you only want to remove stale nodes of type Table that have some kind of relation to a node of type Column, rather than deleting all stale Table nodes.

This change is backwards compatible so users may continue to only specify the type without conditions for deleting stale nodes or relations.

### Tests

The tests have been updated in test_neo4j_staleness_removal_task.py along with the addition of two new tests: test_validation_statement_with_target_condition and test_delete_statement_with_target_condition. These test the validation and deletion of nodes and relations when target conditions are included.

### Documentation

A description of the feature has been added to databuilder/README.md to describe what the feature does and how to use it.

### CheckList

### DIFF
--- a/databuilder/databuilder/task/neo4j_staleness_removal_task.py
+++ b/databuilder/databuilder/task/neo4j_staleness_removal_task.py
@@ -5,7 +5,7 @@ import logging
 import textwrap
 import time
 from typing import (
-    Any, Dict, Iterable,
+    Any, Dict, Iterable, Union
 )
 
 import neo4j
@@ -52,6 +52,12 @@ DEFAULT_CONFIG = ConfigFactory.from_dict({BATCH_SIZE: 100,
 LOGGER = logging.getLogger(__name__)
 
 MARKER_VAR_NAME = 'marker'
+
+
+class TargetWithCondition:
+    def __init__(self, target_type: str, condition: str) -> None:
+        self.target_type = target_type
+        self.condition = condition
 
 
 class Neo4jStalenessRemovalTask(Task):
@@ -124,10 +130,10 @@ class Neo4jStalenessRemovalTask(Task):
 
     def _delete_stale_nodes(self) -> None:
         statement = textwrap.dedent("""
-        MATCH (n:{{type}})
-        WHERE {}
-        WITH n LIMIT $batch_size
-        DETACH DELETE (n)
+        MATCH (target:{{type}})
+        WHERE {{extra_condition}} {{optional_and}} {}
+        WITH target LIMIT $batch_size
+        DETACH DELETE (target)
         RETURN COUNT(*) as count;
         """)
         self._batch_delete(statement=self._decorate_staleness(statement), targets=self.target_nodes)
@@ -142,26 +148,26 @@ class Neo4jStalenessRemovalTask(Task):
         """
         if self.ms_to_expire:
             return statement.format(textwrap.dedent(f"""
-            n.publisher_last_updated_epoch_ms < (timestamp() - ${MARKER_VAR_NAME})
-            OR NOT EXISTS(n.publisher_last_updated_epoch_ms)"""))
+            target.publisher_last_updated_epoch_ms < (timestamp() - ${MARKER_VAR_NAME})
+            OR NOT EXISTS(target.publisher_last_updated_epoch_ms)"""))
 
         return statement.format(textwrap.dedent(f"""
-        n.published_tag <> ${MARKER_VAR_NAME}
-        OR NOT EXISTS(n.published_tag)"""))
+        target.published_tag <> ${MARKER_VAR_NAME}
+        OR NOT EXISTS(target.published_tag)"""))
 
     def _delete_stale_relations(self) -> None:
         statement = textwrap.dedent("""
-        MATCH ()-[n:{{type}}]-()
-        WHERE {}
-        WITH n LIMIT $batch_size
-        DELETE n
+        MATCH (start_node)-[target:{{type}}]-(end_node)
+        WHERE {{extra_condition}} {{optional_and}} {}
+        WITH target LIMIT $batch_size
+        DELETE target
         RETURN count(*) as count;
         """)
         self._batch_delete(statement=self._decorate_staleness(statement), targets=self.target_relations)
 
     def _batch_delete(self,
                       statement: str,
-                      targets: Iterable[str]
+                      targets: Union[Iterable[str], Iterable[TargetWithCondition]]
                       ) -> None:
         """
         Performing huge amount of deletion could degrade Neo4j performance. Therefore, it's taking batch deletion here.
@@ -170,10 +176,17 @@ class Neo4jStalenessRemovalTask(Task):
         :return:
         """
         for t in targets:
-            LOGGER.info('Deleting stale data of %s with batch size %i', t, self.batch_size)
+            if isinstance(t, TargetWithCondition):
+                type = t.target_type
+                formatted_statement = statement.format(type=t.target_type, extra_condition=t.condition, optional_and='AND')
+            else:
+                type = t
+                formatted_statement = statement.format(type=t, extra_condition='', optional_and='')
+
+            LOGGER.info('Deleting stale data of %s with batch size %i', type, self.batch_size)
             total_count = 0
             while True:
-                results = self._execute_cypher_query(statement=statement.format(type=t),
+                results = self._execute_cypher_query(statement=formatted_statement,
                                                      param_dict={'batch_size': self.batch_size,
                                                                  MARKER_VAR_NAME: self.marker},
                                                      dry_run=self.dry_run)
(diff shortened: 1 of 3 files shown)


databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -124,10 +130,10 @@ def validate(self) -> None:
 
     def _delete_stale_nodes(self) -> None:
         statement = textwrap.dedent("""
-        MATCH (n:{{type}})
-        WHERE {}
-        WITH n LIMIT $batch_size
-        DETACH DELETE (n)
+        MATCH (target:{{type}})
+        WHERE {{extra_condition}} {{optional_and}} {}
```
dkunitsk, line 134: could we give the `{}` that was there a name also. Seems a bit asymmetric to leave it out. Perhaps {{staleness_condition}}. 

Also could we reorder as `{{staleness_condition}} {{extra_condition}}`. 

And also I'm rethinking the `{{optional_and}}` part. Probably the more straightforward thing is just, when adding the extra_condition, do it like:

`statement.format(extra_condition='AND ' + extra_condition)`

databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -182,19 +195,16 @@ def _batch_delete(self,
                 total_count = total_count + count
                 if count == 0:
                     break
-            LOGGER.info('Deleted %i stale data of %s', total_count, t)
+            LOGGER.info('Deleted %i stale data of %s', total_count, type)
 
     def _validate_staleness_pct(self,
                                 total_records: Iterable[Dict[str, Any]],
```
dkunitsk, line 201: can we just pass numbers here instead of this Iterable[Dict[str, Any]]. That will make this function simpler and I don't see why passing Iterable[Dict[str, Any]] was ever necessary.

databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -210,47 +220,63 @@ def _validate_staleness_pct(self,
...
 
         stale_nodes_statement = textwrap.dedent("""
-        MATCH (n)
-        WHERE {}
-        WITH DISTINCT labels(n) as node, count(*) as count
+        MATCH (target:{{type}})
+        WHERE {{extra_condition}} {{optional_and}} {}
+        WITH DISTINCT labels(target) as node, count(*) as count
         RETURN head(node) as type, count
         """)
```
dkunitsk, line 234: this can be consolidated into one

databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -210,47 +220,63 @@ def _validate_staleness_pct(self,
...
-        self._validate_staleness_pct(total_records=total_records,
-                                     stale_records=stale_records,
-                                     types=self.target_nodes)
+        for t in self.target_nodes:
+            if isinstance(t, TargetWithCondition):
+                formatted_total_nodes_statement = total_nodes_statement.format(type=t.target_type, optional_where='WHERE ', extra_condition=t.condition)
+                formatted_stale_nodes_statement = stale_nodes_statement.format(type=t.target_type, extra_condition=t.condition, optional_and='AND')
+            else:
+                formatted_total_nodes_statement = total_nodes_statement.format(type=t, optional_where='', extra_condition='')
+                formatted_stale_nodes_statement = stale_nodes_statement.format(type=t, extra_condition='', optional_and='')
```
dkunitsk, line 244: also this i think can be simplified. Inside the `isinstance(t, TargetWithCondition)` just set the interesting variables and do the formatting once below.

databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -210,47 +220,63 @@ def _validate_staleness_pct(self,
...
+
+            total_records = self._execute_cypher_query(statement=formatted_total_nodes_statement)
+            stale_records = self._execute_cypher_query(statement=formatted_stale_nodes_statement,
+                                                       param_dict={MARKER_VAR_NAME: self.marker})
+            self._validate_staleness_pct(total_records=total_records,
+                                         stale_records=stale_records)
 
     def _validate_relation_staleness_pct(self) -> None:
```
dkunitsk, line 252: same story as above

databuilder/tests/unit/task/test_neo4j_staleness_removal_task.py
```diff
@@ -222,26 +271,26 @@ def test_delete_statement_publish_tag(self) -> None:
             mock_execute.assert_any_call(dry_run=False,
                                          param_dict={'marker': u'foo', 'batch_size': 100},
                                          statement=textwrap.dedent("""
-            MATCH (n:Foo)
+            MATCH (target:Foo)
```
dkunitsk, line 274: I'm seeing a lot of repetition of the same template (mostly by previous authors). Could you pull out all/most of the query templates into class variables in the code and then just reference them both in the code and in the tests so you don't have to type out this boilerplate each time.

dkunitsk (commented):
Awesome. Pretty much all my comments are nits.

databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -64,6 +70,32 @@ class Neo4jStalenessRemovalTask(Task):
 
     """
 
+    delete_stale_nodes_statement = textwrap.dedent("""
+        MATCH (target:{{type}})
+        {staleness_condition}{{extra_condition}}
```
dkunitsk, line 75: an we make staleness_condition one of `{{` as well so it has to be named when formatting. Just makes things more uniform IMO. Same points about the other statements.

databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -64,6 +70,32 @@ class Neo4jStalenessRemovalTask(Task):
...
+        DELETE target
+        RETURN count(*) as count;
+        """)
+    validate_node_staleness_statement = textwrap.dedent("""
+        MATCH (target:{{type}})
+        {staleness_condition}{{extra_condition}}
+        WITH DISTINCT labels(target) as node, count(*) as count
+        RETURN head(node) as type, count
+        """)
+    validate_relation_staleness_statement = textwrap.dedent("""
```
dkunitsk, line 93: why are we still returning type here

databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -182,75 +210,84 @@ def _batch_delete(self,
...
+
+        for t in self.target_nodes:
+            if isinstance(t, TargetWithCondition):
+                target_type = t.target_type
+                total_extra_condition = 'WHERE ' + t.condition
+                stale_extra_condition = ' AND ' + t.condition
+            else:
+                target_type = t
+                total_extra_condition = ''
+                stale_extra_condition = ''
```
dkunitsk, line 244: i feel like this could be simpler if you always include the WHERE part (in the statement itself) an for the total statement you format the staleness condition as just "true". WDYT

databuilder/databuilder/task/neo4j_staleness_removal_task.py
```diff
@@ -182,75 +210,84 @@ def _batch_delete(self,
...
+
+            total_records = self._execute_cypher_query(
+                statement=total_relations_statement.format(type=target_type,
+                                                           extra_condition=total_extra_condition))
+            stale_records = self._execute_cypher_query(
+                statement=stale_relations_statement.format(type=target_type,
+                                                           extra_condition=stale_extra_condition),
+                param_dict={MARKER_VAR_NAME: self.marker})
+
+            total_record_value = next(iter(total_records), None)
```
dkunitsk, line 285: if you just return the count from the statement instead of the tuple, i believe you won't have to do this weird thing.

databuilder/databuilder/task/neo4j_staleness_removal_task.py
kristenarmes: Agreed about the uniformity but the reason I put the single brace around `staleness_condition` (and the reason it was single braces before it was named) is bc each time `.format` is called on the statement it removes one set of curly braces, so since we have the `_decorate_staleness` function that handles populating the staleness condition first, and then after that the target `type` and `extra_condition` are left with a single `{}` and are populated in each of the other functions after. Do you think this is alright, or an alternative would be to bring all the formatting into the `_decorate_staleness` function if that is preferable?

databuilder/databuilder/task/neo4j_staleness_removal_task.py
kristenarmes: Users have the ability to override the allowed staleness percentage for types in the config which is why I left it originally, but you're right that I'm not using it from the query since we have the types now so I'll take it out

databuilder/databuilder/task/neo4j_staleness_removal_task.py
kristenarmes: The `_execute_cypher_query` function returns a `Iterable[Dict[str, Any]]` type which is the reason for this, the same next iter thing is being done in `_batch_delete` when only a count is returned (although I didn't realize it when I wrote this part)

dkunitsk (approved):

Issue #1481, opened 2021-09-06
Bug Report: ES 7.x metadata (properties key) parsing fail 


dkdbsd:
## Current Behavior
I'm trying to extract ES7.x metadata, but it cat not extract properties.
If it fails, it shows `something went wrong...` message at amundsen browser when i click resource link.

## Possible Solution
The location of `properties` key at metadata response is varies by ES version.
in 6.x, mappings > table > properties
in 7.x, mappings > properties

so,  `properties = list(mappings.values())[0].get('properties', dict())` is not work at 7.x.

## Steps to Reproduce

1.  first step: add codes for get ES version at init()
```
  self.es_version = self.es.info().get('version', dict()).get('number', dict())
  self.es_major_version = int(self.es_version[:self.es_version.find('.')])
```
2. second step: fix `first step = add codes for get ES version at init()` as below
```
  if self.es_major_version >= 7:
      properties = mappings.get('properties', dict())
  else:
      properties = list(mappings.values())[0].get('properties', dict())
```

## Your Environment
* Amunsen version used: 
* Data warehouse stores:  es 7.6, es6.7
* Deployment (k8s or native): native
* Link to your fork or repository:  https://github.com/dkdbsd/amundsen/blob/main/databuilder/databuilder/extractor/es_base_extractor.py

mgorsk1:
yeah it does make sense, feel free to submit PR with this fix :)

PR #1482, opened 2021-09-06
fix: es_base_extractor.py to support ES 7.x

dkdbsd:
### Summary of Changes

The location of properties key at metadata response is varies by ES version.
in 6.x, mappings > table > properties
in 7.x, mappings > properties
so, properties = list(mappings.values())[0].get('properties', dict()) is not worked at ES 7.x.

### Tests
extract metadata at es7.6 and es6.7,
and check the results at neo4j and amundsen-es.

### Documentation

https://github.com/amundsen-io/amundsen/issues/1481

### DIFF
--- a/databuilder/databuilder/extractor/es_base_extractor.py
+++ b/databuilder/databuilder/extractor/es_base_extractor.py
@@ -29,6 +29,8 @@ class ElasticsearchBaseExtractor(Extractor):
         self._extract_iter = self._get_extract_iter()
 
         self.es = self.conf.get(ElasticsearchBaseExtractor.ELASTICSEARCH_CLIENT_CONFIG_KEY)
+        self.es_version = self.es.info().get('version', dict()).get('number', dict())
+        self.es_major_version = int(self.es_version[:self.es_version.find('.')])
 
     def _get_indexes(self) -> Dict:
         result = dict()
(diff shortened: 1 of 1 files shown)


databuilder/databuilder/extractor/es_base_extractor.py
```diff
@@ -48,7 +50,10 @@ def _get_index_mapping_properties(self, index: Dict) -> Optional[Dict]:
         mappings = index.get('mappings', dict())
 
         try:
-            properties = list(mappings.values())[0].get('properties', dict())
+            if self.es_major_version >= 7:
```
mgorsk1, line 53: would you mind extending tests to verify both cases?

dkdbsd: let me know how can i extend the test, please.  anyway, I already tested and using at v7.6 and v6.7.

mgorsk1: I would create method similar to this one (in the same class): https://github.com/amundsen-io/amundsen/blob/main/databuilder/tests/unit/extractor/test_es_metadata_extractor.py#L75 but mock the es client so it behaves like it's version 7 (and return payload expected by version 7) and make similar assertions

mgorsk1:
would you mind signing DCO @dkdbsd ? without it we cannot merge this change

feng-tao:
ping @dkdbsd

mgorsk1:
fixed in https://github.com/amundsen-io/amundsen/pull/1648

PR #1503, opened 2021-09-24
feat: added TableStats model

allisonsuarez:

### Summary of Changes

Added TableStats model which allows adding Stats directly to table nodes rather than only to table columns.

### CheckList

### DIFF
--- a/databuilder/databuilder/models/table_stats.py
+++ b/databuilder/databuilder/models/table_stats.py
@@ -3,7 +3,6 @@
 from typing import (
     Iterator, Optional, Union,
 )
-import logging
 from amundsen_rds.models import RDSModel
 from amundsen_rds.models.column import ColumnStat as RDSColumnStat
 
@@ -16,15 +15,15 @@ from databuilder.models.table_serializable import TableSerializable
 LABEL = 'Stat'
 STAT_RESOURCE_RELATION_TYPE = 'STAT_OF'
 RESOURCE_STAT_RELATION_TYPE = 'STAT'
-LOGGER = logging.getLogger(__name__)
+
+
 class TableStats(GraphSerializable, TableSerializable):
     """
     Table stats model.
     """
-    
+
     KEY_FORMAT = '{db}://{cluster}.{schema}' \
                  '/{table}/{stat_type}/'
-    
 
     def __init__(self,
                  table_name: str,
--- a/databuilder/databuilder/models/table_stats.py
+++ b/databuilder/databuilder/models/table_stats.py
@@ -3,27 +3,135 @@
 from typing import (
     Iterator, Optional, Union,
 )
-
+import logging
 from amundsen_rds.models import RDSModel
 from amundsen_rds.models.column import ColumnStat as RDSColumnStat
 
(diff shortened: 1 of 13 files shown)


databuilder/databuilder/models/table_stats.py
```diff
@@ -94,7 +202,7 @@ def _create_node_iterator(self) -> Iterator[GraphNode]:
         """
         node = GraphNode(
             key=self.get_table_stat_model_key(),
```
youngyjd, line 204: Let's rename this function to get_column_stat_model_key?

databuilder/databuilder/models/table_stats.py
```diff
@@ -10,20 +10,128 @@
...
-from databuilder.models.table_metadata import ColumnMetadata
+from databuilder.models.table_metadata import ColumnMetadata, TableMetadata
 from databuilder.models.table_serializable import TableSerializable
 
+LABEL = 'Stat'
+STAT_RESOURCE_RELATION_TYPE = 'STAT_OF'
+RESOURCE_STAT_RELATION_TYPE = 'STAT'
+
+
+class TableStats(GraphSerializable, TableSerializable):
```
youngyjd, line 21: What are the stats that we want to surface? There are some metadata on the left side panel in table details page, for example, total number of rows, etc. Will some of the metadata be moved to this stats model?

allisonsuarez: As of right now we are indexing usage stats like dau, query count, and service specific usage like for airflow, flyte, etc. In the future however we could use this to index stats about the data in the table as well, similar to how we already do for columns.

databuilder/databuilder/models/table_stats.py
```diff
@@ -10,20 +10,128 @@
...
+        if schema is None:
+            self.schema, self.table = table_name.split('.')
+        else:
+            self.table = table_name
+            self.schema = schema
+        self.db = db
+        self.start_epoch = start_epoch
+        self.end_epoch = end_epoch
+        self.cluster = cluster
+        self.stat_type = stat_name
```
dkunitsk, line 49: consistent naming plz

databuilder/databuilder/models/table_stats.py
```diff
@@ -10,20 +10,128 @@
...
+    """
+
+    KEY_FORMAT = '{db}://{cluster}.{schema}' \
+                 '/{table}/{stat_type}/'
+
+    def __init__(self,
+                 table_name: str,
+                 stat_name: str,
+                 stat_val: str,
+                 is_metric: bool,
```
dkunitsk, line 33: could you leave a comment about what this means. "metrics" and "stats" are synonyms generally so we should define what distinction we're making

databuilder/databuilder/models/table_stats.py
dkunitsk: probably by renaming the instance variable, not the argument name. Since the existing model uses that arg name too and that's more likely to cause issues.

dkunitsk (approved):
lgtm

PR #1504, opened 2021-09-27
fix: Apache Superset Chart extractor

vazmeee:
### Summary of Changes

Fixed Apache Superser chart extractor _get_dashboard_details function


mgorsk1 (approved):
thanks!

PR #1509, opened 2021-09-30
feat: extend search result click logging

allisonsuarez:
Added field `position` to click log object which captures the position of the clicked search result on the search results page.


Golodhros (approved):

frontend/amundsen_application/static/js/ducks/log/api/v0.ts
```diff
@@ -14,6 +14,7 @@ export interface ClickLogParams {
   target_type?: string;
   label?: string;
   value?: string;
+  position?: string;
```
danwom, line 17: Canonically we've been using `index` instead of position. Might consider changing to be consistent.

PR #1552, opened 2021-10-28
fix: browse badges paddding

mgorsk1:
### Summary of Changes

Adds padding for badges when the list spans more than one line.

### Tests

n/a

### Documentation

n/a

### CheckList

### DIFF
--- a/frontend/amundsen_application/static/js/components/Badges/BadgeBrowseList/styles.scss
+++ b/frontend/amundsen_application/static/js/components/Badges/BadgeBrowseList/styles.scss
@@ -10,8 +10,9 @@
   .badge-list {
     display: block;
     font-weight: $font-weight-body-regular;
-    margin-top: $spacer-3;
+    margin-top: $spacer-2;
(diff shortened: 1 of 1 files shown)


frontend/amundsen_application/static/js/components/Badges/BadgeBrowseList/styles.scss
```diff
@@ -10,8 +10,9 @@
   .badge-list {
     display: block;
     font-weight: $font-weight-body-regular;
-    margin-top: $spacer-3;
+    margin-top: $spacer-2;
     margin-bottom: $spacer-2;
+    line-height: 2em;
```
Golodhros, line 15: I would prefer not to use ems here, as we haven't really used them in other places.
Maybe there is a value related to the spacer that works for this case?

mgorsk1:
sure, I've added the change to use value from variables @Golodhros 
image

Golodhros (approved):

PR #1556, opened 2021-10-28
feat: search v2 API

allisonsuarez:

### Summary of Changes

- added search_v2 endpoint that is general to all resources
- wrote new es search proxy
- made requests to different indexes from search rather than leaving that up to the frontend
- used multi search to execute search requests to ES
- rewrote ES queries using elasticsearch-dsl lib
- added request, response, and filter schema to common

### Tests

- test_es_search_proxy.py


search/search_service/proxy/es_search_proxy.py
```diff
@@ -0,0 +1,316 @@
...
+        Builds the query object for all of the filters given in the search request
+        """
+        mapping = self.RESOUCE_TO_MAPPING.get(resource)
+
+        filter_queries: List = []
+
+        for filter in filters:
+            filter_name = mapping.get(filter.name) if mapping != None and mapping.get(filter.name) != None else filter.name
+
+            queries_per_term = [Q(WILDCARD_QUERY,  **{filter_name: term}) for term in filter.values]
```
dkunitsk, line 177: why not https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/query.py#L494

search/search_service/proxy/es_search_proxy.py
```diff
@@ -0,0 +1,316 @@
...
+
+RESOURCE_STR_MAPPING = {
+    'table': Resource.TABLE,
+    'dashboard': Resource.DASHBOARD,
+    'feature': Resource.FEATURE,
+    'user': Resource.USER,
+}
+
+
+class Filter():
```
dkunitsk, line 38: marshmallow

search/search_service/proxy/es_search_proxy.py
```diff
@@ -0,0 +1,316 @@
...
+            else:
+                return {
+                    "msg": "Failure",
+                    "page_index": page_index,
+                    "results_per_page": results_per_page,
+                    "results": [],
+                    "status_code": 400,
+                    #  TODO surface actual error
+                    # this makes it so if one resource request fails the whole thing errors
+                    # is this the desired behavior?
```
dkunitsk, line 256: seems correct. And yes logging would be good.

search/search_service/proxy/es_search_proxy.py
```diff
@@ -0,0 +1,316 @@
...
+                            # remove "raw" from mapping value
+                            field = fields.get(f).split('.')[0]
+                            result[f] = search_result._source[field]
+                        result["search_score"] = search_result._score
+                        results.append(result)
+                    # replace empty results with actual results
+                    results_per_resource[resource_type] = {
+                        "results": results,
+                        "total_results": results_count
+                    }
```
dkunitsk, line 246: let's think about how to make this nicer

search/search_service/api/search.py
```diff
@@ -0,0 +1,72 @@
...
+        try:
+            search_results = self.search_proxy.search(query_term=query_term,
+                                                      page_index=page_index,
+                                                      results_per_page=results_per_page,
+                                                      resource_types=resources,
+                                                      filters=filter_objs) 
+            return self.results_schema().dump(search_results), HTTPStatus.OK
+
+        except RuntimeError as e:
+            raise e
```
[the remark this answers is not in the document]
allisonsuarez: Yeah I still need to figureout all of the error states, wanted to m ake sure functionality is good first

search/search_service/proxy/es_search_proxy.py
allisonsuarez: I was wondering if it was worth only returning results for the resources that didn't fail but that has the potential for us to have a broken query for a resource and not notice until we're like huh we havent gotten any results for this one particular one idk

search/search_service/proxy/es_search_proxy.py
allisonsuarez: I think this will get way nicer when we have more standard mappings tbh, also having some sort of marshmallow schema for these results

search/search_service/proxy/es_search_proxy.py
allisonsuarez: I think I'll stick to my current approach with queries, because essentially I'd have to add a | or & between all queries and that seems a bit clunky as opposed to just giving the bool query the list of wildcard queries and letting it handle it

search/search_service/__init__.py
```diff
@@ -89,6 +90,9 @@ def create_app(*, config_module_class: str) -> Flask:
     # Health Check
     api.add_resource(HealthcheckAPI, '/healthcheck')
 
+    # New search endpoint
+    api.add_resource(SearchAPI, '/search_resources')
```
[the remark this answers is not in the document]
mgorsk1: just a nit but versioning endpoints should be closer to root path like /api/v2/search not in resource name

search/search_service/proxy/es_search_proxy.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+    def _build_term_query(self, resource: Resource, query_term: str) -> Q:
+        """
+        Builds the query object for the inputed search term
+        """
+        if query_term == "" or not query_term:
+            # We don't want to create multi_match query for ""
+            # because it will result in no matches even with filters
+            return None
```
mgorsk1, line 114: if we allow returning None then return type of this function should be Optional[None]

search/search_service/proxy/es_search_proxy.py
```diff
@@ -0,0 +1,313 @@
...
+
+        fields = []
+
+        # TODO make resources into an enum
+        if resource == Resource.TABLE:
+            fields = ["display_name^1000",
```
mgorsk1, line 120: can we have properties to store those field lists with weights? It'd be easier to fine tune them if required by certain users

dkunitsk: Yes, but a bit later, when we're cleaning up mappings. We know this is ugly right now.

search/search_service/proxy/es_search_proxy.py
mgorsk1: If we need some default query then i would replace wildcard * with match_all query

search/search_service/proxy/es_search_proxy.py
```diff
@@ -0,0 +1,316 @@
+# Copyright Contributors to the Amundsen project.
+# SPDX-License-Identifier: Apache-2.0
+
```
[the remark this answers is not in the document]
allisonsuarez: tere are going to be different ES proxies for different APIs so I want to keep the proxy directory, its nicer than having search, facets, and documents in 1 giant proxy class

search/search_service/__init__.py
```diff
@@ -89,6 +90,9 @@ def create_app(*, config_module_class: str) -> Flask:
     # Health Check
     api.add_resource(HealthcheckAPI, '/healthcheck')
 
+    # New search endpoint
+    api.add_resource(SearchAPI, '/search_v2')
```
mgorsk1, line 94: is there a reason why we are using such naming convention? APIs are usually versioned like that /api/v1/search and api/v2/search so I wouldn't want us to establish different precedence. the resource is still `search`, if the version changes it should go under different root v tag

search/search_service/api/search.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import (  # noqa: F401
+    Any, Iterable, List,
+)
+
+from amundsen_common.models.search import SearchRequestSchema
+from flasgger import swag_from
+from flask_restful import Resource, request
+
+from search_service.proxy.es_search_proxy import (
+    RESOURCE_STR_MAPPING, ElasticsearchProxy, Resource as AMDResource,
```
mgorsk1, line 15: AMD is not an obvious abbreviation for me, we could go just with `AmundsenResource`

search/search_service/proxy/es_search_proxy.py
mgorsk1: but the MR is supposed to improve on what we have :P just saying it doesn't seem like major refactor and it'd be in line with what we have with _MAPPING constants

search/search_service/__init__.py
allisonsuarez: oh I like that actually I'll try that out since I am rewriting documents API and adding a facets API

search/search_service/proxy/es_search_proxy.py
allisonsuarez: The thing is that having cleaner more general mappings will inevitably change how we handle this functionality altogether so we didn't want to spend too much time cleaning this up since it is going to change almost completely anyway soon, definitely agree that this is a part of the code that requires major improvement

dkunitsk (approved):

sagar-raythatha:
@allisonsuarez Do you have any idea that how to make advance search case insensitive? In earlier version I have made this using https://github.com/amundsen-io/amundsen/issues/1566 however, applying the same in this version did not make this work. Can you please help on this?

PR #1568, opened 2021-11-05
feat: Elasticsearch: 7.x support, extract nested columns, and correct sort order

jaredbwasserman:
### Summary of Changes
1. Elasticsearch base extractor retrieves mapping properties for Elasticsearch >= 7.x. This update is needed because the mapping format was updated following the removal of types (https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html).
2. Nested columns optionally extracted in Elasticsearch metadata extractor.
3. Column sort order optionally corrected in Elasticsearch metadata extractor.

### Tests

1. Modified `TestElasticsearchIndexExtractor.test_extractor_without_technical_data_es_v6` (previously called `TestElasticsearchIndexExtractor.test_extractor_without_technical_data`) to explicitly test Elasticsearch version 6 (and lower).
2. Added `TestElasticsearchIndexExtractor.test_extractor_without_technical_data_es_v7` which is the same test as `TestElasticsearchIndexExtractor.test_extractor_without_technical_data_es_v6` but for Elasticsearch version 7.
3. Added `TestElasticsearchIndexExtractor.test_extractor_nested_es_v6` to test nested column extraction and sort order for Elasticsearch version 6 (and lower).
4. Added `TestElasticsearchIndexExtractor.test_extractor_nested_es_v7` to test nested column extraction and sort order for Elasticsearch version 7.

### Documentation

Added information in the [README](https://github.com/amundsen-io/amundsen/blob/main/databuilder/README.md#elasticsearchmetadataextractor) for `ElasticsearchMetadataExtractor` discussing the newly added `extract_nested_columns` and `correct_sort_order` conf options.

### CheckList


feng-tao:
cc @mgorsk1

feng-tao:
also cc @dkunitsk @allisonsuarez as the search is being revamped

mgorsk1:
Thanks for the PR @jaredbwasserman! I think it makes sense to introduce this change and I'd even go as far as making it a defaullt (and only) behavior. 

What's worth noticing is that it's not actually nested fields you propose to extract, just subsequent fields following some naming convention. From your example `field1.field2` - `field2` is not really a nested column within `field1` family, it's just a naming convention. Internally `field1.field2` in elasticsearch is a field that would be completely independent from `field1.field3` for example. Nested type is something completely different in ES: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested

databuilder/databuilder/extractor/es_base_extractor.py
```diff
@@ -2,22 +2,37 @@
...
 
 from databuilder.extractor.base_extractor import Extractor
+from databuilder.models.table_metadata import ColumnMetadata
 
 
 class ElasticsearchBaseExtractor(Extractor):
     """
     Extractor to extract index metadata from Elasticsearch
+
+    By default, the extractor does not extract nested columns. Set ELASTICSEARCH_EXTRACT_NESTED_COLUMNS conf to True
```
mgorsk1, line 18: Can we make it default and only behavior? 

Btw. nested field is sth else so I wouldn't confuse the terminology (see my comment in the PR thread) https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested

databuilder/databuilder/extractor/es_base_extractor.py
```diff
@@ -30,6 +45,18 @@ def init(self, conf: ConfigTree) -> None:
...
+        version = ''
+
+        try:
+            info = self.es.info()
+            if info:
+                version = info.get('version').get('number')
+        except Exception:
+            pass
+
+        return version
```
mgorsk1, line 58: This function currently allows scenario where value = '' is returned. Maybe we should raise an error if that's the case instead of allowing it to happen?

databuilder/databuilder/extractor/es_base_extractor.py
```diff
@@ -47,13 +74,48 @@ def _get_indexes(self) -> Dict:
...
+        no_mapping_types = False
+        version = self._get_es_version()
+        if len(version) > 0:
+            version_numbers = version.split('.')
+            if int(version_numbers[0]) >= 7:
+                no_mapping_types = True
+
         try:
-            properties = list(mappings.values())[0].get('properties', dict())
+            if no_mapping_types:
```
mgorsk1, line 87: I would simplify this and just do `if version.split('.')[0]>=7`.

databuilder/databuilder/extractor/es_base_extractor.py
```diff
@@ -47,13 +74,48 @@ def _get_indexes(self) -> Dict:
...
+                properties = mappings.get('properties', dict())
+            else:
+                properties = list(mappings.values())[0].get('properties', dict())
         except Exception:
             properties = dict()
 
         return properties
 
+    def _get_nested_columns(self,
```
mgorsk1, line 96: please rename to _get_attributes, because:
1. there are no columns in elasticsearch, only attributes/fields of a document
2. nested type is sth else (see my previous comments) https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested

mgorsk1 (changes requested):
Added some comments.

databuilder/databuilder/extractor/es_base_extractor.py
jaredbwasserman: Thanks for the clarification on nested fields! I agree with making it default and only behavior. I updated the code and README.

databuilder/databuilder/extractor/es_base_extractor.py
jaredbwasserman: I agree that it is not good to allow `value = ''`. The scenario where we are unable to retrieve Elasticsearch version would imply a deep enough issue that raising an error would be the best course of action. I updated the code to remove the try/catch because the underlying `info()` call will raise its own error if it encounters a problem.

databuilder/databuilder/extractor/es_base_extractor.py
jaredbwasserman: Agreed. Updated.

databuilder/databuilder/extractor/es_base_extractor.py
jaredbwasserman: Agreed. Updated.

jaredbwasserman:
@mgorsk1 Can you take another look at the code? I addressed your feedback with a few updates. Thanks!

PR #1569, opened 2021-11-06
feat: allow multiple applications

dkunitsk:
We'd like to support associating multiple applications per table. The databuilder model already permits this (Tables to Applications relationship is many-to-many), but the rest of Amundsen does not. 

Here's the design we're working on (i.e. if there's multiple Applications, they'd get a dropdown, grouped per application name).  
Screen Shot 2021-11-05 at 4 54 16 PM

We will make this work for generic applications, following https://github.com/amundsen-io/amundsen/pull/1398.

### DIFF
--- a/common/amundsen_common/models/table.py
+++ b/common/amundsen_common/models/table.py
@@ -183,6 +183,7 @@ class Table:
     owners: List[User] = []
     watermarks: List[Watermark] = []
     table_writer: Optional[Application] = None
+    table_apps: Optional[List[Application]] = None
     resource_reports: Optional[List[ResourceReport]] = None
     last_updated_timestamp: Optional[int] = None
     source: Optional[Source] = None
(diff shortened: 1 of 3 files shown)


allisonsuarez (approved):

common/amundsen_common/models/table.py
```diff
@@ -183,6 +183,7 @@ class Table:
     owners: List[User] = []
     watermarks: List[Watermark] = []
     table_writer: Optional[Application] = None
+    table_apps: Optional[List[Application]] = None
```
feng-tao, line 186: is it for backward compatible that we keep both table_writer and table_apps ? should we just change table_writer to `Optional[List[Application]]` type?

dkunitsk: yes. There is no way to repurpose the field without doing something much uglier (e.g. custom serializer in marshmallow). I looked hard.

feng-tao (approved):
sg

PR #1571, opened 2021-11-09
fix: #1570 - guard against errors from metadata service

### LINKED ISSUE #1570: Frontend swallows errors from metadata service on calls to /dashboard
## Expected Behavior

When an error occurs when the frontend calls the metadata service, it should bubble that error to the front end.

## Current Behavior

The frontend service assumes that the call to metadata succeeds and tries to extract the dashboard data from the json response, without checking if the body has the appropriate key.

https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/api/metadata/v0.py#L778-L779
```python
        response = request_metadata(url=url, method=request.method)
        dashboard = marshall_dashboard_full(response.json())
```

https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/api/utils/metadata_utils.py#L175-L176
```python
    dashboard_dict['owners'] = [_map_user_object_to_schema(owner) for owner in dashboard_dict['owners']]
    dashboard_dict['tables'] = [marshall_table_partial(table) for table in dashboard_dict['tables']]
```

## Possible Solution

Place a guard to assert the status code is 200, and only then decorate the data 

## Steps to Reproduce

1. Scrape Dashboards of some type (I have been using Mode)
2. In Neo4j, manually delete the Cluster associated with the dashboards (this will cause the metadata query for it to fail)
3 .Search for the dashboard in Amundsen, and click into the details for it.
4. You'll see the message "Something went wrong". Open the developer console and you'll see a Key Error for the request.

## Context

This swallowed the root error and basically increased debugging time.

## Your Environment

* Amunsen version used: 3.13.0
* Data warehouse stores: Neo4j
* Deployment (k8s or native): terraform
* Link to your fork or repository:

delwaterman:
### Summary of Changes

Fixes https://github.com/amundsen-io/amundsen/issues/1570. Guards against non 200 responses from the Metadata service.

### Tests

`frontend/tests/unit/api/metadata/test_v0.py`

### CheckList

### DIFF
--- a/frontend/amundsen_application/api/metadata/v0.py
+++ b/frontend/amundsen_application/api/metadata/v0.py
@@ -776,6 +776,8 @@ def get_dashboard_metadata() -> Response:
         url = f'{app.config["METADATASERVICE_BASE"]}{DASHBOARD_ENDPOINT}/{uri}'
 
         response = request_metadata(url=url, method=request.method)
+        if response.status_code != 200:
+            raise RuntimeError("Metadata service errored " + response.json()["message"])
         dashboard = marshall_dashboard_full(response.json())
         status_code = response.status_code
         return make_response(jsonify({'msg': 'success', 'dashboard': dashboard}), status_code)
--- a/frontend/tests/unit/api/metadata/test_v0.py
+++ b/frontend/tests/unit/api/metadata/test_v0.py
@@ -524,6 +524,10 @@ class MetadataTest(unittest.TestCase):
             "url": "test_dashboard_url"
         }
 
+        self.error_response = {
+            "message": "An error occurred on a service"
+        }
+
     @responses.activate
     def test_popular_resources_success(self) -> None:
         """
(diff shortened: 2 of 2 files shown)


frontend/amundsen_application/api/metadata/v0.py
```diff
@@ -776,6 +776,8 @@ def _get_dashboard_metadata(*, uri: str, index: int, source: str) -> None:
         url = f'{app.config["METADATASERVICE_BASE"]}{DASHBOARD_ENDPOINT}/{uri}'
 
         response = request_metadata(url=url, method=request.method)
+        if response.status_code != 200:
```
allisonsuarez, line 779: can you use `HTTPStatus.OK` instead of `200`?

frontend/amundsen_application/api/metadata/v0.py
```diff
@@ -776,6 +776,8 @@ def _get_dashboard_metadata(*, uri: str, index: int, source: str) -> None:
...
+            raise RuntimeError("Metadata service errored " + response.json()["message"])
```
allisonsuarez, line 780: We don't raise errors generally we catch the errors and then put them in the API response like we do for tables https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/api/metadata/v0.py#L172

allisonsuarez: have to make sure the error get's logged also

allisonsuarez (changes requested):

frontend/amundsen_application/api/metadata/v0.py
delwaterman: If you want to run with these changes go for it. But I made this PR  2 years ago and I can't really shift to working back on this project right now.

frontend/amundsen_application/api/metadata/v0.py
delwaterman: If you want to run with these changes go for it. But I made this PR  2 years ago and I can't really shift to working back on this project right now.

Issue #1572, opened 2021-11-09
Bug Report - Postegres / Redshift Extractors have inconsistent behavior


delwaterman:
## Expected Behavior

```python
# For RedshiftMetadataExtractor
redshift_metadata_extractor_config = ConfigFactory.from_dict({
    'extractor.redshift_metadata.where_clause_suffix':  'schema <> "foo"'
})
```

## Current Behavior

```python
# For RedshiftMetadataExtractor
redshift_metadata_extractor_config = ConfigFactory.from_dict({
    'extractor.redshift_metadata.where_clause_suffix':  'where schema <> "foo"'
})
```

## Possible Solution

Update redshift extractor to properly build suffix

## Context

Both the `PostgresMetadataExtractor` and the `RedshiftMetadataExtractor` take the config `extractor.redshift_metadata.where_clause_suffix` as they are both based off of `BasePostgresMetadataExtractor`.

However they require slightly different different sql partials, which is unexpected: 

The `where_clause_suffix` for `RedshiftMetadataExtractor` takes a partial sql statement **without** a `WHERE`.
The `where_clause_suffix` for `PostgresMetadataExtractor` take a partial sql statement **with** a `WHERE`.

The name `where_clause_suffix` implies that its a statement appended to a where clause and should  _not_ (IMHO) require a `WHERE` statement.

## Your Environment

* Amunsen version used: amundsen-databuilder 6.3.1
* Data warehouse stores: Neo4j
* Deployment (k8s or native): terraform
* Link to your fork or repository:

delwaterman:
While changing this behavior might break code for some users, I think that this difference in usage is confusing enough to newer users that its worth fixing.

delwaterman:
Um still waiting on the PR

delwaterman:
bump

delwaterman:
bump

PR #1573, opened 2021-11-09
fix: #1572 - make where_clause_suffix an actual suffix

delwaterman:
Signed-off-by: Orion Delwaterman <delwaterman@gmail.com>

### Summary of Changes

Fixes https://github.com/amundsen-io/amundsen/issues/1572

Changes the config of `RedshiftMetadataExtractor` `where_clause_suffix` to not require a user to include a `WHERE` statement in their sql partial. This follows the behavior of `PostgresMetadataExtractor`, and seems more consistent with the variable name `where_clause_suffix` by treating the partial as a _suffix_ to a where clause.

This will be a breaking change for anyone using redshift extractor (so probably will need a minor version bump). That being, IMHO, it's better to make the behavior follow expected patterns rather than ask users to hunt through source code to determine why their query is breaking (which happened to me).

### Tests

`databuilder/tests/unit/extractor/test_redshift_metadata_extractor.py`

### CheckList

### DIFF
--- a/databuilder/databuilder/extractor/redshift_metadata_extractor.py
+++ b/databuilder/databuilder/extractor/redshift_metadata_extractor.py
@@ -1,6 +1,7 @@
 # Copyright Contributors to the Amundsen project.
 # SPDX-License-Identifier: Apache-2.0
 
+import logging
 from typing import (  # noqa: F401
     Any, Dict, Iterator, Union,
 )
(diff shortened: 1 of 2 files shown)


dkunitsk:
Redshift extractor is reasonably popular and the style fix isn't worth breaking ingestor, imho. `where True` is always a workaround.

delwaterman:
If I may, I lost 1 hour of development tracking down this issue, which doesn't feel like a lot, but forcing a user to dig through source code to understand how to use a public API is bad form. Personally it led me to trust this code base less, because it felt sloppy. I am not trying to pick a fight, but I am trying to tell my experience as a new developer using the project

That being said, why don't we do a comprimise, and support the old pattern, with a deprecation warning:

```python
if where_clause_suffix.lower().startswith("where"):
    logger.Warn("you no longer need to begin with 'where' in your suffix")
    where_clause = where_clause_suffix
else:
    where_clause = f"where {where_clause_suffix}"
```

verdan:
This is kind of an issue we had for Postgres, and we fixed it here: https://github.com/amundsen-io/amundsen/commit/6cf8965ce8b065cc68be8fe021b1e7496b4b7716#diff-01ee0656007972e26ae47d9003cf877aefa524b6e91a6bf6b72035892ba9b757R47

So a suggestion is to make the `where_clause_suffix` a real suffix, what @delwaterman is suggesting, and make the default value as `true`.

delwaterman:
Ok I will do the update

delwaterman:
@verdan @dkunitsk This is now ready

delwaterman:
Bump

delwaterman:
bump

dkunitsk (approved):

PR #1574, opened 2021-11-09
fix: database mapping superset extract

mgorsk1:
### Summary of Changes

Fixes redundant dict wrap messing with input dict.

### Tests

n/a

### Documentation

n/a

### CheckList


feng-tao:
CI fails

cccs-eric:
@mgorsk1 After removing the `dict` wrap, you end up with a `pyhocon.config_tree.ConfigTree` object and its `get()` method expects a string or bytes-like object.  Since we a doing a `get(database_id)` and `database_id` is an `int`, it doesn't work.  We need to convert `database_id` to a string:
```
cluster = self.cluster_mapping.get(str(database_id), self.cluster)
```

We don't have this problem with `self.driver_mapping.get(database_spec.drivername, database_spec.drivername)` because `database_spec.drivername` is a string.

dechoma (approved):

PR #1577, opened 2021-11-10
Feat: Further generalize applications

dkunitsk:
Continuing the work started in https://github.com/amundsen-io/amundsen/pull/1569. To recap, we'd like to generalize `Application` to support multiple apps per resource and add "consumes" as a new possible relationship type in addition to "generates". See design in the earlier PR. 

Here we have generalized the databuilder model and updated the neo4j proxy. Other proxies will keep their current behavior of fetching and returning one app as the `table_writer`.

This PR takes advantage of the fact that the Application model had [an unused field](https://github.com/amundsen-io/amundsen/blob/main/common/amundsen_common/models/table.py#L78) named `kind` which we're using to specify the relationship (consuming or producing).

Note about bw-compatibility:
**Most** of this PR is fully backwards-compatible: `table_apps` is an optional field and `table_writer` continues to be set as before. The refactor of `Application` into a subclass of `GenericApplication` should be fully transparent. 

However, because we doubled down on extending `Application` beyond Airflow, this unfortunately is incompatible with the earlier changes made in that same direction by @jroof88: https://github.com/amundsen-io/amundsen/pull/1398. Those who have started using `Application` for ingesting non-Airflow apps will have a one-time migration cost to `GenericApplication`. The benefit is in not having to forever work around Airflow specifics like `dag_id` etc. 

### DIFF
--- a/databuilder/databuilder/models/application.py
+++ b/databuilder/databuilder/models/application.py
@@ -1,7 +1,9 @@
 # Copyright Contributors to the Amundsen project.
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Iterator, Union, Optional
+from typing import (
+    Iterator, Optional, Union,
+)
 
 from amundsen_common.utils.atlas import (
     AtlasCommonParams, AtlasCommonTypes, AtlasTableTypes,
--- a/databuilder/databuilder/models/application.py
+++ b/databuilder/databuilder/models/application.py
@@ -1,7 +1,7 @@
 # Copyright Contributors to the Amundsen project.
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Iterator, Union
+from typing import Iterator, Union, Optional
 
 from amundsen_common.utils.atlas import (
     AtlasCommonParams, AtlasCommonTypes, AtlasTableTypes,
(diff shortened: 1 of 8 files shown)


databuilder/databuilder/models/application.py
```diff
@@ -21,88 +21,51 @@
 from databuilder.utils.atlas import AtlasRelationshipTypes, AtlasSerializedEntityOperation
 
 
-class Application(GraphSerializable, TableSerializable, AtlasSerializable):
+class GenericApplication(GraphSerializable, TableSerializable, AtlasSerializable):
```
allisonsuarez, line 24: I think just calling it Application is generic enough

feng-tao: +1, do we need to change the model name?

dkunitsk: Ah yes, I wish that was an option. This is being done for bw-compatibility. The Airflow-specific class was/is called `Application` and I have kept that working via https://github.com/amundsen-io/amundsen/pull/1577/files#diff-9ecb77ae83ba63ef4784084c73ecfd0be606ba36d7dcf640e424d9f38dfc3fc6R240.

dkunitsk: In other words, 
- `GenericApplication` is completely new class
- `Application` class is renamed `AirflowApplication`
- `AirflowApplication` is aliased as `Application` for compatibility

databuilder/tests/unit/models/test_application.py
```diff
@@ -93,21 +95,35 @@ def setUp(self) -> None:
...
+            table_key = TableMetadata.TABLE_KEY_FORMAT.format(
+                db='hive',
+                cluster='gold',
                 schema='default',
-                table_name='test_table',
-                application_url_template=url,
+                tbl='test_table',
+            )
+
+            application = GenericApplication(
```
dkunitsk, line 118: @jroof88 This shows what change you'd have to make on your end.
The key difference is that the key you introduced in https://github.com/amundsen-io/amundsen/pull/1398 `application://{application_type}/{database}/{table}` is replaced by an ability to specify a key override. 

The new generic key has been changed to `application://{application_type}/{application_id}`. Because the key is independent of the table, it lets the generic model support many<>many relationships out of the box.

metadata/metadata_service/api/swagger_doc/template.yml
```diff
@@ -485,11 +492,6 @@ components:
           description: '[DEPRECATED] Chart names associated with the Dashboard'
           items:
             type: string
-        frequent_users:
```
dkunitsk, line 488: this was listed twice

metadata/tests/unit/proxy/test_neo4j_proxy.py
```diff
@@ -1456,5 +1460,65 @@ def test_get_resource_generation_code_not_found(self) -> None:
                               resource_type=ResourceType.Feature)
 
 
+class TestNeo4jProxyHelpers:
```
dkunitsk, line 1463: The above class feels more like an integration test. Created this one for testing helpers.

databuilder/tests/unit/models/test_application.py
```diff
@@ -26,7 +28,7 @@
 
 @dataclass
 class ApplicationTestCase:
```
dkunitsk, line 30: Notice that the only change here is migrating non-airflow apps to `GenericApplication`. Everything else is fully bw-compatible. 

Since extension of this model to non-airflow is recent (mid-August), disruption is likely minimal.

dkunitsk:
@feng-tao ptal again

jroof88:
PR LGTM. I never ended up using my Application change (yet) so it'll be good to get this through before I do.

feng-tao:
change overall lgtm without going through the details as I think the change will help to support: 1. multiple tasks within same app on same table; 2. different apps (or different workloads) on the same table

allisonsuarez (approved):

PR #1578, opened 2021-11-11
feat: Frontend work to support multiple generalized applications

kristenarmes:

### Summary of Changes

This frontend change supports the changes made in https://github.com/amundsen-io/amundsen/pull/1577. Currently, there is a link that will be displayed if there is a table writer application in the table metadata but the page does not support multiple applications. This change modifies the link to become a dropdown instead. It will take the existing `table_writer` field and the new optional `table_apps` field, and it will use the information to render one or more dropdown components where the link used to be depending on the amount of types of applications included in the list.

This will support Airflow and Databricks as their own dropdowns since those were the previously supported applications. Any that fall outside of those two would be included in a separate generic dropdown list.

The dropdown(s) will group the applications by their `kind`, such as Producing or Consuming, and then by `id` within those groupings. Any applications in the generic dropdown will be grouped by `name`, then `kind`, then `id`.

### Tests

Added JS tests for the `ApplicationDropdown`.

### Documentation

N/A

### CheckList


frontend/amundsen_application/proxy/issue_tracker_clients/jira_client.py
```diff
@@ -160,7 +162,7 @@ def create_issue(self,
             }, labels=self.issue_labels,
                 summary=title,
                 description=(f'{description} '
-                             f'\n *Reported By:* {user_email} '
+                             f'\n *Reported By:* {reporting_user_str if reporting_user_str else user_email} '
```
kristenarmes, line 165: This is a small unrelated change to follow up on previous frontend Jira work. This will link the profile url for the user who created the Jira ticket rather than only displaying their email.

frontend/amundsen_application/static/js/interfaces/TableMetadata.ts
```diff
@@ -102,6 +103,7 @@ export interface TableMetadata {
   name: string;
   description: string;
   table_writer: TableWriter;
```
dkunitsk, line 105: would it be difficult to change the name from TableWriter to App/TableApp for the benefit of future maintainers.

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const APPLICATIONS_LABEL = 'Applications';
```
dkunitsk, line 1: nit about the filename: WriterDropdown to AppDropdown or some such

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  return (
+    <MenuItem
+      className={menuItemClassName}
+      href={app.application_url}
+      onClick={logClick}
+      target="_blank"
+    >
+      <div className="writer-dropdown-menu-item-row">
+        <span className="menu-item-content">
+          {app.id ? app.id : NONE_VALUE}
```
dkunitsk, line 83: there will always be an app id

frontend/amundsen_application/static/images/icons/Application.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Golodhros, line 1: What does this icon look like?

I would expect it to be named 'application.svg' instead.

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+const getMenuItem = (app: TableWriter, menuItemClassName) => {
+  if (app.name.toLowerCase() === AIRFLOW) {
+    const [dagId, ...task] = app.id.split('/');
+    const taskId = task.join('/');
+    return (
+      <MenuItem
+        className={menuItemClassName}
+        href={app.application_url}
+        onClick={logClick}
+        target="_blank"
```
Golodhros, line 61: Any link that has target="_blank" needs to have rel="noopener noreferrer" for security reasons.
I see this is not exactly a link, but I guess is kind of?

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+      </MenuItem>
+    );
+  }
+  return (
+    <MenuItem
+      className={menuItemClassName}
+      href={app.application_url}
+      onClick={logClick}
+      target="_blank"
+    >
```
Golodhros, line 80: Nit: It seems this gets reused on both cases, so I would keep it here and do something like:
```
<MenuItem
      className={menuItemClassName}
      href={app.application_url}
      onClick={logClick}
      target="_blank"
    >
        {isAirflow ? (
           //airflow menu-item-rows here
        ) : (
           //non-airflow menu item row
        )}
```

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+    </MenuItem>
+  );
+};
+
+const WriterDropdown: React.FC<WriterDropdownProps> = ({
+  tableApps,
+}: WriterDropdownProps) => {
+  const tableAppName = tableApps[0].name.toLowerCase();
+
+  let image = '';
```
Golodhros, line 95: Let's extract this into a helper function, something like `getImagePath(tableAppName)`.
I would also wrap it into a `useCallback` function(https://www.robinwieruch.de/react-usecallback-hook), as this gets executed on every component re-render.

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+      break;
+    case DATABRICKS:
+      image = '/static/images/icons/logo-databricks.png';
+      break;
+    default:
+      image = '/static/images/icons/Application.svg';
+  }
+
+  tableApps.sort(
+    (a, b) => a.name.localeCompare(b.name) || a.id.localeCompare(b.id)
```
Golodhros, line 108: I would extract this oneliner as `sortByNameOrId`

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  );
+  const appNames: string[] = [...new Set(tableApps.map((app) => app.name))];
```
Golodhros, line 110: Nit:
```suggestion
  const appNames: string[] = [...new Set(tableApps.map(({name}) => name))];
```

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const menuItems: JSX.Element[] = [];
```
Golodhros, line 112: I usually don't use JSX.Element, I think React.ReactNode will work better.

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+      menuItems.push(
+        <h5 className="writer-dropdown-menu-title">{sectionTitle}</h5>
+      );
+      menuItems.push(
+        ...tableApps
+          .filter(
+            (app) =>
+              app.name === name &&
+              ((app.kind && app.kind === kind) ||
+                (!app.kind && kind === PRODUCING))
```
Golodhros, line 131: Could you summarize this predicate into something human readable?
Maybe `hasSameNameAndKind` or similar?

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+      tableApps.filter((app) => app.name === name)
+    );
+    appKinds.forEach((kind, kindIdx) => {
+      const sectionTitle = isAirflowOrDatabricksApp(tableApps[0].name)
+        ? kind
+        : name + ' - ' + kind;
+      menuItems.push(
+        <h5 className="writer-dropdown-menu-title">{sectionTitle}</h5>
+      );
+      menuItems.push(
```
Golodhros, line 125: Same comment about `push`, there isn't a good reason to use it in most of the cases.

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+            (app) =>
+              app.name === name &&
+              ((app.kind && app.kind === kind) ||
+                (!app.kind && kind === PRODUCING))
+          )
+          .map((app) => {
+            const menuItemClassName =
+              ++appsDefinedCount === tableApps.length
+                ? 'writer-dropdown-menu-item last-item'
+                : '';
```
Golodhros, line 137: Hmm... this was interesting, could you expand a bit why we do it?

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/styles.scss
```diff
@@ -0,0 +1,57 @@
...
+  .writer-dropdown-menu-title {
+    margin: $spacer-2 $spacer-2 0 $spacer-2;
+    line-height: $spacer-2;
+    text-transform: uppercase;
+    font-size: $menu-title-font-size;
+    white-space: nowrap;
+  }
+
+  .writer-dropdown-menu-item {
+    &.last-item {
```
Golodhros, line 36: You might not need to add this class by using https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child, give it a try!

frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -341,6 +343,50 @@ export class TableDetail extends React.Component<
...
+
+    if (airflowApps.length > 0) {
+      writerDropdowns.push(<WriterDropdown tableApps={airflowApps} />);
+    }
+    if (databricksApps.length > 0) {
+      writerDropdowns.push(<WriterDropdown tableApps={databricksApps} />);
+    }
+    if (remainingApps.length > 0) {
+      writerDropdowns.push(<WriterDropdown tableApps={remainingApps} />);
+    }
```
Golodhros, line 385: I think I see what you are trying to do here, but it feels non-idiomatic in the React world.
Check out this resource: https://reactjs.org/docs/conditional-rendering.html

We tend to favor the 'Inline If with Logical && Operator' method nowadays.

frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -341,6 +343,50 @@ export class TableDetail extends React.Component<
...
   }
 
+  renderTableWriterDropdowns(tableWriter, tableApps) {
+    const apps: TableWriter[] = [];
+    if (
+      (tableApps === null || tableApps.length === 0) &&
+      tableWriter === null
+    ) {
+      return null;
+    }
```
Golodhros, line 353:
```suggestion
const hasNoAppsOrWriter = (tableApps === null || tableApps.length === 0) &&
      tableWriter === null;
    if (hasNoAppsOrWriter) {
      return null;
    }
```

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/styles.scss
```diff
@@ -0,0 +1,57 @@
...
+
+    .menu-item-content {
+      line-height: $menu-item-content-height;
+      max-width: fit-content;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+  }
+}
```
Golodhros, line 57: This looks great, awesome job!

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const appNames: string[] = [...new Set(tableApps.map((app) => app.name))];
+
+  const menuItems: JSX.Element[] = [];
+  let appsDefinedCount = 0;
+  appNames.forEach((name, nameIdx) => {
+    const appKinds = getSortedAppKinds(
+      tableApps.filter((app) => app.name === name)
+    );
+    appKinds.forEach((kind, kindIdx) => {
+      const sectionTitle = isAirflowOrDatabricksApp(tableApps[0].name)
```
Golodhros, line 119: I wonder if it would be better to have just different menuItem components for the different apps.
Even if there is some duplication on the JSX markup, we would probably avoid some of the logic I we are using on this PR.

Think a branching like:
```
<Dropdown.Menu className="writer-dropdown-menu">
        {
  if (isAirflow) {
    return <AirFlowMenuItem />;
  } else if (isDatabricks) {
    return <DataBricksMenuItem  />;
  } else {
    return <GenericMenuItem />;
  }}
 </Dropdown.Menu>
```

frontend/amundsen_application/static/images/icons/Application.svg
kristenarmes: This icon looks like a gear with a square around it to represent generic applications (designed by Knowl). I renamed the file to be lowercase now, I noticed there was a mixture of some upper and some lower for the icon naming and chose wrong 😄

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
kristenarmes: I realized after you pointed this out that this was unnecessary - I was originally adding a `margin-bottom` to the last item in the list, but instead I just removed this extra work and added padding to the bottom of the dropdown itself

frontend/amundsen_application/static/js/pages/TableDetailPage/WriterDropdown/index.tsx
kristenarmes: This is a good suggestion, it will make it easier/cleaner too if anyone else in the future wants to add special styling for a separate application. I'll make this change in the remaining work on this PR

dkunitsk:
Some thoughts after trying this out internally:
- could the dropdown be shifted left (i.e. the right edge of the dropdown aligned with the chevron) and be made wider so that more of the text is visible
- I'm thinking 'not available' is actually a better and more clear placeholder than 'none' for task_id.

frontend/amundsen_application/static/js/pages/TableDetailPage/ApplicationDropdown/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+import './styles.scss';
+
+export interface ApplicationDropdownProps {
+  tableApps: TableApp[];
+}
+
+const getImagePath = (tableAppName) => {
+  switch (tableAppName) {
+    case AIRFLOW.toLowerCase():
+      return '/static/images/airflow.jpeg';
```
Golodhros, line 29: Nit: I would move this strings into constants, something like:
const AIRFLOW_LOGO_PATH
const DATABRICKS_LOGO_PATH
...

frontend/amundsen_application/static/js/pages/TableDetailPage/ApplicationDropdown/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  return [...producingKind, ...consumingKind, ...remainingKinds];
+};
+
+const handleClick = (event) => {
+  event.stopPropagation();
+  logClick(event);
+};
+
+const getDropdownMenuContents = (tableApps) => {
+  if (tableApps[0].name.toLowerCase() === AIRFLOW.toLowerCase()) {
```
Golodhros, line 77: Nit:
const isAirFlow = tableApps[0].name.toLowerCase() === AIRFLOW.toLowerCase();
const isDataBricks = tableApps[0].name.toLowerCase() === DATABRICKS.toLowerCase();

if (isAirFlow) { ...

frontend/amundsen_application/static/js/pages/TableDetailPage/ApplicationDropdown/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+      hasSameNameAndKind={hasSameNameAndKind}
+      handleClick={handleClick}
+    />
+  );
+};
+
+const ApplicationDropdown: React.FC<ApplicationDropdownProps> = ({
+  tableApps,
+}: ApplicationDropdownProps) => {
+  if (tableApps === null || tableApps.length === 0) return null;
```
Golodhros, line 110: We don't do one liners like these, is an historic pattern we follow:
```suggestion
  if (tableApps === null || tableApps.length === 0) {
      return null;
  }
```

frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -341,6 +343,50 @@ export class TableDetail extends React.Component<
...
+    if (hasNoAppsOrWriter) {
+      return null;
+    }
+    if (tableApps !== null && tableApps.length > 0) {
+      apps = [...tableApps];
+    }
+    if (
+      tableWriter !== null &&
+      !apps.some((app) => app.id === tableWriter.id)
+    ) {
```
Golodhros, line 359: Would love to have another 'isNotWriterAnd...' here too

Golodhros (commented):
Looks much better!

Only some minor comments

Golodhros (approved):
LGTM

Thanks!

PR #1604, opened 2021-12-03
fix: Adding missing secretName in ingress configuration.  Helm improvements proposal.

cccs-eric:
### Summary of Changes

This PR adds support for `secretName` in the ingress specification.  The values.yaml contained an example, but it was not supported in the ingress template.

I am also proposing to improve the helm chart, but this could be done in a separate PR if preferred.  But I'd like to know if what I'm proposing is accepted by the community:
- Move dependencies to `Chart.yaml` and drop `requirements.yaml` (effectively adopting the new helm standard)
- Split configmap into individual ones and added checksum for deployment reload on change.  
- Added metadata and search service config class support.
- Adopt [Bitnami Common helm chart](https://github.com/bitnami/charts/tree/master/bitnami/common) as a dependency to improve chart.  Their charts are well done and they have nailed a few patterns.  For example, I am showcasing in my PR their template to resolve the proper version of the ingress spec according to the version of k8s being used.  They have an Apache License, so I think we are good to use it.  But it would be as a dependency, so I am not sure about Amundsen's stance on this...  This is a proposal and if accepted, I could go ahead and make more changes.

### CheckList


cccs-eric:
@pengfei99 I'm adding support for secretName to your previous PR.  I think it was forgotten.  I'm also fixing a bug with your template call and `$dot` string.

@feng-tao Can I have your opinion on using Bitnami's charts to improve Amundsen's?  It's only a proposal, I can remove those changes from my PR.

pengfei99:
Hi @cccs-eric,

1. Thanks for the fix on sectetName. I forgot it.
 
2. For removing the requirement.yaml and including it into Chart.yaml part. I'm OK. I choose to use the requirement.yaml because the origin file layout was like that. 

3. For the $dot, it's not a bug, I have successfully deployed the chart many times. I use $dot. Because "." in a range clause change context in each iteration that will cause problems. $dot was a workaround of this problem. For more details about this bug. https://stackoverflow.com/questions/54180804/chart-name-cant-evaluate-field-chart-in-type-string  
I have not tested your solution yet. If it works, why not.  I don't like this $dot workaround

BR

Pengfei

pengfei99:
Great, I did not know this. I will change all my charts too. Tks

cccs-eric:
I was hoping to get a comment about the Bitnami chart adoption before making this PR ready, but since I'd like to get this done, I'll mark is a Ready to hopefully get some attention.  As I mentioned earlier, I can easily remove the Bitnami aspect of this PR.

feng-tao (approved):

PR #1611, opened 2021-12-07
feat: v2/document API

allisonsuarez:

### Summary of Changes
- Added /v2/document API
- Added methods to update ES documents using elasticsearch-dsl lib
- Added ES mappings


search/search_service/api/document_update.py
```diff
@@ -0,0 +1,44 @@
...
+from amundsen_common.models.search import UpdateDocumentRequestSchema
+from search_service.proxy import get_proxy_client
+from search_service.proxy.es_search_proxy import RESOURCE_STR_MAPPING
+
+
+class DocumentAPI(Resource):
+
+    def __init__(self) -> None:
+        self.proxy = get_proxy_client()
+        self.request = UpdateDocumentRequestSchema().loads(json.dumps(request.get_json()))
```
allisonsuarez, line 19: Accept add or overwrite option to explicitly decide the behavior, either append a value so there are multiple values for a field or overwrite it completely

allisonsuarez: Support multiple values per field so [] instead of just oen value

allisonsuarez: Support updating multiple fields in one single request, {tag: ['t1', 't2'], description: 'i am a table'}

allisonsuarez: ignore last 2, double check the behavior for why we default to empty array for badges and tags, does ES support multi value in a field by default?

search/search_service/proxy/es_mappings.py
```diff
@@ -0,0 +1,75 @@
...
+    badges = Keyword(multi=True)
+    cluster = Text(fields={'raw': Keyword()}, analyzer='simple')
+    database = Text(fields={'raw': Keyword()}, analyzer='simple')
+    key = Keyword()
+    total_usage = Long()
+    unique_usage = Long()
+    programmatic_descriptions = Text(analyzer='simple', multi=True)
+
+
+class Dashboard(Document):
```
allisonsuarez, line 40: we need to add key to dashboard mapping, currently cant update dashboard because of this https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/api/metadata/v0.py#L447 @dkunitsk

search/search_service/api/document_update.py
```diff
@@ -0,0 +1,47 @@
...
+                                                    operation=self.request.operation,
+                                                    delete=False)
+            return {'msg': resp}, HTTPStatus.OK
+        except Exception as e:
+            err_msg = f'Failed to update the field value: {e}'
+            return {'message': err_msg}, HTTPStatus.INTERNAL_SERVER_ERROR
+
+    def delete(self) -> Tuple[Any, int]:
+        try:
+            resp = self.proxy.update_document_field(resource_key=self.request.resource_key,
```
dkunitsk, line 38: imo having a separate method makes more sense than delete=True

search/search_service/proxy/es_mappings.py
```diff
@@ -0,0 +1,75 @@
...
+                                  type='custom',
+                                  char_filter=[],
+                                  filter=['lowercase', 'asciifolding'])
+feature_lowercase_normalizer = normalizer(name_or_instance='lowercase_normalizer',
+                                          type='custom',
+                                          char_filter=[],
+                                          filter=['lowercase', 'asciifolding'])
+
+
+class Table(Document):
```
dkunitsk, line 21: can you remind me why moving this here is a good idea

search/search_service/proxy/es_search_proxy.py
```diff
@@ -317,3 +336,101 @@ def search(self, *,
...
+        if response.success():
+            results_count = response.hits.total.value
+            if results_count > 0:
+                es_result = response.hits.hits[0]
+                resource_es_id = es_result._id
+
+                # return document
+                resource_str = resource_type.name.lower()
+                resource_index = f"{resource_str}_search_index"
+                return RESOURCE_TO_ES_MAPPING[resource_type].get(id=resource_es_id,
```
dkunitsk, line 359: instead of key->id->doc, can we just go key->doc

search/search_service/proxy/es_mappings.py
allisonsuarez: we are not moving it the es mappings we use for databuilder are still in common, I can move this there too

search/search_service/proxy/es_search_proxy.py
allisonsuarez: skip the doc get, just doc update when getting the ID and the current value of the field

search/search_service/proxy/es_mappings.py
allisonsuarez: do not use dsl library to update, since we are not using documents for anything else this is really heavy weight

search/search_service/proxy/es_search_proxy.py
```diff
@@ -319,3 +326,144 @@ def search(self, *,
...
+            resource_type: Q(TERM_QUERY, key=resource_key),
+        }
+        response: Response = self.execute_queries(queries=key_query,
+                                                  page_index=0,
+                                                  results_per_page=1)
+        if len(response) < 1:
+            msg = f'No response from ES for key query {key_query[resource_type]}'
+            LOGGER.error(msg)
+            raise ElasticsearchException(msg)
+        response = response[0]
```
dkunitsk, line 344: also error if >1. Response should have exactly 1 hit

search/search_service/proxy/es_search_proxy.py
```diff
@@ -319,3 +326,149 @@ def search(self, *,
...
+            raise ElasticsearchException(msg)
+        if len(response) > 1:
+            msg = f'Key {key_query[resource_type]} is not unique to a single ES resource'
+            LOGGER.error(msg)
+            raise ElasticsearchException(msg)
+
+        response = response[0]
+        if response.success():
+            results_count = response.hits.total.value
+            if results_count > 0:
```
dkunitsk, line 352: wait weren't you supposed to do that here

allisonsuarez: oh my bad i got confused, checking the number o responses from m search first, then checking the number of results in the particular response, I'll fix it

search/search_service/proxy/es_search_proxy.py
```diff
@@ -319,3 +326,150 @@ def search(self, *,
...
+            LOGGER.error(msg)
+            raise InternalServerError(msg)
+
+    def update_es_document(self, *,
+                           resource_type: Resource,
+                           field: str,
+                           new_value: Union[List, str, None],
+                           document_id: str) -> None:
+        resource_str = resource_type.name.lower()
+        resource_index = f"{resource_str}_search_index"
```
dkunitsk, line 376: abstract out

search/search_service/proxy/es_search_proxy.py
```diff
@@ -319,3 +326,150 @@ def search(self, *,
                                                    resource_types=resource_types)
 
         return formatted_response
+
+    def get_document_id_and_field_value_from_key(self,
```
dkunitsk, line 330: just `get_document_json_by_key` would correctly encapsulate the re-usable part of this function. As for the id and field value, that's better done with another function, maybe called `get_field_from_document_json` or just a .get if the output of the first func is simple

search/search_service/proxy/es_search_proxy.py
```diff
@@ -319,3 +326,150 @@ def search(self, *,
...
+        partial_document = {
+            "doc": {
+                field: new_value
+            }
+        }
+        self.elasticsearch.update(index=resource_index,
+                                  id=document_id,
+                                  body=partial_document)
+
+    def update_document_field(self, *,
```
dkunitsk, line 386: `update_doc_by_key`

search/search_service/proxy/es_search_proxy.py
```diff
@@ -319,3 +326,150 @@ def search(self, *,
...
+                # no doc exists with given key in ES
+                msg = f"Requested key {resource_key} query returned no results in ES"
+                LOGGER.error(msg)
+                raise ValueError(msg)
+        else:
+            msg = f'Request to Elasticsearch failed: {response.failures}'
+            LOGGER.error(msg)
+            raise InternalServerError(msg)
+
+    def update_es_document(self, *,
```
dkunitsk, line 370: `update_doc_by_id`

dkunitsk (approved):

PR #1605, opened 2021-12-03
feat: Adding single apply and clear buttons to apply to all search filters

kristenarmes:
Signed-off-by: Kristen Armes <karmes@lyft.com>

### Summary of Changes

Currently when filtering on the search page, there are `Apply` and `Clear` buttons for every filter section. Only one filter category/value pair could be passed at a time to the search API due to frontend constraints. This change adds just one `Apply filters` button and one `Clear all` button at the bottom of the filters side panel which allows multiple filters to be passed to the API at the same time.

Another change that was made is related to the `Clear all` behavior. The global state was not aware of the filters in the input text fields until they were applied, so with the new UI changes the single clear button did not work at first if the filters had been typed in but not applied. Now the global state is updated whenever anything is typed into the input fields so that the `Clear all` button will work at any time.

### Tests

Since the `Apply filters` and `Clear all` buttons were moved to a higher level component than they were previously a part of, some tests for those were moved to `SearchFilter` where they are now rendered. A test was also added to ensure the filters global state is updated on any input changes.

### Documentation

N/A

### CheckList
Make sure you have checked **all** steps below to ensure a timely review.
- [X] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).
- [X] PR includes a summary of changes.
- [X] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [X] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public functions and the classes in the PR contain docstrings that explain what it does

PR #1612, opened 2021-12-08
fix: Search filtering behavior fix

kristenarmes:

### Summary of Changes

Fixes related to recent changes to search filtering here: https://github.com/amundsen-io/amundsen/pull/1605
- The original change checked whether the `Apply filters` button had been clicked to know if a search had occurred to provide the correct messaging to the user on the page. This didn't account for searches done from other locations, such as clicking on a tag on the home page, and this resulted in an unintended effect where the results were retrieved but not displayed. This update instead sets a boolean state value in the reducer that will tell if a search action has been taken, compared to an update to the global state without a search action (such as updating the filters).
- When the filters global state was updated on any input field changes, the original change was unintentionally updating the global state object reference, causing the filters to not be reset when they should have been. This change properly copies the existing state to a new object and appends the updated filter.

### Tests

Small changes to existing tests to account for the changes noted in the summary above.

### Documentation

N/A

### CheckList

### DIFF
--- a/frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/InputFilter/index.tsx
+++ b/frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/InputFilter/index.tsx
@@ -63,8 +63,11 @@ export class InputFilter extends React.Component<
 
     this.setState({ value: newValue });
 
-    filterState[resourceType][categoryId] = newValue;
-    updateFilterState(filterState);
+    const newFilters = {
+      ...filterState,
+      [resourceType]: { ...filterState[resourceType], [categoryId]: newValue },
+    };
+    updateFilterState(newFilters);
   };
 
   render = () => {
--- a/frontend/amundsen_application/static/js/pages/SearchPage/index.tsx
+++ b/frontend/amundsen_application/static/js/pages/SearchPage/index.tsx
@@ -119,11 +119,10 @@ export class SearchPage extends React.Component<SearchPageProps> {
     const startIndex = RESULTS_PER_PAGE * page_index + 1;
     const tabLabel = this.generateTabLabel(tab);
 
-    // No search input and search action has not been taken
-    if (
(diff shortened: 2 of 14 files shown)


frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/InputFilter/index.tsx
```diff
@@ -63,8 +63,11 @@ export class InputFilter extends React.Component<
 
     this.setState({ value: newValue });
 
-    filterState[resourceType][categoryId] = newValue;
-    updateFilterState(filterState);
+    const newFilters = {
+      ...filterState,
+      [resourceType]: { ...filterState[resourceType], [categoryId]: newValue },
+    };
```
Golodhros, line 69: Yep, mutation is better to be avoided. I hope it didn't take you too long to find this out!

frontend/amundsen_application/static/js/pages/SearchPage/index.tsx
```diff
@@ -132,14 +114,16 @@ export class SearchPage extends React.Component<
...
-    const { didApplyFilters } = this.state;
-    const { hasFilters, searchTerm, setPageIndex } = this.props;
+    const { hasFilters, searchTerm, setPageIndex, didSearch } = this.props;
     const { page_index, total_results } = results;
     const startIndex = RESULTS_PER_PAGE * page_index + 1;
     const tabLabel = this.generateTabLabel(tab);
 
-    // No search input
-    if (searchTerm.length === 0 && (!hasFilters || !didApplyFilters)) {
+    // No search input and search action has not been taken
```
Golodhros, line 122: I would rather read this as the name of a variable, something like:
```const hasNoSearchInputOrAction = searchTerm.length === 0 &&
      (!hasFilters || (!didSearch && total_results === 0));
      ```
      
      or something similar

Golodhros (approved):
LGTM

Only minor comment

PR #1619, opened 2021-12-13
feat: FE document update with v2/document search service endpoint use

allisonsuarez:

### Summary of Changes

- Search service now takes a single field and value, as well as the resource key to update that single field rather than receiving the entire document, modifying it and then replacing it entirely on ES.

### Tests

- updated existing unit tests


frontend/amundsen_application/api/metadata/v0.py
```diff
@@ -525,8 +459,13 @@ def _log_update_table_tags(*, table_key: str, method: str, tag: str) -> None:
         _log_update_table_tags(table_key=table_key, method=method, tag=tag)
 
         metadata_status_code = _update_metadata_tag(table_key=table_key, method=method, tag=tag)
-        search_status_code = _update_search_tag(
-            key=table_key, resource_type=ResourceType.Table, method=method, tag=tag)
+        search_method = method if method == 'DELETE' else 'POST'
```
allisonsuarez, line 462: Method for updating now POST because it has the capability to update and create new fields for documents on ES

dkunitsk: does FE not post to this endpoint as well

frontend/amundsen_application/api/metadata/v0.py
```diff
@@ -525,8 +459,13 @@ def _log_update_table_tags(*, table_key: str, method: str, tag: str) -> None:
...
+        search_status_code = update_search_field(key=table_key,
+                                                 resource_type=ResourceType.Table.name.lower(),
+                                                 field='tag',
+                                                 value=tag,
+                                                 operation='add',
```
dkunitsk, line 467: kinda odd to do operation='add' and method=delete no?

allisonsuarez: this all needs to change based on the feedback from search I implemented over last week

frontend/amundsen_application/api/metadata/v0.py
allisonsuarez: no it PUTs and I decided not to change it because I would have to update metadata to do the same as well

youngyjd (approved):
lgtm

PR #1623, opened 2021-12-16
feat: Use new search endpoint in FE and add search filter operation selection

kristenarmes:
### Summary of Changes

This change has three major sections that are all included together for compatibility:
1. Work done by @allisonsuarez to have the frontend service point to the new search endpoint.
2. Search input filters now accept comma separated values rather than only a single value.
3. When a comma to separate multiple filter terms is typed in any input filter box, a `AND/OR` toggle will appear, allowing the user to specify whether they are filtering by all of the entered filter terms in the same results, or just looking for at least one of the terms in the results.

Some design decisions to note:
1. There is a new optional frontend config in the `BaseFilterCategory` called `allowableOperation`. If it is not set, the default behavior will allow both `AND` and `OR` operations for filtering. If it is set to `FilterOperationType.OR`, only the `OR` filter operation will be enabled for that category, which can be used when each search result would only ever include one of the filter terms in the category - e.g. a given table would only have one source database. If it is set to `FilterOperationType.AND`, only the `AND` filter operation will be enabled for that category.
2. A special type called `FilterOptions` was used to keep track of which checkbox filters were checked or unchecked where a `true` value meant it should be checked and a `false` value unchecked. Since now the input filters accept multiple values, this type was removed and the checkboxes are handled the same way as the input filters, where the name of the checkbox being included in the filter means it should be checked and any not included are unchecked.
3. Since there is now only one search API call per search rather than one per resource, when a search action for a specific resource is taken, only that resource has its state updated from the response to avoid clearing out the previous search for the other resources.
4. Previously, clearing filters would trigger an empty search in order to clear out the results. With the new API, this was causing an error because it did not expect to get empty searches, so instead of doing an unnecessary API call when clearing filters it will just clear the results state instead.

### Tests

Various changes were made to JS tests to account for changes to the way the frontend data is handled with filtering, such as checkboxes using a list of values rather than a special type to make it more consistent with the input filters. New tests were added for the `AND/OR` toggle behavior.

### Documentation

`frontend/docs/application_config.md` was updated to account for the renamed filter category types, along with fixing outdated links.

### CheckList


frontend/amundsen_application/static/js/components/Badges/BadgeList/index.tsx
```diff
@@ -69,6 +69,11 @@ export default class BadgeList extends React.Component<BadgeListProps> {
       <span className="badge-list">
         {alphabetizedBadges.map((badge, index) => {
           let badgeConfig;
+          // search case
+          if (typeof badge === 'string' || badge instanceof String) {
```
Golodhros, line 73: Curious about the double check, just with the typeof wasn't enough?

frontend/amundsen_application/static/js/components/Badges/BadgeList/index.tsx
```diff
@@ -69,6 +69,11 @@ export default class BadgeList extends React.Component<BadgeListProps> {
...
+            const badgeName: string = badge.toString();
```
Golodhros, line 74: Nit: I would remove this temporal variable and do `badgeConfig = getBadgeConfig(badge.toString());` directly

frontend/amundsen_application/static/js/components/Badges/BadgeList/index.tsx
```diff
@@ -69,6 +69,11 @@ export default class BadgeList extends React.Component<BadgeListProps> {
...
+            badgeConfig = getBadgeConfig(badgeName);
+          }
```
Golodhros, line 76: I would refactor this map callback function to separate the two things we are doing: first getting the BadgeConfig and later return the markup.

frontend/amundsen_application/static/js/components/SearchBar/index.tsx
```diff
@@ -68,14 +68,16 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
   constructor(props) {
     super(props);
     this.refToSelf = React.createRef<HTMLDivElement>();
+    const { searchTerm } = this.props;
```
Golodhros, line 71:
```suggestion
    const { searchTerm } = props;
```

frontend/amundsen_application/static/js/ducks/search/api/v0.ts
```diff
@@ -55,38 +58,29 @@ export const isResourceIndexed = (resource: ResourceType) => {
...
-  if (
-    resource === ResourceType.user &&
-    (!indexUsersEnabled() || term.length === 0)
-  ) {
-    return Promise.resolve({});
-  }
-  if (!isResourceIndexed(resource)) {
+  // If given invalid resource in list dont search for that one only for valid ones
+  const validResources = resources.filter((r) => isResourceIndexed(r));
+  if (validResources.length < 1) {
```
Golodhros, line 71: I would usually expect this instead:
```suggestion
  if (!validResources.length) {
```

frontend/amundsen_application/static/js/ducks/search/sagas.ts
```diff
@@ -222,15 +231,44 @@ export function* searchResourceWorker(
...
+        };
+        break;
+      case ResourceType.feature:
+        searchResourceResults = {
+          features: response.feature || initialState.features,
+        };
+        break;
+      default:
+        searchResourceResults = {};
+    }
```
Golodhros, line 268: I would extract this logic into a `computeSearchResourceResults(resource, response, initialState)` function call

frontend/amundsen_application/static/js/pages/DashboardPage/index.tsx
```diff
@@ -111,9 +112,10 @@ export class DashboardPage extends React.Component<
 
   componentDidUpdate() {
     const { location, getDashboard, match } = this.props;
+    const { uri: stateUri } = this.state;
```
Golodhros, line 115: nit: I would call it stateURI, as URI is an acronym

frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/CheckBoxFilter/index.tsx
```diff
@@ -60,26 +59,17 @@ export class CheckBoxFilter extends React.Component<CheckBoxFilterProps> {
...
-      checkedValues = {
-        ...checkedValues,
-        [value]: true,
-      };
+      checkedValues = [...checkedValues, value];
     } else {
-      /* Removing an object key with object destructuring */
-      const { [value]: removed, ...newCheckedValues } = checkedValues;
-      checkedValues = newCheckedValues;
+      checkedValues.splice(checkedValues.indexOf(value), 1);
```
Golodhros, line 69: We should try to avoid operations like splice that mutate an object.
The previous approach would work better, as it is creating a new object.

frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/CheckBoxFilter/index.tsx
```diff
@@ -100,27 +90,21 @@ export class CheckBoxFilter extends React.Component<CheckBoxFilterProps> {
...
-    : {};
-  if (!filterValues) {
-    filterValues = {};
-  }
-
+    : undefined;
+  const value = filterValues ? filterValues.value : '';
   return {
-    checkedValues: filterValues,
+    checkedValues: value ? value.split(',') : [],
```
Golodhros, line 98: Nit: I would create a `const checkedValues = value ? value.split(',') : []` above and make this line just `checkedValues,`

frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/InputFilter/index.tsx
```diff
@@ -61,40 +76,112 @@ export class InputFilter extends React.Component<
     } = this.props;
     const newValue = e.target.value.toLowerCase();
 
-    this.setState({ value: newValue });
+    if (newValue.includes(',')) {
+      this.setState({ value: newValue, showFilterOperationToggle: true });
+    } else {
+      this.setState({ value: newValue, showFilterOperationToggle: false });
+    }
```
Golodhros, line 83: I would do instead:

```suggestion
    const showFilterOperationToggle = newValue.includes(',');
    
    this.setState({ value: newValue, showFilterOperationToggle, });
```

frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/InputFilter/index.tsx
```diff
@@ -61,40 +76,112 @@ export class InputFilter extends React.Component<
...
+      ? { value: prevValue, filterOperation: newOperation }
+      : { filterOperation: newOperation };
 
     const newFilters = {
       ...filterState,
-      [resourceType]: { ...filterState[resourceType], [categoryId]: newValue },
+      [resourceType]: {
+        ...filterState[resourceType],
+        [categoryId]: newFilter,
+      },
```
Golodhros, line 126: This feels similar to the logic on line 85, is there an opportunity to extract a method?

frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/InputFilter/index.tsx
```diff
@@ -61,40 +76,112 @@ export class InputFilter extends React.Component<
...
+                value={FilterOperationType.AND}
+                disabled={!multiValueSelection}
+              >
+                {AND_LABEL}
+              </ToggleButton>
+              <ToggleButton value={FilterOperationType.OR}>
+                {OR_LABEL}
+              </ToggleButton>
+            </ToggleButtonGroup>
+          </div>
```
Golodhros, line 166: I would probably extract a <FilterOperationToggle /> functional component from this

frontend/docs/application_config.md
```diff
@@ -128,7 +128,7 @@ To configure Table related features we have created a new resource configuration
 #### Supported Description Sources
 
 A table resource may have a source of table and column description attached to it. We can customize it by using `supportedDescriptionSources` object which is an optional object.
-This object has `displayName` and `iconPath`, which can be used throughout the UI to represent a particular description source. See example in [config-default.ts](https://github.com/lyft/amundsenfrontendlibrary/blob/master/amundsen_application/static/js/config/config-default.ts).
+This object has `displayName` and `iconPath`, which can be used throughout the UI to represent a particular description source. See example in [config-default.ts](https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/static/js/config/config-default.ts).
```
Golodhros, line 131: Thanks for the updates!

Golodhros (commented):
Looking great!

Only some minor comments

frontend/amundsen_application/static/js/pages/SearchPage/SearchFilter/InputFilter/index.tsx
```diff
@@ -61,40 +76,112 @@ export class InputFilter extends React.Component<
...
+  handleFilterOperationChange = (newOperation) => {
+    const {
+      filterState,
+      resourceType,
+      categoryId,
+      updateFilterState,
+    } = this.props;
+
+    this.setState({ filterOperation: newOperation });
+
```
Golodhros, line 113: We should trigger an event to track this behavior, so we can get analytics information about its usage.

Golodhros (approved):
LGTM

Thanks for going the extra mile!

kristenarmes:
> Another way of doing this on an immutable way would be:
> 
> ```
>       // Remove the categoryId from the filters if the new value is empty
>       const {[categoryId]: remove, ...updatedResourceFilters} = filterState[resourceType];
> 
>       newFilters = {
>          ...filterState,
>          [resourceType]: updatedResourceFilters,
>       };
> ```
> 
> This would avoid the 'delete', that is a mutating operation (although made on a copy of the object)

@Golodhros I did it this way at first actually, but found I was getting a linting error because the `remove` is an unused variable. I didn't know if it was preferable to do it like this but then disable the `no-unused-vars` check for this spot?

Golodhros:
> > Another way of doing this on an immutable way would be:
> > ```
> >       // Remove the categoryId from the filters if the new value is empty
> >       const {[categoryId]: remove, ...updatedResourceFilters} = filterState[resourceType];
> > 
> >       newFilters = {
> >          ...filterState,
> >          [resourceType]: updatedResourceFilters,
> >       };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This would avoid the 'delete', that is a mutating operation (although made on a copy of the object)
> 
> @Golodhros I did it this way at first actually, but found I was getting a linting error because the `remove` is an unused variable. I didn't know if it was preferable to do it like this but then disable the `no-unused-vars` check for this spot? I figured the delete on the copy of the object was a middle ground to avoid linting issues but not directly mutating the original, but let me know if it's better the other way.

I see, try using '_' instead of rest, does that remove the warning?
If not, I would go ahead and disable the rule on this line!

allisonsuarez (approved):

PR #1629, opened 2021-12-20
fix: added missing description fields to mappings

allisonsuarez:
- some fields missing that are needed for FE search results display


youngyjd (approved):
btw, saw some of your work on the v2 elasticsearch with es v7. this is awesome!

PR #1637, opened 2021-12-23
feat: general document mappings for ES

allisonsuarez:
- Updated document mappings to use elasticsearch-dsl objects
- Made general mapping and added ability to extend existing classes to do custom mappings
- Added analyzers to mappings to enable stemming and an english language analyzer
- Created task that extracts records and publishes them to ES with new mappings


databuilder/databuilder/task/search/document_mappings.py
```diff
@@ -0,0 +1,84 @@
...
+    badges = Keyword(multi=True)
+    tags = Keyword(multi=True)
+    usage_metrics = RankFeatures(type=Long)
+    last_updated_timestamp = Date()
+
+
+class Table(SearchableResource):
+    display_name = Keyword(required=True, analyzer=stemming_analyzer)
+    database = Keyword(required=True)
+    cluster = Keyword()
```
mgorsk1, line 51: is there a reason cluster is not a basic `SearchableResource` field? it's common for every resource we have afaik

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,140 @@
...
+        self.extractor.init(Scoped.get_scoped_conf(conf, self.extractor.get_scope()))
+        self.transformer.init(Scoped.get_scoped_conf(conf, self.transformer.get_scope()))
+
+        conf = Scoped.get_scoped_conf(conf, self.get_scope())
+        default_analyzers = [english_analyzer, stemming_analyzer]
+
+        self.entity = conf.get_string(SearchMetadatatoElasticasearchTask.ENTITY_TYPE, default='table').lower()
+        self.document_mapping = conf.get(SearchMetadatatoElasticasearchTask.MAPPING, RESOURCE_TO_MAPPING[self.entity])
+        # not sure if this is the best way to configure custom analyzers since it is not enough to add them to the documents
+        self.analyzers = conf.get_list(SearchMetadatatoElasticasearchTask.CUSTOM_ANALYZERS, default_analyzers)
```
mgorsk1, line 58: I think the second part of the story would be to have`SearchableResource` accept custom analyzers in the constructor. Simple idea would be to pass it in the form of dict `{'name_of_analyzer': spec_of_analyzer}` and then the analyzer would be appended to every field as a subfield. For example `SearchableResource(analyzers={'polish': polish_analyzer_spec})` would create additional `name.polish`, `description.polish` ... etc. fields

And the third side of the story is if we do that then maybe some logging is required so users don't forget to adjust search query accordingly in search API.

databuilder/databuilder/task/search/document_mappings.py
allisonsuarez: features and users do not have clusters

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
allisonsuarez: So the thing is that if the analyzers are not added to the index itself rather than just the Documents it doesn't create the mapping with the analyzers, so maybe passing in a custom index would be a better strategy?

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,140 @@
...
+    def init(self, conf: ConfigTree) -> None:
+        self.extractor.init(Scoped.get_scoped_conf(conf, self.extractor.get_scope()))
+        self.transformer.init(Scoped.get_scoped_conf(conf, self.transformer.get_scope()))
+
+        conf = Scoped.get_scoped_conf(conf, self.get_scope())
+        default_analyzers = [english_analyzer, stemming_analyzer]
+
+        self.entity = conf.get_string(SearchMetadatatoElasticasearchTask.ENTITY_TYPE, default='table').lower()
+        self.document_mapping = conf.get(SearchMetadatatoElasticasearchTask.MAPPING, RESOURCE_TO_MAPPING[self.entity])
+        # not sure if this is the best way to configure custom analyzers since it is not enough to add them to the documents
```
dkunitsk: good to check that document_mapping isinstance of SearchableResource.

databuilder/databuilder/task/search/search_data_queries.py
```diff
@@ -0,0 +1,143 @@
...
+    RETURN db.name as database, cluster.name AS cluster, schema.name AS schema,
+    schema_description.description AS schema_description,
+    table.name AS name, table.key AS key, table_description.description AS description,
+    time_stamp.last_updated_timestamp AS last_updated_timestamp,
+
+    {{
+        total_usage: SUM(read.read_count),
+        unique_usage: COUNT(DISTINCT user.email)
+    }} AS usage,
+
```
dkunitsk, line 41: iiuc these changes are not bw compatible?

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,140 @@
...
+        self.analyzers = conf.get_list(SearchMetadatatoElasticasearchTask.CUSTOM_ANALYZERS, default_analyzers)
+
+        self.elasticsearch_client = conf.get(SearchMetadatatoElasticasearchTask.ELASTICSEARCH_CLIENT_CONFIG_KEY)
+        self.elasticsearch_new_index = conf.get(SearchMetadatatoElasticasearchTask.ELASTICSEARCH_NEW_INDEX_CONFIG_KEY)
+        self.elasticsearch_alias = conf.get(SearchMetadatatoElasticasearchTask.ELASTICSEARCH_ALIAS_CONFIG_KEY)
+
+        self.elasticsearch_batch_size = conf.get(SearchMetadatatoElasticasearchTask.ELASTICSEARCH_PUBLISHER_BATCH_SIZE,
```
dkunitsk, line 60: nit: good to put defaults together in a defaults dict, like other databuilder components. Easier to see at a glance what has sensible defaults and what doesn't

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,140 @@
...
+        conf = Scoped.get_scoped_conf(conf, self.get_scope())
+        default_analyzers = [english_analyzer, stemming_analyzer]
+
+        self.entity = conf.get_string(SearchMetadatatoElasticasearchTask.ENTITY_TYPE, default='table').lower()
+        self.document_mapping = conf.get(SearchMetadatatoElasticasearchTask.MAPPING, RESOURCE_TO_MAPPING[self.entity])
+        # not sure if this is the best way to configure custom analyzers since it is not enough to add them to the documents
+        self.analyzers = conf.get_list(SearchMetadatatoElasticasearchTask.CUSTOM_ANALYZERS, default_analyzers)
+
+        self.elasticsearch_client = conf.get(SearchMetadatatoElasticasearchTask.ELASTICSEARCH_CLIENT_CONFIG_KEY)
+        self.elasticsearch_new_index = conf.get(SearchMetadatatoElasticasearchTask.ELASTICSEARCH_NEW_INDEX_CONFIG_KEY)
```
dkunitsk, line 55: can this be auto-generated as something like alias + uuid? I believe that happens currently.

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,132 @@
...
+        # iterate through records
+        while record:
+            record = self.transformer.transform(record)
+            if not record:
+                # Move on if the transformer filtered the record out
+                record = self.extractor.extract()
+                continue
+            yield self.to_document(document_mapping=self.document_mapping,
+                                   metadata=record,
+                                   index=self.elasticsearch_new_index).to_dict(True)
```
dkunitsk, line 76: do we still need to_dict?

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,132 @@
...
+            # delete old index
+            self._delete_old_index(connection=connection,
+                                   document_index=index)
+
+            LOGGER.info("Elasticsearch Indexing completed")
+        finally:
+            self._closer.close()
+
+    def get_scope(self) -> str:
+        return 'task.search_metadata_to_elasticsearch'
```
dkunitsk, line 132: maybe extract this as a task_name variable for others to use. Also nit i think maybe something like "elasticsearch_dsl_publisher" may be a better name.

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,132 @@
...
+            # publish search metadata to ES
+            cnt = 0
+            for success, info in parallel_bulk(connection,
+                                               self.generator(record=record),
+                                               raise_on_error=False):
+                if not success:
+                    LOGGER.warn(f"There was an error while indexing a document to ES: {info}")
+                else:
+                    cnt += 1
+                if cnt == self.elasticsearch_batch_size:
```
dkunitsk, line 120: batch size only controls logging

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
mgorsk1: yea I think this might be redundant. If I wanted to have custom mapping with custom analyzers I would just subclass any of SearchableResource descendants. +1 to removing this.

allisonsuarez: I do but the issue is that for whatever reason I am unable to create an index with the analyzers just by using the Documents, I am looking into this issue now

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
mgorsk1: does it matter if it's dls or not? :P + this would make it confusing as this is a task with publisher in name while we have concept of publishers in amundsen

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
allisonsuarez: yes because bulk is not part of elasticsearch-dsl

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
allisonsuarez: its not just a publisher, wouldn't want tog et it confused with actual publishers

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,161 @@
...
+            LOGGER.info(f"Creating ES index {self.elasticsearch_new_index}")
+            index = Index(name=self.elasticsearch_new_index, using=self.elasticsearch_client)
+            index.document(self.document_mapping)
+            index.create()
+
+            # publish search metadata to ES
+            cnt = 0
+            for success, info in parallel_bulk(connection,
+                                               self.generate_documents(record=record),
+                                               raise_on_error=False,
```
dkunitsk, line 142: does raise_on_error also imply raise_on_exception?

dkunitsk: https://github.com/elastic/elasticsearch-py/blob/5a6bfe6dd964c62d80626da5aa9e43fef7c9a29e/elasticsearch/helpers/actions.py#L306

It seems like, if you want to hide exceptions, you should set both. Otherwise, it may be better to keep both off so you get the "real" exception instead of the wrapped exception.

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,161 @@
...
+                                               chunk_size=self.elasticsearch_batch_size,
+                                               request_timeout=self.elasticsearch_timeout):
+                if not success:
+                    LOGGER.warn(f"There was an error while indexing a document to ES: {info}")
+                else:
+                    cnt += 1
+                if cnt == self.elasticsearch_batch_size:
+                    LOGGER.info(f'Published {str(cnt)} records to ES')
```
dkunitsk, line 150: cnt*chunk_size i think

databuilder/databuilder/task/search/document_mappings.py
```diff
@@ -0,0 +1,125 @@
...
+                                     tokenizer=general_tokenizer,
+                                     filter=["lowercase"])
+
+stemming_analyzer = analysis.analyzer("stemming_analyzer",
+                                      tokenizer=general_tokenizer,
+                                      filter=["lowercase", "kstem"])
+
+english_stop = token_filter("english_stop", type="stop", stopwords="_english_")
+english_stemmer = token_filter("english_stemmer", type="stemmer", language="english")
+english_possessive_stemmer = token_filter("english_possessive_stemmer",
```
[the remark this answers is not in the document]
allisonsuarez: this uses the porter stemmer https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-stemmer-tokenfilter.html#analysis-stemmer-tokenfilter-language-parm

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,161 @@
...
+                                   metadata=record,
+                                   index=self.elasticsearch_new_index).to_dict(True)
+            record = self.extractor.extract()
+
+    def _delete_old_index(self, connection: Connections, document_index: Index) -> None:
+        alias_updates = [
+            {"add": {"index": document_index._name, "alias": self.elasticsearch_alias}}
+        ]
+        for index_name in connection.indices.get_alias():
+            if index_name.startswith(f"{self.elasticsearch_alias}_"):
```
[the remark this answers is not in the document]
allisonsuarez: that wouldn't work because we wan to remove the old indexes who's names start with the alias

databuilder/databuilder/task/search/document_mappings.py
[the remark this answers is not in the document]
allisonsuarez: can change to just the default stemmer

databuilder/databuilder/task/search/document_mappings.py
```diff
@@ -0,0 +1,125 @@
...
+                  analyzer=general_analyzer)
+    tags = Keyword(multi=True)
+    usage = RankFeatures()  # values can be used to boost document search score
+    last_updated_timestamp = Date()
+
+
+class Table(SearchableResource):
+    display_name = Text(required=True,
+                        fields={"keyword": Keyword()},
+                        analyzer=stemming_analyzer)
```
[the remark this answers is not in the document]
allisonsuarez: will change to keyword

databuilder/databuilder/task/search/search_data_queries.py
```diff
@@ -0,0 +1,143 @@
+# Copyright Contributors to the Amundsen project.
+# SPDX-License-Identifier: Apache-2.0
+
+import textwrap
+
+DEFAULT_NEO4J_TABLE_CYPHER_QUERY = textwrap.dedent(
```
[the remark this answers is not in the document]
allisonsuarez: add comment explaining that these are specific to the search task not the neo4j search extractor

databuilder/databuilder/task/search/search_data_queries.py
```diff
@@ -0,0 +1,143 @@
...
+
+    RETURN db.name as database, cluster.name AS cluster, schema.name AS schema,
+    schema_description.description AS schema_description,
+    table.name AS name, table.key AS key, table_description.description AS description,
+    time_stamp.last_updated_timestamp AS last_updated_timestamp,
+
+    {{
+        total_usage: SUM(read.read_count),
+        unique_usage: COUNT(DISTINCT user.email)
+    }} AS usage,
```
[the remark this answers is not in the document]
allisonsuarez: add a format variable for usage with default

databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class SearchMetadatatoElasticasearchTask(Task):
+
+    ENTITY_TYPE = 'doc_type'
+    ELASTICSEARCH_CLIENT_CONFIG_KEY = 'client'
+    MAPPING_CLASS = 'document_mapping'
+    ELASTICSEARCH_ALIAS_CONFIG_KEY = 'alias'
+    ELASTICSEARCH_PUBLISHER_BATCH_SIZE = 'batch_size'
+    ELASTICSEARCH_TIMEOUT = 'es_timeout'
```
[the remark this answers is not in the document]
allisonsuarez: seconds, will put it in the variable name

databuilder/databuilder/task/search/document_mappings.py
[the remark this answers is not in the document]
allisonsuarez: thinking about this further we don't filter by display name so keyword doesn't make sense here either

databuilder/databuilder/task/search/document_mappings.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class Tokenizer():
+    # separate tokens on all non-alphanumeric characters and whitespace
+    alphanum_tokenizer = tokenizer("alphanum_tokenizer",
+                                   type="pattern",
+                                   pattern="[^a-zA-Z0-9]")
+
+
+class Filter():
```
dkunitsk, line 18: nit remove empty parentheses

dkunitsk (approved):

PR #1638, opened 2021-12-25
fix: use strtobool to help csv_extractor handle bool fields

wey-gu:
When introducing sample applications on CONSUMES relation,
it's found hard to add generates_table w/o introducing
separated extractor for it to handle it as bool correctly,
for example, the is_view under table_metadata cannot parse
False correctly, too.

I consider this makes sense and introduced no surprising
corner cases but only handle str like False, NO, no, false
as bool(False), although, it's an implicit implementation.

### Summary of Changes

- Added sample data for applications with awareness of boolean field: generates_resource
- Now boolean fields in `dataloader.models` can handle string typed data in a more expected fashion

### Tests

> I mark this approach as WIP if it's agreed to do so(after all it's an implicit approach).
> If the team are not against it, I will add some tests.

### Documentation

n/a

### CheckList

Thank you all Amundsen team, I really appreciate it!

### DIFF
--- a/databuilder/databuilder/extractor/csv_extractor.py
+++ b/databuilder/databuilder/extractor/csv_extractor.py
@@ -4,6 +4,7 @@
 import csv
 import importlib
 from collections import defaultdict
+from distutils.util import strtobool
 from typing import Any, List
 
 from pyhocon import ConfigTree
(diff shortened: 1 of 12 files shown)


mgorsk1:
I think you brought up valid issue but am not 100% behind this approach for fix. Right now you are altering models towards your solution while the issue is with csv extractor only. Is there anything we could do to limit the changes to CsvExtractor only?

wey-gu:
> 

Thanks @mgorsk1 , I will try that way :)

PR #1672, opened 2022-01-18
fix: add region argument to glue extractor boto3 client instantiation

AykCanDem:
### Summary of Changes

**glue extractor** does not get `region_name` argument. It might be needed in different scenarios. For example, while using **Databricks** for extracting data from AWS Glue, it gives `NoRegionError: You must specify a region.`

You might be questioning why cant we use aws configuration and set region with **.aws/config** $HOME directory. Databricks is using instance profile for associating cluster with AWS IAM role. It is not a good practice to install aws cli to databricks cluster and configure aws credentials. So, typically, there is no **.aws/config** file stored in databricks cluster, and it is not needed.

In a nutshell, it might be better to have an option to provide `region_name` parameter while creating glue extractor.

### CheckList

### DIFF
--- a/databuilder/databuilder/extractor/glue_extractor.py
+++ b/databuilder/databuilder/extractor/glue_extractor.py
@@ -26,7 +26,8 @@ class GlueExtractor(Extractor):
         CLUSTER_KEY: 'gold',
         FILTER_KEY: None,
         MAX_RESULTS_KEY: 500,
-        RESOURCE_SHARE_TYPE: "ALL"
+        RESOURCE_SHARE_TYPE: "ALL",
+        REGION_NAME_KEY: None
     })
 
     def init(self, conf: ConfigTree) -> None:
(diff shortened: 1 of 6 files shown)


mgorsk1 (commented):
to be consistent across all keys I would do `REGION_NAME_KEY`: `region_name`. WDYT?

databuilder/databuilder/extractor/glue_extractor.py
```diff
@@ -21,11 +21,13 @@ class GlueExtractor(Extractor):
...
     MAX_RESULTS_KEY = 'max_results'
     RESOURCE_SHARE_TYPE = 'resource_share_type'
+    REGION_NAME = "region"
     DEFAULT_CONFIG = ConfigFactory.from_dict({
         CLUSTER_KEY: 'gold',
         FILTER_KEY: None,
         MAX_RESULTS_KEY: 500,
-        RESOURCE_SHARE_TYPE: "ALL"
+        RESOURCE_SHARE_TYPE: "ALL",
+        REGION_NAME: None
```
mgorsk1, line 30: If there is no default we don't really need this in the config

AykCanDem: I agree about `REGION_NAME_KEY`. However, if we wont put it as default, tests gives errors `pyhocon.exceptions.ConfigMissingException: 'No configuration setting found for key region' `.
I think it should be there as `FILTER_KEY`.

AykCanDem:
any update on this?

mgorsk1:
closed in favor of https://github.com/amundsen-io/amundsen/pull/1687

PR #1684, opened 2022-01-30
feat: Add ElasticsearchWatermarkExtractor

jaredbwasserman:
### Summary of Changes

Add `ElasticsearchWatermarkExtractor`.

### Tests

I added a few unit tests for the newly added `ElasticsearchWatermarkExtractor`.

### Documentation

I added a new documentation section for `ElasticsearchWatermarkExtractor` just under `ElasticsearchColumnStatsExtractor` ( https://github.com/amundsen-io/amundsen/tree/main/databuilder#elasticsearchcolumnstatsextractor ).

### CheckList

### DIFF
--- a/databuilder/databuilder/extractor/es_base_extractor.py
+++ b/databuilder/databuilder/extractor/es_base_extractor.py
@@ -17,6 +17,8 @@ class ElasticsearchBaseExtractor(Extractor):
 
     By default, the extractor does not add sort_order to columns. Set ELASTICSEARCH_CORRECT_SORT_ORDER conf to True
     for columns to have correct sort order.
+
+    Set ELASTICSEARCH_TIME_FIELD to the name of the field representing time.
     """
 
     ELASTICSEARCH_CLIENT_CONFIG_KEY = 'client'
@@ -25,6 +27,9 @@ class ElasticsearchBaseExtractor(Extractor):
     # For backwards compatibility, the Elasticsearch extractor does not add sort_order to columns by default.
     # Set this to true in the conf for columns to have correct sort order.
     ELASTICSEARCH_CORRECT_SORT_ORDER = 'correct_sort_order'
+
+    # Set this to the name of the field representing time.
+    ELASTICSEARCH_TIME_FIELD = 'time_field'
 
     CLUSTER = 'cluster'
     SCHEMA = 'schema'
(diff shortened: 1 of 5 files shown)


databuilder/databuilder/extractor/es_base_extractor.py
```diff
@@ -129,6 +148,15 @@ def _correct_sort_order(self) -> bool:
         except Exception:
             return False
 
+    # Default time field is @timestamp to match ECS
+    # See https://www.elastic.co/guide/en/ecs/master/ecs-base.html
+    @property
+    def _time_field(self) -> str:
+        try:
+            return self.conf.get(ElasticsearchBaseExtractor.ELASTICSEARCH_TIME_FIELD)
```
mgorsk1, line 156: why not just make @timestamp a default for this config key?

databuilder/databuilder/extractor/es_base_extractor.py
```diff
@@ -55,6 +60,20 @@ def _get_indexes(self) -> Dict:
...
+    def _get_index_creation_date(self, index_metadata: Dict) -> Optional[float]:
+        if index_metadata is None:
+            return None
+        settings = index_metadata.get('settings', dict())
+        if len(settings) == 0:
+            return None
+        index = settings.get('index', dict())
+        if len(index) == 0:
+            return None
+        creation_date = index.get('creation_date')
```
mgorsk1, line 72: so if creation_date is not none we shall return it. in any other case we return None. could this section be written in a less complex approach?

mgorsk1 (changes requested):

databuilder/databuilder/extractor/es_base_extractor.py
jaredbwasserman: Yes, that makes sense to me. Does it make sense to remove the try/catch and just have `@timestamp` as the second parameter?

databuilder/databuilder/extractor/es_base_extractor.py
jaredbwasserman: Yes, I'll rewrite it. I wrote it this way to get around some mypy errors, but I agree this could be written more simply.

mgorsk1 (approved):

PR #1693, opened 2022-02-01
feat: Add column statistics indicator for columns with TableColumnStats available

mgorsk1:
### Summary of Changes

Currently there is no indication if there is a need to expand the column details to see whether column has column stats available or not. Users need to expand details even if there is nothing there.

This PR introduces an icon which indicates that a given column has statistics available. I am not strongly opinionated towards icon or placement choice but I think such indicator is a must to improve user experience.

This is how it looks like after this initial implementation:
image

### Tests

tbd

### Documentation

tbd

### CheckList
- [] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.

### DIFF
--- a/frontend/amundsen_application/static/js/features/ColumnList/index.tsx
+++ b/frontend/amundsen_application/static/js/features/ColumnList/index.tsx
@@ -85,7 +85,6 @@ type ContentType = {
   title: string;
   description: string;
   nestedLevel: number;
-  stats: boolean;
 };
 
 type DatatypeType = {
@@ -253,7 +252,6 @@ const ColumnList: React.FC<ColumnListProps> = ({
         title: item.name,
         description: item.description,
         nestedLevel: item.nested_level || 0,
-        stats: hasItemStats,
       },
       type: {
         type: item.col_type,
@@ -328,7 +326,7 @@ const ColumnList: React.FC<ColumnListProps> = ({
     {
       title: 'Name',
       field: 'content',
-      component: ({ title, description, nestedLevel, stats }: ContentType) => (
+      component: ({ title, description, nestedLevel }: ContentType) => (
         <>
           {nestedLevel > 0 && (
             <>
--- a/frontend/amundsen_application/static/js/features/ColumnList/index.tsx
+++ b/frontend/amundsen_application/static/js/features/ColumnList/index.tsx
@@ -43,6 +43,8 @@ import { logAction } from 'utils/analytics';
 import { buildTableKey, TablePageParams } from 'utils/navigationUtils';
 import { getUniqueValues, filterOutUniqueValues } from 'utils/stats';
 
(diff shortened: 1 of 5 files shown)


mgorsk1:
@Golodhros @danwom I'd appreciate any feedback you might have on this 🙏

mgorsk1:
ping @Golodhros

Golodhros:
Pinging Knowl!

knowlb:
We love the idea of indicating whether there are column stats to expand, but the faint icon on the far right was really hard to notice, so would like to suggest using this treatment (see attached img):
column stats indicator
This is also to be consistent with a new feature we're developing to highlight search keyword matches in results, including pulling out column names for tables and charts & queries for dashboards: 
image

mgorsk1:
thanks @knowlb that totally makes sense! where do you source your icons from (or are they custom made) so I can submit PR with matching aesthetics?

mgorsk1:
I refactored to your suggestion @knowlb with minor adjustment - I didn't like the fact that icon presence could shift some column names and in results they'd not be aligned throughout so I've added the icon to the left of col name & description. Let me know wdyt

knowlb:
Thanks and like the alignment idea too! Looks great-

frontend/amundsen_application/static/js/features/ColumnList/index.tsx
```diff
@@ -309,6 +311,18 @@ const ColumnList: React.FC<ColumnListProps> = ({
   });
 
   let formattedColumns: ReusableTableColumn[] = [
+    {
+      title: '',
+      field: 'stats',
+      width: 24,
```
Golodhros, line 317: Let's make this magic number into a constant.
Maybe `const STATS_COLUMN_WIDTH = 24;`

frontend/amundsen_application/static/js/features/ColumnList/index.tsx
```diff
@@ -43,6 +43,8 @@ import { logAction } from 'utils/analytics';
 import { buildTableKey, TablePageParams } from 'utils/navigationUtils';
 import { getUniqueValues, filterOutUniqueValues } from 'utils/stats';
 
+import { GraphIcon } from 'components/SVGIcons/GraphIcon';
```
Golodhros, line 46: Could you add a couple of extra test cases in 'js/features/ColumnList/index.spec.tsx', in the 'when columns with one usage data entry are passed' and 'when columns with no usage data are passed', looking for this icon to show up?

Golodhros (changes requested):

frontend/amundsen_application/static/js/features/ColumnList/index.tsx
mgorsk1: done

Golodhros (approved):
LGTM

Thanks a lot!!

PR #1695, opened 2022-02-02
feat: Backend support for nested columns

kristenarmes:

### Summary of Changes

This PR adds backend support for nested columns by introducing a new TypeMetadata base class, along with Map/Array/Struct/Scalar subclasses (with opportunity to add more as needed). Columns will have an empty `TypeMetadata` by default and Amundsen will not require it for the time being. To eliminate breaking changes, setting TypeMetadata is done via the `ComplexTypeTransformer`. With time this logic can easily be integrated into extractors themselves.

The TypeMetadata classes are intended to be general and reusable. Syntax-specific parsers will need to be added per-database for converting type strings into structured objects. In this PR we have implemented a Hive parse that uses `pyparsing`.

After this we'll work on the APIs and then the frontend.

### Tests

Added test cases in `test_table_metadata.py`, `test_type_metadata.py`, `test_complex_type_transformer.py`, `test_hive_complex_type_parser.py`

### Documentation

Added details about the `ComplexTypeTransformer` in `databuilder/README.md`

### CheckList


databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,306 @@
...
+            key=self._get_node_key(start_key),
+            label=TypeMetadata.TYPE_NODE_LABEL,
+            attributes={
+                TypeMetadata.TYPE_KIND: 'array',
+                TypeMetadata.TYPE_DATA_TYPE: self.data_type.__str__()
+            }
+        )
+        return node
+
+    def _get_relation(self,
```
dkunitsk, line 92: i generally like to inline these into `create_node_iterator`. The separate function isn't necessary.

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,306 @@
...
+        if not self.is_terminal_type():
+            yield self._get_relation(self.start_label, self.start_key)
+            yield from self.value.create_relation_iterator()
+
+
+class StructItem():
+    def __init__(self,
+                 name: str,
+                 description: Union[str, None],
+                 data_type: Union[TypeMetadata, str]) -> None:
```
dkunitsk, line 193: in general i feel like Optional[TypeMetadata] is the right thing.

dkunitsk: but it's debatable

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,306 @@
...
+            raise Exception('Required start node label cannot be None')
+        if not self.start_key:
+            raise Exception('Required start node key cannot be None')
+
+        if not self.is_terminal_type():
+            yield self._get_relation(self.start_label, self.start_key)
+            yield from self.value.create_relation_iterator()
+
+
+class StructItem():
```
dkunitsk, line 189: having the name ON TypeMetadata would eliminate the need for this. let's discuss.

dkunitsk (commented):
No major problems, looks great. A few tweaks I'd suggest, let's discuss IRL.

databuilder/databuilder/models/type_metadata.py
kristenarmes: fyi I got rid of StructItem and added the description to the base class as we discussed, but I didn't add the name or type there due to the different way they're represented in a couple places. I had started to add them but it felt like it didn't fully translate to all the subclasses, mostly in the struct class since it contains a dict of names to types rather than just one type. PTAL and we can discuss if you see a better way of making it more generic

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,349 @@
...
+from typing import Any, Dict, Iterator, Optional
+
+from databuilder.models.graph_node import GraphNode
+from databuilder.models.graph_relationship import GraphRelationship
+from databuilder.models.graph_serializable import GraphSerializable
+from databuilder.models.table_metadata import DescriptionMetadata
+
+
+class TypeMetadata(abc.ABC, GraphSerializable):
+    TYPE_NODE_LABEL = 'Column_Subtype'
```
dkunitsk, line 14: nit: to refer to all of these you'll be doing TypeMetadata.FOO anyway, so remove `TYPE_` as a prefix

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,349 @@
...
+    TYPE_RELATION_TYPE = 'COLUMN_SUBTYPE'
```
dkunitsk, line 15: maybe just SUBTYPE. Since TypeMetadatas can have other child TypeMetadatas, so COLUMN_SUBTYPE will not be true after level 1.

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,349 @@
...
+    TYPE_DATA_TYPE = 'data_type'
+    TYPE_MAP_KEY = 'map_key'
+    TYPE_MAP_VALUE = 'map_value'
+    TYPE_SORT_ORDER = 'sort_order'
+
+    @abc.abstractmethod
+    def __init__(self,
+                 description: Optional[str] = None,
+                 start_label: Optional[str] = None,
+                 start_key: Optional[str] = None) -> None:
```
dkunitsk, line 29: so, not parent?

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,349 @@
...
+                                                start_label,
+                                                start_key)
+        self.data_type = data_type
+
+        self.name = '__array_inner'
+        self.kind = 'array'
+
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, ArrayTypeMetadata):
+            return (self.data_type.__eq__(other.data_type) and
```
dkunitsk, line 98: i think `self.data_type == other.data_type` will actually call self.data_type.__eq__ already

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,349 @@
...
+                    return False
+            return (self.start_label == other.start_label and
+                    self.start_key == other.start_key)
+        return False
+
+    def __str__(self) -> str:
+        inner_string = ''
+        for name, data_type in self.struct_items.items():
+            inner_string += f"{name}:{data_type.__str__()},"
+        return f"{self.kind}<{inner_string[:-1]}>"
```
dkunitsk, line 297: this should be somehow configurable

databuilder/databuilder/models/type_metadata.py
kristenarmes: re: the parent in the base class, after further thought and trying things out I'm not sure that it is worth it to have for the same reason I mentioned in a previous comment about why I didn't put the name in the base class

1. children of structs wouldn't know which `name` to use from their parent bc they would all have the same parent obj with a dict including all of them, so the benefit of being able to get the key in the base class wouldn't be available
2. if we are setting the child `data_type.parent` with the parent obj after the recursive bottom up children are created, we'd still need to be passing the start_label/start_key through the inits so it doesn't remove the need for that

I've added the parser with the recursive logic in my latest changes, check out the new stuff and we can discuss further

databuilder/databuilder/models/description_metadata.py
```diff
@@ -0,0 +1,150 @@
+# Copyright Contributors to the Amundsen project.
```
kristenarmes, line 1: Separated `DescriptionMetadata` from `TableMetadata` due to circular import issues with `TypeMetadata`

databuilder/databuilder/models/table_metadata.py
```diff
@@ -27,19 +27,18 @@
...
 from databuilder.models.table_serializable import TableSerializable
+from databuilder.models.type_metadata import TypeMetadata
 from databuilder.serializers.atlas_serializer import (
     add_entity_relationship, get_entity_attrs, get_entity_relationships,
 )
 from databuilder.utils.atlas import AtlasRelationshipTypes, AtlasSerializedEntityOperation
 
-DESCRIPTION_NODE_LABEL_VAL = 'Description'
-DESCRIPTION_NODE_LABEL = DESCRIPTION_NODE_LABEL_VAL
-
```
dkunitsk, line 42: i think keep for bw compatibility

databuilder/databuilder/models/description_metadata.py
```diff
@@ -0,0 +1,150 @@
...
+    INVERSE_DESCRIPTION_RELATION_TYPE = 'DESCRIPTION_OF'
+
+    # The default editable source.
+    DEFAULT_SOURCE = "description"
+
+    def __init__(self,
+                 text: Optional[str],
+                 source: str = DEFAULT_SOURCE,
+                 description_key: Optional[str] = None,
+                 start_label: Optional[str] = None,  # Table, Column, Schema
```
dkunitsk, line 37: nit: or `TypeMetadata`, now.

databuilder/databuilder/models/description_metadata.py
```diff
@@ -0,0 +1,150 @@
...
+from databuilder.models.atlas_serializable import AtlasSerializable
+from databuilder.models.graph_node import GraphNode
+from databuilder.models.graph_relationship import GraphRelationship
+from databuilder.models.graph_serializable import GraphSerializable
+
+DESCRIPTION_NODE_LABEL_VAL = 'Description'
+DESCRIPTION_NODE_LABEL = DESCRIPTION_NODE_LABEL_VAL
+
+
+# TODO: this should inherit from ProgrammaticDescription in amundsen-common
```
dkunitsk, line 19: i think this is wrong/outdated at this point, can remove

databuilder/databuilder/transformer/complex_type_transformer.py
```diff
@@ -0,0 +1,42 @@
...
+class ComplexTypeTransformer(Transformer):
+    """
+    Transforms complex types for columns in a table.
+    Takes a TableMetadata object and iterates over the columns.
+    The configured parser takes each column's type and key and sets the
+    column's type_metadata field with the parsed results contained in a
+    TypeMetadata object.
+    """
+    def init(self, conf: ConfigTree) -> None:
+        parsing_function = conf.get_string(PARSING_FUNCTION)
```
dkunitsk, line 24: a check here to make sure `parsing_function` is the right type of function would be good

databuilder/databuilder/models/type_metadata.py
dkunitsk: base class:
```
def key(self):
  # key is always parent's key + own name
  parent_key = None
  if isinstance(self.parent, ColumnMetadata):
    parent_key = self.parent.get_column_key() # this doesn't exist atm
  else:
    parent_key = self.parent.key()
  return parent_key + '.' + self.name
```

dkunitsk: - let columns know their own key (maybe pass the table's key to each column metadata, for example)
- if an array has a child, that child's `name` will be a fake name, like `_inner_`. 
- `key()` can be implemented as above. Each node doesn't need to know it's own key, it only needs to know its own name and its parent and it can always get the key via logic above.

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+            return f"{self.parent_key()}/type/{self.name}"
+        else:
+            return f"{self.parent_key()}/{self.name}"
+
+    def description_key(self) -> str:
+        if self.description:
+            description_id = self.description.get_description_id()
+            return f"{self.key()}/{description_id}"
+        else:
+            return ''
```
dkunitsk, line 84: None is better

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+
+    def parent_key(self) -> str:
+        if isinstance(self.parent, ColumnMetadata):
+            column_key = self.parent.get_column_key()
+            return column_key if column_key else ''
```
dkunitsk, line 89: better to crash than return ''

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+    def __repr__(self) -> str:
+        return f"TypeMetadata({self.type_str!r})"
+
+
+class ArrayTypeMetadata(TypeMetadata):
+    kind = 'array'
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super(ArrayTypeMetadata, self).__init__(*args, **kwargs)
+        self.data_type: Optional[TypeMetadata] = None
```
dkunitsk, line 108: nit: for clarity, how about `array_inner_type`. 

btw: To make mypy happy, I recommend throwing in assertions later so you don't have to deal with None cases.

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+        return False
+
+    def is_terminal_type(self) -> bool:
+        return isinstance(self.data_type, ScalarTypeMetadata)
+
+    def create_node_iterator(self) -> Iterator[GraphNode]:
+        if not self.parent_key():
+            raise Exception('Required parent node key cannot be None')
+        if not self.data_type:
+            raise Exception('Must set inner data type')
```
dkunitsk, line 127:
```suggestion
        assert self.array_inner_type, f"Array inner type must be set for {self.name}"
```

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+
+        node_attributes: Dict[str, Union[str, None, int]] = {
+            TypeMetadata.KIND: self.kind,
+            TypeMetadata.NAME: self.name,
+            TypeMetadata.DATA_TYPE: self.data_type.type_str
```
dkunitsk, line 132: shouldn't this be the entire thing, not just the inner thing?

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+
+        if self.description:
+            yield self.description.get_relation(
+                TypeMetadata.NODE_LABEL,
+                self.key(),
+                self.description_key()
+            )
+
+        if not self.is_terminal_type():
+            yield from self.data_type.create_relation_iterator()
```
dkunitsk, line 174: nice

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class MapTypeMetadata(TypeMetadata):
+    kind = 'map'
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super(MapTypeMetadata, self).__init__(*args, **kwargs)
+        self.map_key: Optional[TypeMetadata] = None
```
dkunitsk, line 182: map_key_type and map_value_type maybe better

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+        if isinstance(self.parent, ColumnMetadata):
+            column_key = self.parent.get_column_key()
+            return column_key if column_key else ''
+        else:
+            return self.parent.key()
+
+    def parent_label(self) -> str:
+        if isinstance(self.parent, ColumnMetadata):
+            return ColumnMetadata.COLUMN_NODE_LABEL
+        else:
```
dkunitsk, line 96: niiit redundant else

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+            description_id = self.description.get_description_id()
+            return f"{self.key()}/{description_id}"
+        else:
+            return ''
+
+    def parent_key(self) -> str:
+        if isinstance(self.parent, ColumnMetadata):
+            column_key = self.parent.get_column_key()
+            return column_key if column_key else ''
+        else:
```
dkunitsk, line 90: niiit redundant else

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+        if isinstance(self.parent, ColumnMetadata):
+            return f"{self.parent_key()}/type/{self.name}"
+        else:
+            return f"{self.parent_key()}/{self.name}"
+
+    def description_key(self) -> str:
+        if self.description:
+            description_id = self.description.get_description_id()
+            return f"{self.key()}/{description_id}"
+        else:
```
dkunitsk, line 83: niiit redundant else

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+
+        self._node_iter = self.create_node_iterator()
+        self._relation_iter = self.create_relation_iterator()
+
+    @abc.abstractmethod
+    def __eq__(self, other: Any) -> bool:
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def is_terminal_type(self) -> bool:
```
dkunitsk, line 50: add comment for what this means. the definition we came up with is not obvious

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+            return (self.name == other.name and
+                    self.type_str == other.type_str and
+                    self.description == other.description and
+                    self.sort_order == other.sort_order and
+                    self.data_type == other.data_type and
+                    self.key() == other.key())
+        return False
+
+    def is_terminal_type(self) -> bool:
+        return isinstance(self.data_type, ScalarTypeMetadata)
```
dkunitsk, line 121: put a comment here, maybe

dkunitsk: oh I also realized a thing. Maybe we should let people specify is_terminal_type from the constructor.

databuilder/databuilder/transformer/complex_type_transformer.py
```diff
@@ -0,0 +1,43 @@
...
+        except (ValueError, ModuleNotFoundError, AttributeError):
+            raise Exception("Invalid parsing function provided to ComplexTypeTransformer")
+
+    def transform(self, record: Any) -> TableMetadata:
+        if not isinstance(record, TableMetadata):
+            raise Exception("ComplexTypeTransformer expects record of type TableMetadata")
+
+        for column in record.columns:
+            column.set_column_key(record._get_col_key(column))
+            column.set_type_metadata(self._parsing_function(column.type, column.name, column))
```
dkunitsk, line 38: very clean

databuilder/README.md
```diff
@@ -1678,6 +1678,9 @@ Adds the same set of tags to all tables produced by the job.
 #### [GenericTransformer](./databuilder/transformer/generic_transformer.py)
 Transforms dictionary based on callback function that user provides.
 
+#### [ComplexTypeTransformer](./databuilder/transformer/complex_type_transformer.py)
```
dkunitsk, line 1681: let's remove this for now and then document everything once there's end to end support

databuilder/databuilder/models/type_metadata.py
```diff
@@ -0,0 +1,386 @@
...
+            assert description_key is not None, f"Could not retrieve description key for {self.name}"
+            yield self.description.get_node(description_key)
+
+        if not self.is_terminal_type():
+            yield from self.map_key_type.create_node_iterator()
+            yield from self.map_value_type.create_node_iterator()
+
+    def create_relation_iterator(self) -> Iterator[GraphRelationship]:
+        assert self.map_key_type is not None, f"Map key type must be set for {self.name}"
+        assert self.map_value_type is not None, f"Map value type must be set for {self.name}"
```
dkunitsk, line 230: this seems wrong (likewise for `create_node_iterator`). If the node is terminal and the key/values are not set, the node still needs to be GraphSerializable. I.e. `create_node_iterator` would just create it's own node, and `create_relation_iterator` would just create the relationship between parent and this node, and to description node.

dkunitsk (approved):
very very nice. Love the test coverage also.

PR #1719, opened 2022-02-18
fix: on Neo4j Query to grab Personal popular resources

Danucas:
### CheckList


mgorsk1:
Could you explain why this change is required in your opinion @Danucas ? My understanding of this query prior to your change was:
1. we grab a tables that user X follows
2. we check who else is following those tables - we get list of A, B, C users
3. we return tables followed by A, B, C users as personalized popular resources

This way if you are following table A, your peer is following tables A and B - you will see table B in your results even though you haven't used it yet - cause it's popular amongst users who have similar but not the same patterns of behavior.

Is there anything wrong with that logic?

PR #1736, opened 2022-02-28
chore: add upper boundary for amundsen-common pin

mgorsk1:
### Summary of Changes

Currently pinning with only lower boundary makes our packages vulnerable to breaking changes in Amundsen common. This PR adds upper boundary so our packages are relying on common within range of neighbouring versions so we minimize risk of creating incompatible combinations.

Previous approach resulted in a fact that a significantly older version of any service (databuilder/frontend/search/metadata) was relying on the newest possible version of common (established on pip install) which could (and actually did on one occasion) break things.

### Tests

n/a

### Documentation

n/a

### CheckList

### DIFF
--- a/requirements-common.txt
+++ b/requirements-common.txt
@@ -5,7 +5,7 @@
 
 # It is recommended to always pin the exact version (not range) - otherwise common upgrade won't trigger unit tests
 # on all repositories reyling on this file and any issues that arise from common upgrade might be missed.
-amundsen-common>=0.25.0,<0.26.0
+amundsen-common>=0.26.0,<0.27.0
 attrs>=19.1.0
 boto3==1.17.23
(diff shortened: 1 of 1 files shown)


mgorsk1:
cc @dkunitsk as we had different opinions on this so I'd appreciate giving a second look. thanks 🙏

feng-tao (approved):

dkunitsk (approved):
I'm a +1 as well. It's probably better if we didn't need it and enforced strict bw-compatibility checking for `common`, but that hasn't always been the case and there's valid reasons for breaking compatibility (occasionally) so the pin is the next best option.

mgorsk1:
thanks @dkunitsk @feng-tao! 🙇

PR #1754, opened 2022-03-08
feat: configurable default filter values

allisonsuarez:

### Summary of Changes

Added the ability to configure a default value for filters. If this value is set the filter is on by default and applies to all searches unless reset by the user.

### CheckList

### DIFF
--- a/frontend/amundsen_application/static/js/ducks/search/filters/reducer.ts
+++ b/frontend/amundsen_application/static/js/ducks/search/filters/reducer.ts
@@ -55,7 +55,7 @@ export function getDefaultFiltersForResource(
   const defaultFilters =
     filterCategories
       ?.filter(
-        filterCategory => filterCategory.defaultValue && filterCategory.defaultValue !== []
+        (filterCategory) => filterCategory.defaultValue && filterCategory.defaultValue !== []
       )
       .reduce((acc, currentFilter) => {
         return {
--- a/frontend/amundsen_application/static/js/ducks/search/filters/reducer.ts
+++ b/frontend/amundsen_application/static/js/ducks/search/filters/reducer.ts
@@ -55,8 +55,7 @@ export function getDefaultFiltersForResource(
   const defaultFilters =
     filterCategories
       ?.filter(
-        ({defaultValue}) =>
-          defaultValue && defaultValue.length > 0
+        ({defaultValue}) => defaultValue && defaultValue.length > 0
       )
       .reduce((acc, currentFilter) => {
         return {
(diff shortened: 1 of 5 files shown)


mgorsk1:
awesome idea which also crossed my mind recently ❤️

frontend/amundsen_application/static/js/ducks/search/filters/reducer.ts
```diff
@@ -46,10 +47,33 @@ export interface ResourceFilterReducerState {
   };
 }
 
+export function getDefaultFiltersForResource(
+  resourceType: ResourceType
+): ResourceFilterReducerState {
+  const defaultFilters = {};
+  const resourceConfig = appConfig.resourceConfig[resourceType];
```
Golodhros, line 54: We should use 'getFilterConfigByResource' instead

frontend/amundsen_application/static/js/ducks/search/filters/reducer.ts
```diff
@@ -46,10 +47,33 @@ export interface ResourceFilterReducerState {
...
+  resourceConfig.filterCategories.forEach((filter) => {
+    const filterDefaultValue: string[] = filter.defaultValue;
+    if (filterDefaultValue && filterDefaultValue !== []) {
+      defaultFilters[filter.categoryId] = {
+        value: filterDefaultValue.join(),
+        filterOperation: filter.allowableOperation,
+      };
+    }
+  });
```
Golodhros, line 63: Let's try to do this logic with a reduce and a filter instead of this 'forEach'.
It would be something like:
```
defaultFilters = filterCategories.filter(hasDefaultValue).reduce((acc, filter) => {
  return {
       ...acc, 
       [filter.categoryId]: {
           value: filterDefaultValue.join(),
            filterOperation: filter.allowableOperation
        }
}, {});
```

Golodhros (changes requested):

frontend/amundsen_application/static/js/ducks/search/filters/reducer.ts
```diff
@@ -46,10 +47,39 @@ export interface ResourceFilterReducerState {
...
+export function getDefaultFiltersForResource(
+  resourceType: ResourceType
+): ResourceFilterReducerState {
+  const filterCategories = getFilterConfigByResource(resourceType);
+  const initialValue = {};
+  const defaultFilters =
+    filterCategories
+      ?.filter(
+        (filterCategory) =>
+          filterCategory.defaultValue && filterCategory.defaultValue !== []
```
Golodhros, line 59:
```suggestion
        ({defaultValue) =>
          defaultValue && defaultValue.length
```

We don't usually compare to empty arrays in JavaScript, because:
```
[] === []
false
``` 
:)

Golodhros (changes requested):

Golodhros (approved):

PR #1766, opened 2022-03-15
refactor: Change error handling in ComplexTypeTransformer

kristenarmes:

### Summary of Changes

Prior to this change, if there was anything that went wrong in the `ComplexTypeTransformer` then an exception would be thrown. Since `TypeMetadata` is an optional field for columns, if anything fails in this transformer it makes more sense to log the issues rather than throw exceptions so that any process using the transformer can continue.

### Tests

Updated related tests in `test_complex_type_transformer.py` and added a test for if the parsing function fails.

### Documentation

N/A

### CheckList

### DIFF
--- a/databuilder/databuilder/transformer/complex_type_transformer.py
+++ b/databuilder/databuilder/transformer/complex_type_transformer.py
@@ -33,12 +33,12 @@ class ComplexTypeTransformer(Transformer):
             module_name, function_name = parsing_function.rsplit(".", 1)
             mod = importlib.import_module(module_name)
             self._parsing_function = getattr(mod, function_name)
-        except Exception as e:
-            LOGGER.error(f"Invalid parsing function provided to ComplexTypeTransformer: {e}")
+        except (ValueError, ModuleNotFoundError, AttributeError):
+            raise Exception("Invalid parsing function provided to ComplexTypeTransformer")
 
     def transform(self, record: Any) -> TableMetadata:
         if not isinstance(record, TableMetadata):
-            raise Exception("ComplexTypeTransformer expects record of type TableMetadata")
+            raise Exception(f"ComplexTypeTransformer expects record of type TableMetadata, received {type(record)}")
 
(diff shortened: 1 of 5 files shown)


dkunitsk (approved):

databuilder/databuilder/transformer/complex_type_transformer.py
```diff
@@ -21,21 +25,32 @@ class ComplexTypeTransformer(Transformer):
...
+
         try:
             parsing_function = conf.get_string(PARSING_FUNCTION)
             module_name, function_name = parsing_function.rsplit(".", 1)
             mod = importlib.import_module(module_name)
             self._parsing_function = getattr(mod, function_name)
-        except (ValueError, ModuleNotFoundError, AttributeError):
-            raise Exception("Invalid parsing function provided to ComplexTypeTransformer")
+        except Exception as e:
+            LOGGER.error(f"Invalid parsing function provided to ComplexTypeTransformer: {e}")
```
dkunitsk, line 37: i think better to not catch here

databuilder/databuilder/transformer/complex_type_transformer.py
```diff
@@ -33,12 +33,12 @@ def init(self, conf: ConfigTree) -> None:
             module_name, function_name = parsing_function.rsplit(".", 1)
             mod = importlib.import_module(module_name)
             self._parsing_function = getattr(mod, function_name)
-        except Exception as e:
-            LOGGER.error(f"Invalid parsing function provided to ComplexTypeTransformer: {e}")
+        except (ValueError, ModuleNotFoundError, AttributeError):
+            raise Exception("Invalid parsing function provided to ComplexTypeTransformer")
```
dkunitsk, line 37: even better is to do nothing. this erases information about the actual exception that occurred and instead raises a generic exception with a generic message.

dkunitsk (approved):

PR #1772, opened 2022-03-23
fix: open file in read mode and close on cleanup for ElasticsearchPublisher

m1racoli:
* fix: open file in read mode
* fix: close file handler on cleanup

### Summary of Changes

Since we only read from the file, we do not need to open it in write mode. Furthermore we close the file on cleanup in order to avoid issues with open file handlers. In particular this affects the automatic cleanup of [temporary directories](https://docs.python.org/3.10/library/tempfile.html#tempfile.TemporaryDirectory) containing the file.

### Tests

This affects low level aspect of how we open and close files, which might be beyond the scope of tests.

### Documentation

Nothing to document as there is not effective change in behaviour.

### CheckList

### DIFF
--- a/databuilder/databuilder/publisher/elasticsearch_publisher.py
+++ b/databuilder/databuilder/publisher/elasticsearch_publisher.py
@@ -44,7 +44,7 @@ class ElasticsearchPublisher(Publisher):
         self.conf = conf
 
         self.file_path = self.conf.get_string(ElasticsearchPublisher.FILE_PATH_CONFIG_KEY)
-        self.file_mode = self.conf.get_string(ElasticsearchPublisher.FILE_MODE_CONFIG_KEY, 'w')
+        self.file_mode = self.conf.get_string(ElasticsearchPublisher.FILE_MODE_CONFIG_KEY, 'r')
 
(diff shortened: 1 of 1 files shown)


m1racoli:
This still needs review.

databuilder/databuilder/publisher/elasticsearch_publisher.py
```diff
@@ -124,5 +124,13 @@ def publish_impl(self) -> None:
         # perform alias update and index delete in single atomic operation
         self.elasticsearch_client.indices.update_aliases(update_action)
 
+    def close(self) -> None:
```
mgorsk1, line 127: why do we need this function? is it used anywhere?

m1racoli: It's part of the parent class [Scoped](https://github.com/amundsen-io/amundsen/blob/3cc8bd587dd296d99ed94a5ee08a420be359249d/databuilder/databuilder/__init__.py#L56) and automatically executed at the end. We do the same already for [FSElasticsearchJSONLoader](https://github.com/amundsen-io/amundsen/blob/3cc8bd587dd296d99ed94a5ee08a420be359249d/databuilder/databuilder/loader/file_system_elasticsearch_json_loader.py#L59).

m1racoli:
This still needs review.

PR #1789, opened 2022-03-30
feat: add boolean config for adding publisher metadata to neo4j node

wli-brex:
### Summary of Changes

Adding a boolean config to control whether publisher metadata (e.g. ) need to be added as properties. Through this config (default True), we can [retain the data](https://github.com/amundsen-io/amundsen/blob/c9727bffaa6aeb53623ad2a808013e35c549f18a/databuilder/databuilder/task/neo4j_staleness_removal_task.py#L173) (won't get purged in async mode) when such metadata is not included.

### Tests

n/a

### Documentation

n/a

### CheckList

### DIFF
--- a/databuilder/databuilder/publisher/neo4j_csv_publisher.py
+++ b/databuilder/databuilder/publisher/neo4j_csv_publisher.py
@@ -63,6 +63,10 @@ PUBLISHED_TAG_PROPERTY_NAME = 'published_tag'
 # Neo4j property name for last updated timestamp
 LAST_UPDATED_EPOCH_MS = 'publisher_last_updated_epoch_ms'
 
+# A boolean flag to indicate if publisher_metadata (e.g. published_tag, publisher_last_updated_epoch_ms)
+# will be add as properties of Neo4j nodes
+ADD_PUBLISHER_METADATA = 'add_publisher_metadata'
+
 RELATION_PREPROCESSOR = 'relation_preprocessor'
 
 # CSV HEADER
@@ -99,6 +103,7 @@ DEFAULT_CONFIG = ConfigFactory.from_dict({NEO4J_TRANSACTION_SIZE: 500,
                                           NEO4J_MAX_CONN_LIFE_TIME_SEC: 50,
                                           NEO4J_ENCRYPTED: True,
                                           NEO4J_VALIDATE_SSL: False,
+                                          ADD_PUBLISHER_METADATA: True,
                                           RELATION_PREPROCESSOR: NoopRelationPreprocessor()})
 
 # transient error retries and sleep time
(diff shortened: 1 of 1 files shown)


youngyjd (commented):
We also don't need to raise an exception without publish_tag if this config is set to False

https://github.com/amundsen-io/amundsen/blob/3e3b0b189d6ce2af0d42bdc4838ce2903a4bca18/databuilder/databuilder/publisher/neo4j_csv_publisher.py#L159

dkunitsk:
👁️

dkunitsk:
I'd like to discourage this approach, which essentially tries to make some nodes "look" like they were created in the UI in order to avoid their deletion when using the stale data remover. 

There is already a better way for this, which we introduced some months back: https://github.com/amundsen-io/amundsen/tree/main/databuilder#using-node-and-relation-conditions-to-remove-stale-data

wli-brex:
> I'd like to discourage this approach, which essentially tries to make some nodes "look" like they were created in the UI in order to avoid their deletion when using the stale data remover.
> 
> There is already a better way for this, which we introduced some months back: https://github.com/amundsen-io/amundsen/tree/main/databuilder#using-node-and-relation-conditions-to-remove-stale-data

@dkunitsk Thanks for the suggestions! I'll try it out first.

youngyjd:
@dkunitsk Thanks for the feedback. It looks like the approach you pointed out is for removing stale nodes and relationships. This PR is for a different purpose and is backward compatible. The default value is still set to True.

Basically we back up these edited through UI nodes in a persistent storage. We want to make sure if our Neo4j dies, we can still restore those nodes created / edited from UI, using this csv_neo4j_publisher. We cannot use this version csv_neo4j_publisher and have to create our own version of the publisher to achieve it. We think it would be beneficial to the upstream too with the new config option. Happy to know your ideas.

dkunitsk:
That's interesting that you back up neo4j into a different database and then use the publisher to re-publish back those nodes. To me this sounds like a bit of a unique setup that doesn't easily extend beyond Brex. Can I recommend subclassing the Neo4jCsvPublisher (making a BrexNeo4jCsvPublisher) which overrides `_create_props_body` to implement this behavior? We do exactly that at Lyft for a handful of similar publisher "hacks". 

If that's not feasible for you, feel free to go ahead with this approach.

youngyjd:
> That's interesting that you back up neo4j into a different database and then use the publisher to re-publish back those nodes.

We have Neo4j running with persistent volume in kubernetes. However, if someone accidentally updates some descriptions and would like to restore then to previous states, it would be hard. neo4j-admin backup/restore tool requires downtime unless we opts in for Enterprise edition. Hence we decided to only back up neo4j nodes that have been edited on UI into a different database. With this setup, we can easily revert edited descriptions to whatever version customers want. We would also be able to track change histories for UI edited descriptions and collect some engagement metrics.

PR #1819, opened 2022-04-19
feat: added highlight options to search API definition

allisonsuarez:
Related PR for Search Service #1820 
Added optional `highlight_options` field to search request, these allow highlighting to be enabled for each resource like:
```
{
    "results_per_page": 10,
    "page_index": 0,
    "resource_types": [
        "table"
    ],
    "query_term": "amundsen",
    "filters": [],
    "highlight_options": {
        "table": {
            "enable_highlight": true
        }
    }
}
```

### DIFF
--- a/common/amundsen_common/models/search.py
+++ b/common/amundsen_common/models/search.py
@@ -1,11 +1,23 @@
 # Copyright Contributors to the Amundsen project.
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import List, Optional, Dict
+from typing import Any, List, Optional, Dict
 
 import attr
 
 from marshmallow3_annotations.ext.attrs import AttrsSchema
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class HighlightOptions:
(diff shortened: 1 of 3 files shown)


common/amundsen_common/models/search.py
```diff
@@ -28,6 +28,8 @@ class SearchRequest:
     page_index: Optional[int] = 0
     results_per_page: Optional[int] = 10
     filters: List[Filter] = []
+    highlight_matches: bool = False
```
dkunitsk, line 31: single "highlighting_options" field is better.

allisonsuarez: so does that have a subfield like `highlight_matches`? I want to make sure we provide an option to just enable highlight with some default fields if the field is set to`true`, and then override that functionality if `false`.

dkunitsk: yes, has nested field like `enabled: true` and others.

dkunitsk (approved):

allisonsuarez:
💨

PR #1823, opened 2022-04-20
fix: issue with neo4j query for ES data

allisonsuarez:
- removed some unnecessary fields
- made fields which are used for filters lowercase to match FE input requirement
- fixed issue where columns where duplicated over and over because of how usage was calculate

### DIFF
--- a/databuilder/databuilder/task/search/search_data_queries.py
+++ b/databuilder/databuilder/task/search/search_data_queries.py
@@ -11,27 +11,23 @@ NEO4J_TABLE_CYPHER_QUERY = textwrap.dedent(
     MATCH (db:Database)<-[:CLUSTER_OF]-(cluster:Cluster)
     <-[:SCHEMA_OF]-(schema:Schema)<-[:TABLE_OF]-(table:Table)
     {publish_tag_filter}
-    OPTIONAL MATCH (table)-[:DESCRIPTION]->(table_description:Description)
     OPTIONAL MATCH (schema)-[:DESCRIPTION]->(schema_description:Description)
-    OPTIONAL MATCH (table)-[:DESCRIPTION]->(prog_descs:Programmatic_Description)
-    WITH db, cluster, schema, schema_description, table, table_description,
-    COLLECT(prog_descs.description) as programmatic_descriptions
+    OPTIONAL MATCH (table)-[:DESCRIPTION]->(table_description:Description)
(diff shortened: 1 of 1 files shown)


databuilder/databuilder/task/search/search_data_queries.py
```diff
@@ -11,82 +11,76 @@
...
     <-[:SCHEMA_OF]-(schema:Schema)<-[:TABLE_OF]-(table:Table)
     {publish_tag_filter}
-    OPTIONAL MATCH (table)-[:DESCRIPTION]->(table_description:Description)
     OPTIONAL MATCH (schema)-[:DESCRIPTION]->(schema_description:Description)
-    OPTIONAL MATCH (table)-[:DESCRIPTION]->(prog_descs:Programmatic_Description)
-    WITH db, cluster, schema, schema_description, table, table_description,
-    COLLECT(prog_descs.description) as programmatic_descriptions
+    OPTIONAL MATCH (table)-[:DESCRIPTION]->(table_description:Description)
     OPTIONAL MATCH (table)-[:TAGGED_BY]->(tags:Tag) WHERE tags.tag_type='default'
-    WITH db, cluster, schema, schema_description, table, table_description, programmatic_descriptions,
```
feng-tao, line 20: is it intended to remove programmatic_descriptions from search?

allisonsuarez: Yes, if you look at search right now it doesn't actually use programmatic descriptions. If needed it can be added by configuring the query with custom values instead of default values of the variables within the query.

databuilder/databuilder/task/search/search_data_queries.py
```diff
@@ -11,82 +11,76 @@
...
     {{
         {usage_fields}
     }} AS usage,
     columns,
     column_descriptions,
     tags,
-    badges,
     {additional_field_return}
-    programmatic_descriptions
+    badges
```
dkunitsk, line 39: why move this

dkunitsk (approved):

databuilder/databuilder/task/search/search_data_queries.py
allisonsuarez: because we need a comma before the additional fields and it felt weird to make people put a comma in there

dkunitsk: oh right order doesn't matter in `to_document`

PR #1831, opened 2022-04-25
feat: Create right side column details panel

kristenarmes:
### Summary of Changes

This change adds in a right side panel to the table details page. When a column name is clicked, a panel on the right side of the page is expanded to display column details that have previously been displayed in the dropdown underneath the column name. This change is meant to free up the use of the dropdown for future nested columns work that is coming soon. See [this RFC](https://github.com/amundsen-io/rfcs/blob/master/rfcs/046-column-details-access-and-side-panels.md) for further details around design decisions for this feature.

When a column name is selected and the panel is expanded, that row is highlighted so it is easy to see which column name the current panel contents relate to. In addition, to save space within the main column table, when the panel is open the badges and usage columns are removed since that info is visible in the side panel.

For now, the existing drop down to display column details is not being removed to transition into the new usage.

Screen Shot 2022-04-27 at 11 14 48 AM

### Tests

JS tests were added to test the rendering of the right side panel and to test that the toggle panel function is called when a column name is clicked.

### Documentation

N/A

### CheckList

### DIFF
--- a/frontend/amundsen_application/static/css/_layouts.scss
+++ b/frontend/amundsen_application/static/css/_layouts.scss
@@ -5,12 +5,13 @@
 @import 'typography';
 
 $resource-header-height: 84px;
-$aside-separation-space: 30px;
+$aside-separation-space: $spacer-4;
 $screen-lg-container: 1440px;
 $header-link-height: 32px;
 $icon-header-size: 32px;
 $inner-column-size: 175px;
-$left-panel-size: 425px;
+$side-panel-size: 425px;
+$close-btn-size: 24px;
 
 .resource-detail-layout {
   height: calc(100vh - #{$nav-bar-height} - #{$footer-height});
@@ -106,7 +107,7 @@ $left-panel-size: 425px;
 
     > .left-panel {
       border-right: 4px solid $divider;
-      flex-basis: $left-panel-size;
+      flex-basis: $side-panel-size;
       flex-shrink: 0;
       min-height: min-content;
       overflow-y: auto;
@@ -156,6 +157,58 @@ $left-panel-size: 425px;
       }
     }
 
+    > .right-panel {
+      border-left: 4px solid $divider;
+      flex-basis: $side-panel-size;
+      flex-shrink: 0;
+      min-height: min-content;
+      overflow-y: auto;
+      padding: 0 24px $aside-separation-space;
+
+      .panel-header {
+        display: flex;
+        justify-content: space-between;
+      }
+
+      .panel-title {
+        @extend %text-title-w1;
+
+        margin-top: $aside-separation-space;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+
+      .btn.btn-close {
+        flex-basis: $close-btn-size;
+        flex-shrink: 0;
+        margin-top: $aside-separation-space;
+      }
+
+      .buttons-row {
+        display: flex;
+        gap: $spacer-2;
+        margin-top: $aside-separation-space;
+      }
+
+      .btn.btn-default {
+        line-height: $spacer-2;
+      }
+
+      .section-title {
+        @extend %text-title-w3;
+
+        color: $text-tertiary;
+        margin-bottom: 8px;
+      }
+
+      .editable-section,
+      .metadata-section {
+        margin-top: $aside-separation-space;
+        position: relative;
+      }
+    }
+
     > .main-content-panel {
       flex-basis: 500px;
       flex-grow: 1;
@@ -194,6 +247,7 @@ $left-panel-size: 425px;
   }
 
   .left-panel,
+  .right-panel,
   .main-content-panel {
     display: flex;
     flex-direction: column;
--- a/frontend/amundsen_application/static/css/_layouts.scss
+++ b/frontend/amundsen_application/static/css/_layouts.scss
@@ -20,7 +20,7 @@ $close-btn-size: 24px;
     border-bottom: 2px solid $divider;
     display: flex;
     height: $resource-header-height;
-    padding: 16px 24px;
+    padding: $spacer-2 $spacer-3;
 
     .icon-header {
       height: $icon-header-size;
@@ -67,7 +67,7 @@ $close-btn-size: 24px;
         flex-shrink: 0;
 
         > * {
-          margin-right: 16px;
+          margin-right: $spacer-2;
         }
 
         .header-link {
@@ -90,7 +90,7 @@ $close-btn-size: 24px;
         flex-shrink: 0;
 
         > * {
-          margin-right: 8px;
+          margin-right: $spacer-1;
 
           &:last-child {
             margin-right: 0;
@@ -106,25 +106,25 @@ $close-btn-size: 24px;
     height: calc(100% - #{$resource-header-height});
 
     > .left-panel {
-      border-right: 4px solid $divider;
+      border-right: $spacer-half solid $divider;
       flex-basis: $side-panel-size;
       flex-shrink: 0;
       min-height: min-content;
       overflow-y: auto;
-      padding: 0 24px $aside-separation-space;
+      padding: 0 $spacer-3 $aside-separation-space;
 
       > .banner {
         border: 1px solid $stroke;
         height: 40px;
-        margin: 24px 24px 0;
-        padding: 8px;
+        margin: $spacer-3 $spacer-3 0;
+        padding: $spacer-1;
       }
 
       .section-title {
         @extend %text-title-w3;
 
         color: $text-tertiary;
-        margin-bottom: 8px;
+        margin-bottom: $spacer-1;
       }
 
       .editable-section,
@@ -158,12 +158,12 @@ $close-btn-size: 24px;
     }
 
     > .right-panel {
-      border-left: 4px solid $divider;
+      border-left: $spacer-half solid $divider;
       flex-basis: $side-panel-size;
       flex-shrink: 0;
       min-height: min-content;
       overflow-y: auto;
-      padding: 0 24px $aside-separation-space;
+      padding: 0 $spacer-3 $aside-separation-space;
 
       .panel-header {
         display: flex;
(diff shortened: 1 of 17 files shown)


frontend/amundsen_application/static/css/_layouts.scss
```diff
@@ -156,6 +157,58 @@ $left-panel-size: 425px;
       }
     }
 
+    > .right-panel {
+      border-left: 4px solid $divider;
+      flex-basis: $side-panel-size;
+      flex-shrink: 0;
+      min-height: min-content;
+      overflow-y: auto;
+      padding: 0 24px $aside-separation-space;
```
Golodhros, line 166: $spacer-6?

frontend/amundsen_application/static/css/_layouts.scss
```diff
@@ -156,6 +157,58 @@ $left-panel-size: 425px;
       }
     }
 
+    > .right-panel {
+      border-left: 4px solid $divider;
```
Golodhros, line 161: $spacer-half

frontend/amundsen_application/static/css/_layouts.scss
```diff
@@ -156,6 +157,58 @@ $left-panel-size: 425px;
...
+      .panel-title {
+        @extend %text-title-w1;
+
+        margin-top: $aside-separation-space;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+
+      .btn.btn-close {
```
Golodhros, line 182: Would it work without the first .btn? if so, let's simplify

frontend/amundsen_application/static/css/_layouts.scss
```diff
@@ -156,6 +157,58 @@ $left-panel-size: 425px;
...
+
+      .btn.btn-default {
+        line-height: $spacer-2;
+      }
+
+      .section-title {
+        @extend %text-title-w3;
+
+        color: $text-tertiary;
+        margin-bottom: 8px;
```
Golodhros, line 202: $spacer-1

frontend/amundsen_application/static/js/features/ColumnList/ColumnDetailsPanel/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const EDITABLE_SECTION_TITLE = 'Description';
```
Golodhros, line 1:
```suggestion
// Copyright Contributors to the Amundsen project.
// SPDX-License-Identifier: Apache-2.0

export const EDITABLE_SECTION_TITLE = 'Description';
```

frontend/amundsen_application/static/js/features/ColumnList/ColumnDetailsPanel/index.spec.tsx
```diff
@@ -0,0 +1,278 @@
...
+      }).not.toThrow();
+    });
+
+    it('triggers the panel toggle when the X button is clicked', () => {
+      const { props, wrapper } = setup();
+
+      wrapper.find('.btn-close').simulate('click');
+
+      expect(props.togglePanel).toHaveBeenCalled();
+    });
```
Golodhros, line 92: This is not a 'render' test, more a 'lifecycle' test, as it involves user actions (clicking). Same for some of the ones below.

frontend/amundsen_application/static/js/features/ColumnList/ColumnDetailsPanel/index.spec.tsx
```diff
@@ -0,0 +1,278 @@
...
+      it('should render badges', () => {
+        const { wrapper } = setup();
+
+        const actual = wrapper.find(BadgeList).length;
+        const expected = 1;
+
+        expect(actual).toEqual(expected);
+      });
+
+      it('should not render badges', () => {
```
Golodhros, line 137: I would usually wrap this one under `describe('when no badges are passed', ...`

frontend/amundsen_application/static/js/features/ColumnList/ColumnDetailsPanel/index.spec.tsx
```diff
@@ -0,0 +1,278 @@
...
+      it('should render a description', () => {
+        const { wrapper } = setup();
+
+        const actual = wrapper.find(ColumnDescEditableText).length;
+        const expected = 1;
+
+        expect(actual).toEqual(expected);
+      });
+
+      it('should still render a description if only content is set', () => {
```
Golodhros, line 161: Nit:
describe('when only content is set', () => {
   it('should still render a description', () => {

frontend/amundsen_application/static/js/features/ColumnList/ColumnDetailsPanel/index.spec.tsx
```diff
@@ -0,0 +1,278 @@
...
+        mockStats = mockColumnDetails.stats;
+        const { wrapper } = setup();
+
+        const actual = wrapper.find(ColumnStats).length;
+        const expected = 1;
+
+        expect(actual).toEqual(expected);
+      });
+
+      it('should not render stats', () => {
```
Golodhros, line 215: Same here, if we don't set the context with the 'describe', it will be confusing to read the failures

frontend/amundsen_application/static/js/features/ColumnList/ColumnDetailsPanel/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const uniqueValueStats = stats && getUniqueValues(stats);
+
+  return (
+    <aside className="right-panel">
+      <div className="panel-header">
+        <div className="panel-title">{name}</div>
+        <button
+          type="button"
+          className="btn btn-close"
+          onClick={togglePanel.bind(null, null)}
```
Golodhros, line 64: Why the bind?

I would usually call here 'handleCloseButton', and there call 'togglePanel()'

frontend/amundsen_application/static/js/features/ColumnList/ColumnDetailsPanel/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+    badges,
+  } = columnDetails;
+
+  const normalStats = stats && filterOutUniqueValues(stats);
+  const uniqueValueStats = stats && getUniqueValues(stats);
+
+  return (
+    <aside className="right-panel">
+      <div className="panel-header">
+        <div className="panel-title">{name}</div>
```
Golodhros, line 60: This should be a header probably, not sure which level (maybe h3?). We don't want to render text on div tags, as they are not semantic.

frontend/amundsen_application/static/js/features/ColumnList/ColumnDetailsPanel/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+        >
+          {COPY_COL_NAME_LABEL}
+        </button>
+        <button
+          className="btn btn-default"
+          id="copy-col-link"
+          type="button"
+          onClick={() => {
+            const link = getColumnLink(tableParams, name);
+            navigator.clipboard.writeText(link);
```
Golodhros, line 86: Removing the temporal would probably work here better:
```suggestion
            navigator.clipboard.writeText(getColumnLink(tableParams, name));
```

Golodhros: That said, I prefer to extract a 'handleCopyButtonClick'

frontend/amundsen_application/static/js/features/ColumnList/index.tsx
```diff
@@ -357,7 +354,19 @@ const ColumnList: React.FC<ColumnListProps> = ({
...
+                {columnDetails.isExpandable ? (
+                  <span
+                    className="column-name-link"
+                    role="button"
+                    tabIndex={0}
+                    onClick={toggleRightPanel.bind(null, columnDetails)}
+                    onKeyDown={toggleRightPanel.bind(null, columnDetails)}
+                  >
+                    <h3 className="column-name">{title}</h3>
+                  </span>
```
Golodhros, line 366: We usually avoid adding interactivity to non-interactive tags.

I think we would be better to use a text-styled button here, even if we need to do some CSS overrides on top of it.

Golodhros: Not sure about the 'onKeyDown' either, what's its purpose?

frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -152,6 +157,10 @@ export class TableDetail extends React.Component<
   state = {
     sortedBy: SORT_CRITERIAS.sort_order,
     currentTab: this.getDefaultTab(),
+    isRightPanelOpen: false,
+    isRightPanelPreExpanded: false,
```
Golodhros, line 161: Open and Expanded? are they different?

frontend/amundsen_application/static/js/utils/navigationUtils.ts
```diff
@@ -138,3 +140,15 @@ export function setUrlParam(key: string, value: string) {
   const queryString = qs.stringify(params);
   BrowserHistory.replace(`${location.pathname}?${queryString}`);
 }
+
+export const getColumnLink = (
```
Golodhros, line 144: Let's add a test for this util

Golodhros (commented):
Looking good!

frontend/amundsen_application/static/css/_layouts.scss
allisonsuarez: 3 right? Spacer is 8px?

frontend/amundsen_application/static/js/features/ColumnList/index.tsx
kristenarmes: onKeyDown was a a11y requirement for an interactive span, but I changed it to be a button based on your suggestion so it's removed now

frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
kristenarmes: I have the "preexpanded" var that is only set once, if someone accesses the page based on the link to a specific column then it would expand the panel and set this var to true so it knew it was expanded and wouldn't keep updating the state with future renders. and then the regular "is open" var is used to keep track for the regular toggling of the panel

Golodhros (approved):
LGTM

PR #1832, opened 2022-04-26
feat: update search service to use new search mappings

allisonsuarez:
This PR leverages the new search mappings in the new databuilder task to index metadata into elasticsearch: https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/task/search/search_metadata_to_elasticsearch_task.py

**Please use the task above in order for Elasticsearch to have the correct mappings, this change checks if new mappings are indexed in ES, if not it falls back to ELASTICSEARCH_V2 which doesn't have the newest functionality** 
For a smoother transition you can follow the tutorial added to the docs in this PR.

This PR introduces the following functionality with ELASTICSEARCH_V3:

- Specifying field specific matching rather than using MultiMatch
- Support fuzziness
- The new mappings analyze text such that we can get matches with stemmed fields and other english language analysis
- Use [RankFeatures](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-rank-feature-query.html) to boost search results based on their usage
With a request like:
```
{
    "results_per_page": 10,
    "page_index": 0,
    "resource_types": [],
    "query_term": "amundsen",
    "filters": [
        {
            "operation": "AND",
            "values": [
                "ds",
                "user_id"
            ],
            "name": "column"
        }
    ]
}
```
This version of ES proxy would build a query to Elasticsearch that looks something like:
https://gist.github.com/allisonsuarez/c41f583a22f2f765ca4afb66d8703814

_**Note: all of this functionality can be overwritten on search by creating your own client class which extends ElasticsearchProxy**_


docs/tutorials/search-v3.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Overview
+
+Amundsen `/v2/search` supports [fuzziness](https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness), word [stemming](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/stemming.html), and [boosted search ranking](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/query-dsl-rank-feature-query.html) based on usage. In order for search to function, metadata for all searchable resources configured is indexed into Elasticsearch with a databuilder task and then Elasticsearch is queried via the search service. In Elasticsearch there are two main concepts that are fundamental to understanding how search works, [mappings](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html) and [search queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-your-data.html). Mappings define how fields are indexed into Elasticsearch by specifying the types of fields and how text fields are analyzed. The queries should then be written with the mapping in mind since there certain field types are required to perform certain types of queries efficiently.
```
dkunitsk, line 7: The versions are a bit confusing. I think it'd be good to clear up that `/v2/search` is a new unified search API between frontend and search service. And the thing that supports fuzziness etc is the newest proxy.

docs/tutorials/search-v3.md
```diff
@@ -0,0 +1,31 @@
+# Amundsen `/v2/search`
```
dkunitsk, line 1: I feel like this needs a `tl;dr` and, below that, a "to configure such and such" section.

search/search_service/config.py
```diff
@@ -31,7 +32,7 @@ class Config:
     LOG_CONFIG_FILE = None
 
     # Config used by ElastichSearch
-    ELASTICSEARCH_INDEX = 'table_search_index'
+    ELASTICSEARCH_INDEX = 'new_table_search_index'
```
dkunitsk, line 35: instead of "new" can we use our version terminology.

search/search_service/config.py
```diff
@@ -15,7 +15,8 @@
 PROXY_CLIENT_KEY = 'PROXY_CLIENT_KEY'
 PROXY_CLIENTS = {
     'ELASTICSEARCH': 'search_service.proxy.elasticsearch.ElasticsearchProxy',
-    'ELASTICSEARCH_V2': 'search_service.proxy.es_search_proxy.ElasticsearchProxy'
+    'ELASTICSEARCH_V2': 'search_service.proxy.es_proxy_v2.ElasticsearchProxyV2',
+    'ELASTICSEARCH_V3': 'search_service.proxy.es_proxy_v3.ElasticsearchProxyV3'
```
dkunitsk, line 19: idea: instead of V3, how about "V2.1" or "V2_1". I was getting anxious about calling this V3 while search is called V2 and having to explain that V2 is the API version that both V2 and V3 proxies are compatible with, although the V3 proxy is only compatible with the "version: 2" mapping. 

Doing semver takes the edge off, IMO. Makes it clear that 2.1 is a more minor evolution of whatever happened in 2.

search/search_service/proxy/es_proxy_v2.py
```diff
@@ -129,6 +130,7 @@ def __init__(self, *,
                  password: str = '',
                  client: Elasticsearch = None,
                  page_size: int = 10) -> None:
+        LOGGER.info("V2")
```
dkunitsk, line 133: not super useful

search/search_service/proxy/es_proxy_v3.py
```diff
@@ -0,0 +1,310 @@
...
+        Resource.USER: USER_MAPPING,
+    }
+
+    def __init__(self, *,
+                 host: str = None,
+                 user: str = '',
+                 password: str = '',
+                 client: Elasticsearch = None,
+                 page_size: int = 10) -> None:
+        LOGGER.info("V3")
```
dkunitsk, line 88: please make into a sentence, or remove

search/search_service/proxy/es_proxy_v3.py
```diff
@@ -0,0 +1,310 @@
...
+        super().__init__(host=host,
+                         user=user,
+                         password=password,
+                         client=client,
+                         page_size=page_size)
+
+    def get_index_for_resource(self, resource_type: Resource) -> str:
+        resource_str = resource_type.name.lower()
+        return f"new_{resource_str}_search_index"
```
dkunitsk, line 97: this shouldn't be hardcoded

search/search_service/proxy/es_proxy_v3.py
```diff
@@ -0,0 +1,310 @@
...
+        return must_clauses
+
+    def _build_should_query(self, resource: Resource, query_term: str) -> List[Q]:
+
+        # no scoring happens if there is no search term
+        if query_term == '':
+            return []
+
+        # general usage metric for searcheable resources
+        usage_metric_fields = {
```
dkunitsk, line 243: nit: rename to usage_metric_boosts. Otherwise makes it look like you're hard-coding usage metrics.

search/search_service/proxy/es_proxy_v3.py
```diff
@@ -0,0 +1,310 @@
...
+        for metric in usage_metric_fields.keys():
+            field_name = f'usage.{metric}'
+            boost = usage_metric_fields[metric]
+            rank_feature_query = RankFeature(field=field_name,
+                                             boost=boost)
+            rank_feature_queries.append(rank_feature_query)
+
+        return rank_feature_queries
+
+    def __new__(cls: Any,
```
dkunitsk, line 270: please put at the top, above __init__

search/tests/unit/proxy/v3/test_es_proxy_v3.py
```diff
@@ -0,0 +1,283 @@
...
+                                                  "search_score": 62.66787,
+                                                  "resource_type": "feature"
+                                              }
+                                          ],
+                                          "total_results": 3
+                                      }
+                                  },
+                                  status_code=200)
+        print(formatted_response)
+        print(expected)
```
dkunitsk, line 282: remove

docs/tutorials/search-v3.md
dkunitsk: Also see my other comment about doing semver.

search/search_service/proxy/es_proxy_v3.py
dkunitsk: moreover, is this an index or just an alias

search/search_service/proxy/es_proxy_v3.py
allisonsuarez: I knew I had missed something on cleanup of my debugging prints 😅

search/search_service/proxy/es_proxy_v3.py
```diff
@@ -0,0 +1,310 @@
...
+    GENERAL_MAPPING = {
+        'key': 'key',
+        'description': 'description',
+        'resource_type': 'resource_type',
+    }
+
+    TABLE_MAPPING = {
+        **GENERAL_MAPPING,
+        'badges': 'badges.keyword',
+        'tag': 'tags.keyword',
```
kristenarmes, line 35: Is this correct as singular "tag"? Noticed that it's the same for dashboard mapping, but it is "tags" for features so just wanted to check

allisonsuarez: yeah this is the way the field was named in frontend for filters for some reason

search/search_service/proxy/es_proxy_v2_1.py
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    RESOUCE_TO_MAPPING = {
+        Resource.TABLE: TABLE_MAPPING,
+        Resource.DASHBOARD: DASHBOARD_MAPPING,
+        Resource.FEATURE: FEATURE_MAPPING,
+        Resource.USER: USER_MAPPING,
+    }
+
+    def __new__(cls: Any,
```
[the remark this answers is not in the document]
dkunitsk: and pass those to the call to `__new__`

search/search_service/proxy/es_proxy_v2_1.py
```diff
@@ -0,0 +1,319 @@
...
+            index_mapping = elasticsearch_client.indices.get_mapping(index=index).get(index)
+            mapping_meta_field = index_mapping.get('mappings').get('_meta')
+            if mapping_meta_field is not None and mapping_meta_field.get('version') == 2:
+                mappings_up_to_date = True
+                break
+
+        if mappings_up_to_date:
+            # Use ElasticsearchProxyV2_1 if indexes are up to date with mappings
+            obj = super().__new__(ElasticsearchProxyV2_1)
+            obj.__init__(host=host,
```
dkunitsk, line 110:
```
If __new__ return instance of  it’s own class, then the __init__ method of newly created instance will be invoked with instance as first (like __init__(self, [, ….]) argument following by arguments passed to __new__ or call of class.  So, __init__ will called implicitly.

If __new__ method return something else other than instance of class,  then instances __init__ method will not be invoked. In this case you have to call __init__ method yourself.
```

dkunitsk: i.e. you shouldn't need this `__init__`

search/search_service/proxy/es_proxy_v2_1.py
```diff
@@ -0,0 +1,319 @@
...
+        'resource_type': 'resource_type',
+    }
+
+    RESOUCE_TO_MAPPING = {
+        Resource.TABLE: TABLE_MAPPING,
+        Resource.DASHBOARD: DASHBOARD_MAPPING,
+        Resource.FEATURE: FEATURE_MAPPING,
+        Resource.USER: USER_MAPPING,
+    }
+
```
dkunitsk, line 83: need a comment here. Overriding new is a classic "don't do this" so need to give some explanation to folks. And also lay out a plan for removing this (i.e. once enough people are using the new mappings).

search/search_service/proxy/es_proxy_v2_1.py
```diff
@@ -0,0 +1,319 @@
...
+                 page_size: int = 10) -> None:
+        super().__init__(host=host,
+                         user=user,
+                         password=password,
+                         client=client,
+                         page_size=page_size)
+
+    def get_index_alias_for_resource(self, resource_type: Resource) -> str:
+        resource_str = resource_type.name.lower()
+        alias_config = current_app.config.get(
```
dkunitsk, line 140: can we do this once at init.

dkunitsk (commented):
Should also insert some comments/logging into the ES publisher.

docs/tutorials/search-v2_1.md
```diff
@@ -0,0 +1,34 @@
+# Amundsen +
+
+The goal of this tutorial is to explain what the new search functionality offers, how to transition into using `v2/search` search service endpoint out of the box as well as how to customize it ahead of deprecation of the old search endpoints. The updated search service offers [fuzziness](https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness), word [stemming](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/stemming.html), and [boosted search ranking](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/query-dsl-rank-feature-query.html) based on usage.
```
dkunitsk, line 3: Nit:
I still think it's worth it to explain what is `v2/search` API (searches all resources and handles filtered and unfiltered search) and how it related to the proxies. 

I.e. there's a really old proxy that can't support v2/search API at all. 
There's a new proxy that supports v2/search API but doesn't have any of the search quality enhancements, since those need new mappings to be created. 
And finally there's the latest proxy, supports v2/search API and has the search quality improvements that rely on the new mappings being created in databuilder. It will fall back to the old V2 proxy if it cannot find the new mappings.

dkunitsk:
nit: for the PR description, can you make gists (https://gist.github.com/) and link them. It's too long.

dkunitsk:
also: reminder to post in slack about this change

search/search_service/config.py
```diff
@@ -4,22 +4,32 @@
...
 PROXY_CLIENTS = {
     'ELASTICSEARCH': 'search_service.proxy.elasticsearch.ElasticsearchProxy',
-    'ELASTICSEARCH_V2': 'search_service.proxy.es_search_proxy.ElasticsearchProxy'
+    'ELASTICSEARCH_V2': 'search_service.proxy.es_proxy_v2.ElasticsearchProxyV2',
+    'ELASTICSEARCH_V2_1': 'search_service.proxy.es_proxy_v2_1.ElasticsearchProxyV2_1'
 }
 
 
 class Config:
+    ELASTICSEARCH_INDEX = 'new_table_search_index'  # deprecated
```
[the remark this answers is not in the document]
allisonsuarez: this is the old way we used to configure aliases, I changed it because index and alias are not equivalent

search/search_service/config.py
```diff
@@ -4,22 +4,32 @@
...
 PROXY_PASSWORD = 'PROXY_PASSWORD'
-PROXY_CLIENT = 'PROXY_CLIENT'
-PROXY_CLIENT_KEY = 'PROXY_CLIENT_KEY'
+ELASTICSEARCH_CLIENT = 'ELASTICSEARCH_CLIENT'
+
+# Elasticsearch proxy class configuration
+ES_PROXY_CLIENT = 'ES_PROXY_CLIENT'
+ES_INDEX_ALIAS_TEMPLATE = 'ES_INDEX_ALIAS_TEMPLATE'
 PROXY_CLIENTS = {
     'ELASTICSEARCH': 'search_service.proxy.elasticsearch.ElasticsearchProxy',
```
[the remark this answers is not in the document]
allisonsuarez: These 2 configs are not deprecated

allisonsuarez:
> Should also insert some comments/logging into the ES publisher.

We already did:
https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/publisher/elasticsearch_publisher.py#L88

search/search_service/config.py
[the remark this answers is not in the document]
allisonsuarez: Remove this altogether and remove from v1 client

search/search_service/proxy/es_proxy_v2_1.py
allisonsuarez: You mean get a mapping of all aliases for all resources at init?

dkunitsk: ya

dkunitsk (approved):
Please update the info in `search/README.md` and also link your doc there near the top so it's more visible.

Golodhros (approved):

PR #1847, opened 2022-05-04
refactor: Refactor various column details and add TypeMetadata to TableColumn model

kristenarmes:

### Summary of Changes

These changes are made in anticipation of the upcoming work to add in nested columns retrieved from the backend rather than frontend parsing outlined [here](https://github.com/amundsen-io/rfcs/blob/master/rfcs/047-nested-columns.md). This removes the dependence on the columns' indices around descriptions and side panel column details, and now relies on names and keys to be able to extend the same usage to nested columns.

- Add `TypeMetadata` in frontend model as an optional `TableColumn` field - this data was already being retrieved from the metadata API so now it is being populated
- Use the column `name` directly for getting/setting column descriptions rather than its index
- Add column `key` field to `TableColumn`s to be used to keep track of which column is currently selected when the right side column details panel is open
- Remove the now unnecessary extra `col_index` field in `TableColumn`
- Also fixed the `ListSortingDropdown` to maintain the correct previous selection after the right side panel opens and closes

### Tests

General cleanup of tests for changes outlined above, and added a test for the `ListSortingDropdown` to make sure it selects the new prop for the currently selected option

### Documentation

N/A

### CheckList

### DIFF
--- a/frontend/amundsen_application/static/js/ducks/tableMetadata/api/helpers.ts
+++ b/frontend/amundsen_application/static/js/ducks/tableMetadata/api/helpers.ts
@@ -44,14 +44,15 @@ export function getRelatedDashboardSlug(key: string): string {
  */
 export function processColumns(
   columns: TableColumn[],
+  tableKey: string,
   databaseId?: string
 ): TableColumn[] {
-  return columns.map((column, index) => {
+  return columns.map((column) => {
     const nestedType = parseNestedType(column.col_type, databaseId);
 
     return {
       ...column,
-      col_index: index,
+      key: tableKey + '/' + column.name,
       children:
         nestedType && isNestedColumnsEnabled()
           ? convertNestedTypeToColumns(nestedType)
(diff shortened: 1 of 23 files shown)


frontend/amundsen_application/static/js/ducks/tableMetadata/api/helpers.ts
```diff
@@ -44,14 +44,15 @@ export function getRelatedDashboardSlug(key: string): string {
...
   databaseId?: string
 ): TableColumn[] {
-  return columns.map((column, index) => {
+  return columns.map((column) => {
     const nestedType = parseNestedType(column.col_type, databaseId);
 
     return {
       ...column,
-      col_index: index,
+      key: tableKey + '/' + column.name,
```
allisonsuarez, line 55: Do we have to rebuild the column key here? is it not indexed in neo4j?

Golodhros (approved):
LGTM

frontend/amundsen_application/static/js/ducks/tableMetadata/api/helpers.ts
kristenarmes: You're right that it is indexed, but it wasn't being returned from the metadata API. I thought about adding it to the neo4j proxy query but then realized that wouldn't guarantee that the key was available for those using other proxies, and this way would guarantee we'd always have the key in FE

PR #1852, opened 2022-05-10
fix: better behavior for search filters

allisonsuarez:
Change filters so if the field is not mapped for the resource the filter won't be applied to the query for that resource

### DIFF
--- a/search/search_service/proxy/es_search_proxy.py
+++ b/search/search_service/proxy/es_search_proxy.py
@@ -1,6 +1,7 @@
 # Copyright Contributors to the Amundsen project.
 # SPDX-License-Identifier: Apache-2.0
 
+import json
 import logging
 from typing import (
     Any, Dict, List, Union,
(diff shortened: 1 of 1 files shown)


search/search_service/proxy/es_search_proxy.py
```diff
@@ -198,21 +199,24 @@ def _build_filters(self, resource: Resource, filters: List[Filter]) -> List:
         filter_queries: List = []
 
         for filter in filters:
-            filter_name = mapping.get(filter.name) if mapping is not None \
-                and mapping.get(filter.name) is not None else filter.name
+            if mapping is not None and mapping.get(filter.name):
```
kristenarmes, line 202: nit: `if mapping is not None and mapping.get(filter.name) is not None:` for the second check to make sure the filter value doesn't evaluate to false

allisonsuarez: .get evaluates to the value or None and None is false I believe but good call I'll make it consistent

kristenarmes (approved):

dkunitsk (approved):

PR #1869, opened 2022-05-19
fix: Handle qualified tableau datasources more gracefully

alanmcruickshank:
### Summary of Changes

When unpacking tableau datasources, if a table reference had more than two parts (i.e. `database.schema.table` rather than just `table` or `schema.table`) this would raise a `KeyError` during unpacking and the script would stop abruptly.

This fix adds handling code for greater degrees of qualification and also a graceful path for dbapis which have potentially very long qualification options (e.g. `foo.bar.baz.db.schema.table`). I haven't accounted for anything more specific that _something that will work_ for now, although that would be something that could be built on at a later date. This solution solves for the most common use case where the database may be qualified in the table reference.

### Tests

No tests added as I don't believe there is an existing test which could be extended to cover this case. I'm still new to the codebase to any pointers on a sensible test to add would be excellent and I'm happy to implement one.

### Documentation

No documentation added - I believe this should be the expected behaviour.

### CheckList


alanmcruickshank:
@feng-tao - anything I need to do to help this through the review process?

feng-tao (approved):

feng-tao:
thanks @alanmcruickshank ! looks good and merge it now.

PR #1876, opened 2022-05-27
feat: Use type metadata description get/update apis

kristenarmes:

### Summary of Changes

- Use the new type metadata description get and update APIs to allow users to view and add/edit descriptions for nested columns from the UI
- The column description APIs were changed on the frontend side only to take a column key instead of column name to be consistent with the type metadata APIs
- The logic to decide whether the badges column should be displayed was updated to check if any type metadata values have any badges as well instead of checking only the top level columns

### Tests

- Added tests for get/update type metadata description APIs
- Added tests for `ColumnDescEditableText` to test when it is used by a top level column or a nested column

### Documentation

N/A

### CheckList

### DIFF
--- a/frontend/amundsen_application/api/utils/metadata_utils.py
+++ b/frontend/amundsen_application/api/utils/metadata_utils.py
@@ -10,7 +10,7 @@ from typing import Any, Dict, List
 from amundsen_common.models.dashboard import DashboardSummary, DashboardSummarySchema
 from amundsen_common.models.feature import Feature, FeatureSchema
 from amundsen_common.models.popular_table import PopularTable, PopularTableSchema
-from amundsen_common.models.table import Table, TableSchema
+from amundsen_common.models.table import Table, TableSchema, TypeMetadata
 from amundsen_application.models.user import load_user, dump_user
 from amundsen_application.config import MatchRuleObject
 from flask import current_app as app
@@ -102,6 +102,12 @@ def is_table_editable(schema_name: str, table_name: str, cfg: Any = None) -> boo
     return True
 
 
+def _set_type_metadata_is_editable(type_metadata: TypeMetadata, is_editable: bool) -> None:
+    type_metadata['is_editable'] = is_editable
+    for tm in type_metadata['children']:
+        _set_type_metadata_is_editable(tm, is_editable)
+
+
 def marshall_table_full(table_dict: Dict) -> Dict:
     """
     Forms the full version of a table Dict, with additional and sanitized fields
@@ -128,6 +134,8 @@ def marshall_table_full(table_dict: Dict) -> Dict:
     for col in columns:
         # Set editable state
         col['is_editable'] = is_editable
+        if col['type_metadata']:
+            _set_type_metadata_is_editable(col['type_metadata'], is_editable)
         # If order is provided, we sort the column based on the pre-defined order
         if app.config['COLUMN_STAT_ORDER']:
             # the stat_type isn't defined in COLUMN_STAT_ORDER, we just use the max index for sorting
--- a/frontend/amundsen_application/api/utils/metadata_utils.py
+++ b/frontend/amundsen_application/api/utils/metadata_utils.py
@@ -5,7 +5,7 @@ import logging
 
 from dataclasses import dataclass
 from marshmallow import EXCLUDE
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from amundsen_common.models.dashboard import DashboardSummary, DashboardSummarySchema
 from amundsen_common.models.feature import Feature, FeatureSchema
@@ -102,10 +102,11 @@ def is_table_editable(schema_name: str, table_name: str, cfg: Any = None) -> boo
     return True
 
 
-def _set_type_metadata_is_editable(type_metadata: TypeMetadata, is_editable: bool) -> None:
-    type_metadata['is_editable'] = is_editable
-    for tm in type_metadata['children']:
-        _set_type_metadata_is_editable(tm, is_editable)
+def _recursive_set_type_metadata_is_editable(type_metadata: Optional[TypeMetadata], is_editable: bool) -> None:
+    if type_metadata is not None:
+        type_metadata['is_editable'] = is_editable
+        for tm in type_metadata['children']:
+            _recursive_set_type_metadata_is_editable(tm, is_editable)
 
 
 def marshall_table_full(table_dict: Dict) -> Dict:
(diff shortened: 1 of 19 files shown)


frontend/amundsen_application/api/utils/metadata_utils.py
```diff
@@ -102,6 +102,12 @@ def is_table_editable(schema_name: str, table_name: str, cfg: Any = None) -> boo
     return True
 
 
+def _set_type_metadata_is_editable(type_metadata: TypeMetadata, is_editable: bool) -> None:
+    type_metadata['is_editable'] = is_editable
+    for tm in type_metadata['children']:
+        _set_type_metadata_is_editable(tm, is_editable)
```
allisonsuarez, line 108: so right now editability is only controlled down to the parent level? Lets say we want to set a nested column to be have an editable description we can't then set one of it's sub column's to not have an editable description right?

frontend/amundsen_application/api/utils/metadata_utils.py
```diff
@@ -128,6 +134,8 @@ def marshall_table_full(table_dict: Dict) -> Dict:
     for col in columns:
```
allisonsuarez, line 134: I feel like it is a little weird to have this functionality where we iterate through columns and set editable state in the marshal function. Maybe we have a function that takes columns and maps them to whatever schema we expect similar to for other properties here like owners? That way if there is any further metadata transformation needed for columns it can happen there instead?

frontend/amundsen_application/static/js/components/EditableText/index.tsx
```diff
@@ -68,53 +68,67 @@ class EditableText extends React.Component<
   }
 
   componentDidUpdate(prevProps) {
-    const { value, isEditing, refreshValue } = this.props;
-    if (prevProps.value !== value) {
-      this.setState({ value });
+    const { value: stateValue, isDisabled } = this.state;
```
allisonsuarez, line 71: If we are typing one argument can we type the isDisabled too for consistency?

frontend/amundsen_application/api/utils/metadata_utils.py
kristenarmes: That's true as of now, I followed how the columns are handled as well, since it appears that as of now it only has table level configurations to determine whether columns are editable or not. I think this could be adjusted in the future pretty easily if that were to change

frontend/amundsen_application/api/utils/metadata_utils.py
kristenarmes: I looked into it a bit more, and the `is_editable` field is not a part of the table or column schemas, and according to the comment for this marshall function it is meant to take care of "additional or sanitized fields". So this may not be the "best" place to do it but it does seem to fit in with the purpose of this function for now, and it is a little different than the owner situation since that one is filling in fields that are in the schema but just aren't populated by the metadata service

frontend/amundsen_application/api/utils/metadata_utils.py
```diff
@@ -128,6 +134,8 @@ def marshall_table_full(table_dict: Dict) -> Dict:
...
         # Set editable state
         col['is_editable'] = is_editable
+        if col['type_metadata']:
```
dkunitsk, line 137: the other functions below this do the if-not-none check themselves, so put that into _set_type_metadata_is_editable

dkunitsk (approved):

frontend/amundsen_application/api/utils/metadata_utils.py
dkunitsk: is_editable should almost certainly be on the schema so it can be a normal field and not this snowflake thing. But maybe this is not the time to go and refactor that. I vote for leaving a TODO.

frontend/amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -78,7 +79,7 @@ export function getTableDashboards(tableKey: string) {
       const { response } = e;
       let msg = '';
       if (response && response.data && response.data.msg) {
-        msg = response.data.msg;
+        ({ msg } = response.data);
       }
```
Golodhros, line 83: I think this would be the same as:
```
const msg = response?.data?.msg || '';
```

frontend/amundsen_application/static/js/ducks/tableMetadata/api/v0.ts
```diff
@@ -168,17 +169,47 @@ export function getColumnDescription(
...
 export function updateColumnDescription(
   description: string,
-  columnName: string,
+  columnKey: string,
   tableData: TableMetadata
 ) {
   return axios.put(`${API_PATH}/put_column_description`, {
     description,
-    column_name: columnName,
+    column_name: columnKey.substring(columnKey.lastIndexOf('/') + 1),
```
Golodhros, line 177: What if we extract a `extractColumnName(columnKey)` function?

frontend/amundsen_application/static/.betterer.results
```diff
@@ -65,22 +65,11 @@ exports[`eslint`] = {
...
-      [83, 53, 21, "Must use destructuring state assignment", "1159122654"],
-      [85, 6, 13, "Do not use setState in componentDidUpdate", "57229240"],
-      [90, 4, 22, "Must use destructuring props assignment", "2225424112"],
-      [94, 4, 22, "Must use destructuring props assignment", "2225424112"],
-      [98, 27, 23, "Must use destructuring props assignment", "2982501243"],
-      [101, 23, 23, "Must use destructuring props assignment", "2982501243"],
-      [109, 6, 22, "Must use destructuring props assignment", "2225424112"],
-      [116, 4, 24, "Must use destructuring props assignment", "3049746099"],
-      [134, 19, 20, "Script URL is a form of eval.", "3373049033"],
-      [146, 8, 207, "A control must be associated with a text label.", "2914986446"]
```
Golodhros, line 83: Thanks a lot!

frontend/amundsen_application/static/js/ducks/tableMetadata/api/helpers.ts
```diff
@@ -115,3 +115,29 @@ export function getColumnCount(columns: TableColumn[]) {
...
+export function getTypeMetadataFromKey(
+  tmKey: string,
+  tableData: TableMetadata
+) {
+  const tmNamePath = tmKey.replace(tableData.key + '/', '');
+
+  const [
+    columnName,
+    typeConstant, // eslint-disable-line @typescript-eslint/no-unused-vars
+    topLevelTmName, // eslint-disable-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
Golodhros: I see, makes sense to complain with two of those.
Whatever works will be fine.

PR #1887, opened 2022-06-06
fix: Highlighting styling improvement

allisonsuarez:
- added more spacing for matching columns and other metadata field matches
- Removed gray text


Golodhros (approved):
LGTM

PR #1895, opened 2022-06-09
feat: added ngram subfield with no stemming on ES mappings

allisonsuarez:
Added mapping subfields that contain the tokenized text with no stemming so we can prioritize exact text matches over stemmed ones when possible, using ngram to allow combinations of words too.

### DIFF
--- a/databuilder/databuilder/task/search/document_mappings.py
+++ b/databuilder/databuilder/task/search/document_mappings.py
@@ -66,6 +66,10 @@ class Subfield:
     alphanumeric = Text(analyzer=Analyzer.alphanum_analyzer,
                         term_vector=POSITIONS_OFFSETS)
 
+    alphanumeric_multi = Text(multi=True,
+                              analyzer=Analyzer.alphanum_analyzer,
+                              term_vector=POSITIONS_OFFSETS)
+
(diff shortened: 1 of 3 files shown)


databuilder/databuilder/task/search/document_mappings.py
```diff
@@ -57,31 +57,55 @@ class Analyzer:
...
+
 # Resource Mappings
 
 
 class SearchableResource(Document):
     # For better understanding of field type rationale read "Mapping unstructured content"
     # https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#wildcard-field-type
     key = Text(required=True,
-               fields={"keyword": Keyword()},
+               fields={"keyword": Subfield.keyword},
```
dkunitsk, line 87: this is the same instance every time - is that okay?

allisonsuarez: it doesn't matter because what elasticsearch-dsl does is it turns this into json and makes the request

allisonsuarez: if you look at the analyzers and stuff too in the library `analyzer = Analyzer._type_shortcut`

dkunitsk (approved):

PR #1897, opened 2022-06-10
feat: Add clickable rows to table detail page and new expand/collapse arrow icons

kristenarmes:

### Summary of Changes

- Added clickable rows to Table Detail Page to allow users to expand the right side panel by clicking on the row instead of only the column name, and included a hover box shadow that matches the same effect used on the dashboard page
- The active row that has been clicked is now indigo rather than dark gray to differentiate more from nested column rows
- Use new tree structure arrow icons that point right when a row is collapsed and down when a row is expanded
- Added the `Request Description` link to the right side panel and removed the action column from the main table since it was now redundant
- Fixed the `Copy column name` button in the right side panel to copy the whole column name path rather than only the name itself to be more useful

Rows all expanded:
Screen Shot 2022-06-10 at 10 52 44 AM

Rows collapsed:
Screen Shot 2022-06-10 at 10 54 57 AM

### Tests

- Added tests for clickable rows in the table component
- Added tests for the `Request Description` link in the side panel
- Removed tests around the action buttons column since that no longer exists
- Changed the `RequestDescriptionText` component to accept props, so updated the tests to reflect that

### CheckList

### DIFF
--- a/frontend/amundsen_application/static/css/_layouts.scss
+++ b/frontend/amundsen_application/static/css/_layouts.scss
@@ -209,6 +209,10 @@ $close-btn-size: 24px;
         margin-top: $aside-separation-space;
         position: relative;
       }
+
+      .editable-text {
+        font-size: 16px;
+      }
     }
 
     > .main-content-panel {
--- a/frontend/amundsen_application/static/js/components/Table/index.tsx
+++ b/frontend/amundsen_application/static/js/components/Table/index.tsx
@@ -3,15 +3,11 @@
 
 import * as React from 'react';
 
-import { FormattedDataType } from 'interfaces/ColumnList';
+import { ContentType, FormattedDataType } from 'interfaces/ColumnList';
+import { IconSizes } from 'interfaces/Enums';
 
 import ShimmeringResourceLoader from '../ShimmeringResourceLoader';
-import {
-  UpIcon,
-  DownIcon,
-  DoubleChevronUp,
-  DoubleChevronDown,
-} from '../SVGIcons';
+import { RightTriangleIcon, DownTriangleIcon } from '../SVGIcons';
 
 import {
   ARRAY_KIND,
@@ -70,6 +66,8 @@ export interface TableOptions {
   onExpand?: (rowValues: any, key: string) => void;
   /** Callback when a row is collapsed */
   onCollapse?: (rowValues: any, key: string) => void;
+  /** Callback when a row is clicked */
+  onRowClick?: (rowValues: any, index: string) => void;
   /** Optional empty table message to be shown */
   emptyMessage?: string;
   /** Row key of the currently seleected row */
@@ -104,6 +102,7 @@ export interface TableRowProps {
   rowStyles: { height: string };
   onExpand?: (rowValues: any, key: string) => void;
   onCollapse?: (rowValues: any, key: string) => void;
+  onRowClick?: (rowValues: any, key: string) => void;
   expandRowRef?: React.RefObject<HTMLTableRowElement>;
   expandedRows: RowKey[];
   setExpandedRows: (key) => void;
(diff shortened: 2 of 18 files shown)


frontend/amundsen_application/static/js/components/Table/styles.scss
```diff
@@ -86,6 +86,19 @@ $selected-row-color: $gray15;
   &.has-child-expanded {
     border-bottom: 0;
   }
+
+  &.is-interactive-row {
+    position: relative;
+    cursor: pointer;
+
+    &:hover {
+      box-shadow: 0 0 1px 0 rgba(0, 0, 0, 0.12), 0 2px 3px 0 rgba(0, 0, 0, 0.16);
```
allisonsuarez, line 95: does this match the hover in the search page? If so can we make it a variable?

frontend/amundsen_application/static/css/_layouts.scss
```diff
@@ -209,6 +209,10 @@ $close-btn-size: 24px;
         margin-top: $aside-separation-space;
         position: relative;
       }
+
+      .editable-text {
+        font-size: 16px;
```
Golodhros, line 214: Let's use $w2-font-size

frontend/amundsen_application/static/js/components/Table/index.tsx
```diff
@@ -557,7 +555,11 @@ const ExpandingButton: React.FC<ExpandingButtonProps> = ({
         }}
```
Golodhros, line 555: Nit: I would extract a 'handleExpandButton' function above.

frontend/amundsen_application/static/js/components/Table/index.tsx
```diff
@@ -585,20 +588,31 @@ const TableRow: React.FC<TableRowProps> = ({
...
     <React.Fragment key={columnKey}>
       <tr
         className={`ams-table-row ${
           rowValues.isNestedColumn ? 'is-nested-column-row' : ''
-        } ${currentSelectedKey === columnKey ? 'is-selected-row' : ''}`}
+        } ${currentSelectedKey === columnKey ? 'is-selected-row' : ''} ${
+          onRowClick && !isFrontendParsedNestedColumn
+            ? 'is-interactive-row'
+            : ''
+        }`}
```
Golodhros, line 611: Nit: This is a bit too much now, I would create a 'rowClasses' temporal variable and move the logic there.

frontend/amundsen_application/static/js/pages/TableDetailPage/RequestDescriptionText/index.tsx
```diff
@@ -10,40 +10,40 @@ import { bindActionCreators } from 'redux';
...
-  RequestDescriptionTextProps,
-  RequestDescriptionTextState
-> {
-  public static defaultProps: Partial<RequestDescriptionTextProps> = {};
-
-  openRequest = () => {
-    this.props.openRequestDescriptionDialog(
-      RequestMetadataType.TABLE_DESCRIPTION
-    );
+export const RequestDescriptionText: React.FC<RequestDescriptionTextProps> = ({
```
Golodhros, line 28: Nice! Thanks!

frontend/amundsen_application/static/js/pages/TableDetailPage/RequestDescriptionText/index.tsx
```diff
@@ -10,40 +10,40 @@ import { bindActionCreators } from 'redux';
...
-      >
-        {REQUEST_DESCRIPTION}
-      </a>
-    );
-  }
-}
+  return (
+    <a
+      className="request-description body-link"
+      href="JavaScript:void(0)"
```
Golodhros, line 40: Nit: this is really oldschool, I would do '#' nowadays

Golodhros (approved):
LGTM

Golodhros (approved):
LGTM

PR #1904, opened 2022-06-16
fix: Updating tour feature to store wildcard path instead of the url path

kristenarmes:

### Summary of Changes

Previously, if you configured a product/feature tour by using a wildcard path, the actual path of the page you visited would be stored rather than with the wildcard. This resulted in the tour showing up for every new page you visited under the wildcard. This change stores the actual configured path so that if wildcards are used, it will only show the tour on the first page you visit.

### Tests

Updated tests to handle the change to the return value of `getProductToursFor`

### Documentation

N/A

### CheckList


Golodhros (approved):
LGTM

Thanks for the fix!

PR #1905, opened 2022-06-16
fix: Fixing markdown for truncated column descriptions in the table

kristenarmes:

### Summary of Changes

Markdown styling has been available for table and column descriptions, but it would not display the styling correctly for the truncated descriptions that are displayed in the table detail page under each column name. This is a small change to address that so the markdown will display correctly.

### Tests

N/A

### Documentation

N/A

### CheckList


allisonsuarez (approved):
thank you!

feng-tao:
@kristenarmes @allisonsuarez do we allow markdown for column description as well? if so, could you point me the code where the limit of comment length for table and column level comment?

kristenarmes:
@feng-tao [Here](https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/static/js/config/config-default.ts#L16) is the default config that defines max lengths for both. Markdown has been around for both table and column descriptions, but it just wasn't displaying correctly in the description preview under each column name before

feng-tao:
interesting, I thought we only supported table comment markdown before :) It must be a long time since I look at this code

PR #1919, opened 2022-07-01
refactor: Remove all old frontend parsing for nested columns

kristenarmes:

### Summary of Changes

This change removes all the code related to the previous implementation of nested columns that used frontend parsing. This helps to avoid having two unrelated implementations of the same feature.

Things to note:
- If you are setting the `nestedColumns.isEnabled` config, that has been removed from the default config and config types since it is no longer used, so it should be removed from your configuration as well.
- If you are interested in having the new nested column display, refer to the details of how to configure the [ComplexTypeTransformer](https://github.com/amundsen-io/amundsen/blob/main/databuilder/README.md#complextypetransformer). This will set up the required backend data to display complex nested types on the frontend.

### Tests

Removed any test that were no longer relevant

### Documentation

N/A

### CheckList


frontend/amundsen_application/static/js/config/config-default.ts
```diff
@@ -385,7 +385,6 @@ const configDefault: AppConfig = {
     isEnabled: false,
   },
   nestedColumns: {
-    isEnabled: true,
```
Golodhros, line 388: Are there any documentation that we need to update due to the removal of this option?

Golodhros (approved):
LGTM

frontend/amundsen_application/static/js/config/config-default.ts
kristenarmes: There is not, I referred to the original PRs for the addition of these changes to make sure I covered all the parts to be removed, and didn't come across any related docs

PR #1924, opened 2022-07-05
feat: Sticky TabsComponent and Table headers

kristenarmes:

### Summary of Changes

- Make the `TabsComponent` and the `Table` header sticky when you scroll on pages that use them
- Hide the Expand/Collapse All button in the top right of the `TableDetailPage` when the page is resized to smaller than a set width to not overflow on top of the tabs (this action is still available by clicking the arrow in the header in the top right of the table)
- Small bug fix to close the right side panel on the `TableDetailPage` if it is open when the tab is changed

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
+++ b/frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
@@ -238,12 +238,12 @@ export class TableDetail extends React.Component<
     const minWidth = isRightPanelOpen
       ? Constants.MIN_WIDTH_DISPLAY_BTN_WITH_OPEN_PANEL
       : Constants.MIN_WIDTH_DISPLAY_BTN;
+    let newState = { isExpandCollapseAllBtnVisible: false };
 
     if (window.matchMedia(`(min-width: ${minWidth}px)`).matches) {
-      this.setState({ isExpandCollapseAllBtnVisible: true });
-    } else {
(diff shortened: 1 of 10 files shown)


frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx
```diff
@@ -223,6 +233,19 @@ export class TableDetail extends React.Component<
...
+    const { isRightPanelOpen } = this.state;
+    const minWidth = isRightPanelOpen
+      ? Constants.MIN_WIDTH_DISPLAY_BTN_WITH_OPEN_PANEL
+      : Constants.MIN_WIDTH_DISPLAY_BTN;
+
+    if (window.matchMedia(`(min-width: ${minWidth}px)`).matches) {
+      this.setState({ isExpandCollapseAllBtnVisible: true });
+    } else {
+      this.setState({ isExpandCollapseAllBtnVisible: false });
+    }
```
Golodhros, line 246: Nit, I would do:
```suggestion
    let newState = { isExpandCollapseAllBtnVisible: false };
    
    if (window.matchMedia(`(min-width: ${minWidth}px)`).matches) {
      newState = { isExpandCollapseAllBtnVisible: true };
    }
    this.setState(newState);
```
So we asses the new state, execute it only once. I also saves one branch!

Issue #1875, opened 2022-05-26
bug: Links in the Announcement html content don't work properly in homepage


ozandogrultan:
## Expected Behavior
Links in the Announcement html contents should not be overridden by the Announcement list page navigation

## Current Behavior
When there are links in the Announcement html content, they are overridden by the navigation of the Card component and when they are clicked, announcement list page is shown instead of the link itself.

## Possible Solution
N/A

## Steps to Reproduce

1. Have an Announcement with link in the html content
2. In the homepage, click the link
3. Announcement list page opens up, instead of the link itself

## Screenshots (if appropriate)
N/A

## Context

This causes poor UX for all users.

## Your Environment

* Amunsen version used:
* Data warehouse stores:
* Deployment (k8s or native):
* Link to your fork or repository:

sagar-raythatha:
+1

ozandogrultan:
This issue still persists, so it is not stale

ozandogrultan:
This issue still persists, so it is not stale

ozandogrultan:
https://github.com/amundsen-io/amundsen/pull/1934 is created for the fix

PR #1934, opened 2022-07-19
fix: Fix links in Announcements

ozandogrultan:
Signed-off-by: Ozan Dogrultan <ozan.dogrultan@deliveryhero.com>

Fixes https://github.com/amundsen-io/amundsen/issues/1875

### Summary of Changes

Previously, when there were links in the Announcement html content, they were overridden by the navigation of the Card component and when they were clicked, announcement list page was shown instead of the link itself.

The changes in the PR introduces the logic the stop bubbling up the click event to the Card component so that the link navigation works as expected.

### Tests

No test change is needed.

### Documentation

No documentation change is needed.

### CheckList


feng-tao (approved):

feng-tao:
cc @Golodhros

ozandogrultan:
@Golodhros could you review this please? 🙏🏼

frontend/amundsen_application/static/js/components/Card/index.tsx
```diff
@@ -44,7 +45,17 @@ const Card: React.FC<CardProps> = ({
...
+        {copy && (
+          <div
+            className="card-copy"
+            role="button"
+            tabIndex={0}
+            onClick={stopEvent}
+            onKeyDown={stopEvent}
+          >
+            {copy}
+          </div>
```
Golodhros, line 57: This creation of a fake button and adding a stop propagation handler to it seems a bit fishy to me.
Could we achieve the same effect by using CSS pointer-events?
Check in https://css-tricks.com/almanac/properties/p/pointer-events/

Golodhros (changes requested):

frontend/amundsen_application/static/js/components/Card/index.tsx
Golodhros: It is something like this: https://stackoverflow.com/a/46707009

ozandogrultan: I've tried this suggestion as well, but it didn't work either.. but I might be doing something wrong, I don't have much experience with `pointer-events`

Another option in my opinion is that eventPropagation can be stopped on the SanitizedHtml level [here](https://github.com/deliveryhero/amundsen/blob/main/frontend/amundsen_application/static/js/components/Announcements/AnnouncementsList/index.tsx#L49)

ozandogrultan: if we stop the propagation in the "html_content" level it also works, but for me it's just another workaround and doesn't solve the problem fully

ozandogrultan:
@Golodhros I've changed the fix strategy slightly with the latest changes. Now the only change is to allow `onclick` handlers to be customized in the sanitized announcement content, so stopping event propagation logic can be set if needed by the content which is passed. Can you please review again?

Golodhros (approved):
LGTM

PR #1939, opened 2022-07-22
feat: Exclude stats icon if configured stat types are the only ones present

kristenarmes:

### Summary of Changes

This change adds a config to allow adding a list of stat types where, if they are the only ones present for a column, the stats icon will not be displayed next to that column's name. This is useful in the case of columns only having very common stats, such as column usage, so the users will not click on the column expecting to see more info. The default value is an empty list so this will only affect those who configure it.

### Tests

Added a test for the config, and modified tests for determining whether the icon is displayed for a given column

### Documentation

N/A

### CheckList

### DIFF
--- a/frontend/amundsen_application/static/js/config/config-default.ts
+++ b/frontend/amundsen_application/static/js/config/config-default.ts
@@ -285,7 +285,7 @@ const configDefault: AppConfig = {
         },
       },
       stats: {
-        statTypesToExcludeForIcon: [],
+        iconNotRequiredTypes: [],
       },
       notices: {},
       searchHighlight: {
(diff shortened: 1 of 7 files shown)


frontend/amundsen_application/static/js/config/config-default.ts
```diff
@@ -284,6 +284,9 @@ const configDefault: AppConfig = {
           iconPath: '/static/images/github.png',
         },
       },
+      stats: {
+        statTypesToExcludeForIcon: [],
```
allisonsuarez, line 288: `iconNotRequiredTypes` ? or something like that makes the intended behavior clearer I think

allisonsuarez (approved):
LGTM just comment on config naming

PR #1942, opened 2022-07-25
feat: Neo4j 4.x support

allisonsuarez:
- [Made `trust` and `encrypted` options from driver configuration optional](https://github.com/neo4j/neo4j-python-driver#connection-settings-breaking-change)
- Added configuration to specify a database name
- Added configuration to specify a driver object instead of the components to create one


databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -66,15 +67,29 @@ def _get_driver(self) -> Any:
...
-                                    max_connection_lifetime=self.conf.get_int(
-                                        Neo4jExtractor.NEO4J_MAX_CONN_LIFE_TIME_SEC),
-                                    auth=(self.conf.get_string(Neo4jExtractor.NEO4J_AUTH_USER),
-                                          self.conf.get_string(Neo4jExtractor.NEO4J_AUTH_PW)),
-                                    encrypted=self.conf.get_bool(Neo4jExtractor.NEO4J_ENCRYPTED),
-                                    trust=trust)
+        # The config settings 'encrypted' and 'trust' can only be used with the URI
+        # schemes ['bolt', 'neo4j'].
+        uri = self.graph_url
+        if uri.startswith('bolt:') or uri.startswith('neo4j:'):
```
dkunitsk, line 73: i am not sure that this determines anything

databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -66,15 +67,29 @@ def _get_driver(self) -> Any:
...
+            trust = neo4j.TRUST_SYSTEM_CA_SIGNED_CERTIFICATES if self.conf.get_bool(Neo4jExtractor.NEO4J_VALIDATE_SSL) \
+                else neo4j.TRUST_ALL_CERTIFICATES
+            driver = \
+                GraphDatabase.driver(uri=uri,
```
dkunitsk, line 77: not clear why this is necessary

databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -88,16 +103,21 @@ def _get_extract_iter(self) -> Iterator[Any]:
...
-                self.results = session.read_transaction(self._execute_query)
-
-            for result in self.results:
-                if hasattr(self, 'model_class'):
-                    obj = self.model_class(**result)
-                    yield obj
-                else:
-                    yield result
+        db_name = self.conf.get(Neo4jExtractor.NEO4J_DATABASE_NAME, None)
+        if db_name:
```
dkunitsk, line 107: if i'm not mistaken, `neo4j.DEFAULT_DATABASE` can be used as the db_name if one is not passed in, eliminating this check.

databuilder/databuilder/extractor/neo4j_extractor.py
allisonsuarez: oh I see https://github.com/neo4j/neo4j-python-driver#connecting-with-python-driver-4x-to-neo4j-35

dkunitsk: i meant in general having the two-sections does not make sense to me yet. we can chat offline.

databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -65,16 +63,13 @@ def close(self) -> None:
...
         """
-        trust = neo4j.TRUST_SYSTEM_CA_SIGNED_CERTIFICATES if self.conf.get_bool(Neo4jExtractor.NEO4J_VALIDATE_SSL) \
-            else neo4j.TRUST_ALL_CERTIFICATES
         return GraphDatabase.driver(uri=self.graph_url,
                                     max_connection_lifetime=self.conf.get_int(
                                         Neo4jExtractor.NEO4J_MAX_CONN_LIFE_TIME_SEC),
                                     auth=(self.conf.get_string(Neo4jExtractor.NEO4J_AUTH_USER),
-                                          self.conf.get_string(Neo4jExtractor.NEO4J_AUTH_PW)),
-                                    encrypted=self.conf.get_bool(Neo4jExtractor.NEO4J_ENCRYPTED),
-                                    trust=trust)
```
dkunitsk, line 77: we still need these two for people running older neo4j to be able to explicitly disable encryption or configure different certificate trust policies

dkunitsk: or maybe i'm wrong. See https://neo4j.com/docs/python-manual/current/client-applications/#python-driver-connection-uris

databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -43,7 +47,22 @@ def init(self, conf: ConfigTree) -> None:
         self.conf = conf.with_fallback(Neo4jExtractor.DEFAULT_CONFIG)
         self.graph_url = conf.get_string(Neo4jExtractor.GRAPH_URL_CONFIG_KEY)
         self.cypher_query = conf.get_string(Neo4jExtractor.CYPHER_QUERY_CONFIG_KEY)
-        self.driver = self._get_driver()
+
+        driver = conf.get(Neo4jExtractor.NEO4J_DRIVER, None)
+        if driver and isinstance(driver, Driver):
+            self.driver = driver
```
dkunitsk, line 53: should check that the other arguments are not passed

databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -25,15 +26,18 @@ class Neo4jExtractor(Extractor):
...
+    NEO4J_DATABASE_NAME = 'neo4j_database'
     NEO4J_MAX_CONN_LIFE_TIME_SEC = 'neo4j_max_conn_life_time_sec'
     NEO4J_ENCRYPTED = 'neo4j_encrypted'
     """NEO4J_ENCRYPTED is a boolean indicating whether to use SSL/TLS when connecting."""
     NEO4J_VALIDATE_SSL = 'neo4j_validate_ssl'
     """NEO4J_VALIDATE_SSL is a boolean indicating whether to validate the server's SSL/TLS cert against system CAs."""
+    NEO4J_DRIVER = 'neo4j_driver'
 
-    DEFAULT_CONFIG = ConfigFactory.from_dict({NEO4J_MAX_CONN_LIFE_TIME_SEC: 50,
-                                              NEO4J_ENCRYPTED: True,
```
dkunitsk, line 35: pretty sure this is a break?

databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -43,7 +47,22 @@ def init(self, conf: ConfigTree) -> None:
         self.conf = conf.with_fallback(Neo4jExtractor.DEFAULT_CONFIG)
         self.graph_url = conf.get_string(Neo4jExtractor.GRAPH_URL_CONFIG_KEY)
         self.cypher_query = conf.get_string(Neo4jExtractor.CYPHER_QUERY_CONFIG_KEY)
-        self.driver = self._get_driver()
+
+        driver = conf.get(Neo4jExtractor.NEO4J_DRIVER, None)
+        if driver and isinstance(driver, Driver):
+            self.driver = driver
+        elif driver and not isinstance(driver, Driver):
```
dkunitsk, line 54: on the other hand, this is less important to check since usages of another type will result in "x does not have callable foo" type of errors

databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -25,15 +26,18 @@ class Neo4jExtractor(Extractor):
     MODEL_CLASS_CONFIG_KEY = 'model_class'
     NEO4J_AUTH_USER = 'neo4j_auth_user'
     NEO4J_AUTH_PW = 'neo4j_auth_pw'
+    NEO4J_DATABASE_NAME = 'neo4j_database'
```
dkunitsk, line 29: nit: lets leave a comment like: `newer neo4j backends have multi-db support`

databuilder/databuilder/extractor/neo4j_extractor.py
```diff
@@ -43,7 +47,22 @@ def init(self, conf: ConfigTree) -> None:
...
+        driver = conf.get(Neo4jExtractor.NEO4J_DRIVER, None)
+        if driver and isinstance(driver, Driver):
+            self.driver = driver
+        elif driver and not isinstance(driver, Driver):
+            msg = f'Driver should be of type neo4j.Driver, but an object of type {type(driver)} was given.'
+            LOGGER.error(msg)
+            raise TypeError(msg)
+        else:
+            self.driver = create_neo4j_driver(uri=self.graph_url,
+                                              max_connection_lifetime=conf.get_int(
```
dkunitsk, line 60: i believe because you're not using self.conf, the default fallbacks will not be read

databuilder/databuilder/extractor/neo4j_extractor.py
dkunitsk: encrypted was true by default in the old client. in the new client, it is off by default. One thing I am not clear about is what specifying it explicitly as "true" in the new client will do. Ideally i'd just copy the behavior of the old.

dkunitsk: the motivation is we don't want folks who were used to getting encryption to suddenly not get it. We'd want (assuming the server neo4j is new enough) for the behavior people get to be unchanged if they don't do anything. At least when it comes to really significant things like encryption. If that's not possible, they should get an error asking them to change their parameters.

databuilder/databuilder/extractor/neo4j_extractor.py
dkunitsk: that way, if both driver and other params are passed, it's clear what will happen.

dkunitsk (approved):

mgorsk1:
so do I understand it right no changes were required to make metadata microservice compatible @allisonsuarez ?

allisonsuarez:
Metadata will require changes as well, the version of the neo4j driver we are using for metadata has not changed @mgorsk1

PR #1944, opened 2022-07-28
fix: driver object pickle error

allisonsuarez:
Removed neo4j driver config since it can't be used with out config library


dkunitsk (approved):
Thanks

PR #1952, opened 2022-08-04
feat: Neo4j driver 4.4.5 on metadata

allisonsuarez:
- Updated neo4j python driver to v4.4.5
- Added `PROXY_DATABASE_NAME` to specify database to use if your neo4j version supports multiple databases
- Made `PROXY_PORT` config optional
**Note:** this is the bare minimum necessary changes to support newer neo4j versions, a more extensive refactor is required to get the neo4j_proxy working in line with the most up to date [Cypher](https://neo4j.com/docs/cypher-manual/current/deprecations-additions-removals-compatibility/)/[Neo4j driver](https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.Session.read_transaction) changes.


metadata/metadata_service/proxy/neo4j_proxy.py
```diff
@@ -520,14 +539,15 @@ def _safe_get(self, dct, *keys):
...
         if LOGGER.isEnabledFor(logging.DEBUG):
             LOGGER.debug('Executing Cypher query: {statement} with params {params}: '.format(statement=statement,
                                                                                              params=param_dict))
         start = time.time()
         try:
-            with self._driver.session() as session:
-                return session.run(statement, **param_dict)
+            with self._driver.session(database=self._database_name) as session:
+                result = session.run(query=statement, **param_dict)
+                return [record for record in result]
```
dkunitsk, line 550: why. IIUC result is not actually a list but it does already implement the list interface methods, so i don't see a reason for this.

allisonsuarez: result needs to be consumed within the session or it ends up being []

metadata/metadata_service/proxy/neo4j_proxy.py
```diff
@@ -99,10 +118,10 @@ def health(self) -> health_check.HealthCheck:
         try:
             # dbms.cluster.overview() is only available for enterprise neo4j users
             cluster_overview = self._execute_cypher_query(statement='CALL dbms.cluster.overview()', param_dict={})
-            checks = dict(cluster_overview.single())
+            checks = dict(safe_list_get(cluster_overview, 0))
```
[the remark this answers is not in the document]
allisonsuarez: same problem with having to consume the result immediately

metadata/metadata_service/proxy/atlas_proxy.py
```diff
@@ -76,6 +76,7 @@ def __init__(self, *,
                  password: str = '',
                  encrypted: bool = False,
                  validate_ssl: bool = False,
+                 database_name: str = None,
```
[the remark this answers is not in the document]
allisonsuarez: `get_proxy_client` calls the constructor on the configured proxy class, so the constructors have to match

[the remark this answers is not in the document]
dkunitsk: i believe the pythony way would be to give this function an `*args, **kwargs` so it won't complain about extra arguments

metadata/metadata_service/proxy/neo4j_proxy.py
dkunitsk: found something instructing to do exactly this: https://community.neo4j.com/t5/drivers-stacks/python-neo4j-driver-get-result-from-session-run/td-p/13567
Thanks!

metadata/metadata_service/proxy/neo4j_proxy.py
```diff
@@ -54,40 +56,54 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def safe_list_get(lst: List, idx: int) -> Any:
```
dkunitsk, line 59: recommendation:
1. since you're only ever using this to get the first element, calling it `unlist` and removing the idx argument may be more obvious.
2. same as the `single()` method used to do, perhaps its good to log a warning if >1 item exists

dkunitsk (approved):
Nice. Only two outstanding nits:
1. about the `safe_list_get` being `unlist`
2. about using `*args, **kwargs`

metadata/metadata_service/proxy/neo4j_proxy.py
allisonsuarez: We already had a safe map key get some I was trying to follow this, but I could also have it be `get_single_result` or something, the word `unlist` feels not descriptive enough to me

metadata/metadata_service/proxy/atlas_proxy.py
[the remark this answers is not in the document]
allisonsuarez: good point

PR #1938, opened 2022-07-21
chore: migrate databuilder to neo4j-driver 4.4.5

allisonsuarez:
## NOT BW COMPATIBLE, UPGRADE [NEO4J DB >= 3.5.5](https://neo4j.com/developer/kb/neo4j-supported-versions/)
The newest driver is compatible with neo4j DB version >= 3.5 which was released in 2018.

- Migrated amundsen-databuilder to use neo4j-driver 4.4.5
- The driver for Metadata will be upgraded in a separate PR
**Issue:** https://github.com/amundsen-io/amundsen/issues/1008

PR #1957, opened 2022-08-11
perf: New neo4j csv publisher to improve performance using batched params

kristenarmes:
### Summary of Changes

- Use `unwind` clause in the cypher queries when merging nodes and relations to be able to batch params and improve performance
- Updated to use managed transactions in line with Neo4j's recommended best practices
- Added `PUBLISH_REVERSE_RELATIONSHIPS` config
  - True (default): keep the existing behavior of publishing two way relations between nodes
  - False: only publish a one way relationship
- Added `PRESERVE_ADHOC_UI_DATA` config
  - True (default): stops the publisher from updating a node or relationship created via the UI, e.g. a description or owner added manually by an Amundsen user. Such nodes/relationships will not have a `published_tag` property that is set by databuilder.
  - False: nothing changes

### Tests

Basic unit tests for the publisher

### Documentation

N/A

### CheckList


databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+PUBLISH_REVERSE_RELATIONSHIPS = 'publish_reverse_relationships'
+
+# If enabled, stops the publisher from updating a node or relationship
+# created via the UI, e.g. a description or owner added manually by an Amundsen user.
+# Such nodes/relationships will not have a 'published_tag' property that is set by databuilder.
+PRESERVE_ADHOC_UI_DATA = 'preserve_adhoc_ui_data'
+
+# CSV HEADER
+# A header with this suffix will be pass to Neo4j statement without quote
+UNQUOTED_SUFFIX = ':UNQUOTED'
```
dkunitsk, line 87: this i think does nothing and is no longer needed.

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+    """
+    A Publisher takes two folders for input and publishes to Neo4j.
+    One folder will contain CSV file(s) for Node where the other folder will contain CSV
+    file(s) for Relationship.
+
+    Neo4j follows Label Node properties Graph and more information about this is in:
+    https://neo4j.com/docs/developer-manual/current/introduction/graphdb-concepts/
+    """
+
+    def __init__(self) -> None:
```
dkunitsk, line 135: will happen on its own

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+
+        LOGGER.info('Committed total %i statements', self._count)
+
+        # TODO: Add statsd support
+        LOGGER.info('Successfully published. Elapsed: %i seconds', time.time() - start)
+
+    def get_scope(self) -> str:
+        return 'publisher.neo4j'
+
+    def _create_indices(self, node_file: str) -> None:
```
dkunitsk, line 242: idea: allow for list of pre and post-publish actions. and set this as a default pre-publish action

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+                k = k[:-len(UNQUOTED_SUFFIX)]
+
+            props.append(f'{identifier}.{k} = row.{k}')
+
+        if self.add_publisher_metadata:
+            props.append(f"{identifier}.{PUBLISHED_TAG_PROPERTY_NAME} = '{self.publish_tag}'")
+            props.append(f"{identifier}.{LAST_UPDATED_EPOCH_MS} = timestamp()")
+
+        # add additional metadata fields from config
+        for k, v in self.additional_fields.items():
```
dkunitsk, line 453: rename additional_fields to maybe `additional_publisher_metadata_fields`

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+        :param tx:
+        :param stmt:
+        :param expect_result: By having this True, it will validate if result object is not None.
+        :return:
+        """
+
+        LOGGER.debug('Executing statement: %s with params %s', stmt, params)
+
+        result = tx.run(stmt, parameters=params)
+        if expect_result and not result.single():
```
dkunitsk, line 476: needed?

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+    def is_create_only_node(self, node_record: dict) -> bool:
+        """
+        Check if node can be updated
+        :param node_record:
+        :return:
+        """
+        if self.create_only_nodes:
+            return node_record[NODE_LABEL_KEY] in self.create_only_nodes
+        else:
+            return False
```
dkunitsk, line 314: return `self.create_only_nodes.get(node_record[NODE_LABEL_KEY], False)` can replace this if self.create_only_nodes defaults to empty dict. Also, if this is only a one liner, do we even need a separate function.

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+        # if NEO4J_VALIDATE_SSL or NEO4J_ENCRYPTED are set in config pass them to the driver
+        validate_ssl_conf = conf.get(NEO4J_VALIDATE_SSL, None)
+        encrypted_conf = conf.get(NEO4J_ENCRYPTED, None)
+        if validate_ssl_conf is not None:
+            driver_args['trust'] = neo4j.TRUST_SYSTEM_CA_SIGNED_CERTIFICATES if validate_ssl_conf \
+                else neo4j.TRUST_ALL_CERTIFICATES
+        if encrypted_conf is not None:
+            driver_args['encrypted'] = encrypted_conf
+
+        self._driver = GraphDatabase.driver(**driver_args)
```
dkunitsk, line 171: move to separate driver-initialization function, which takes `conf` as argument

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+
+        # config is list of node label.
+        # When set, this list specifies a list of nodes that shouldn't be updated, if exists
+        self.create_only_nodes = set(conf.get_list(NEO4J_CREATE_ONLY_NODES, default=[]))
+        self.labels: Set[str] = set()
+        self.publish_tag: str = conf.get_string(JOB_PUBLISH_TAG)
+        self.additional_fields: Dict = conf.get(ADDITIONAL_FIELDS)
+        self.add_publisher_metadata: bool = conf.get_bool(ADD_PUBLISHER_METADATA)
+        self.publish_reverse_relationships: bool = conf.get_bool(PUBLISH_REVERSE_RELATIONSHIPS)
+        self._preserve_adhoc_ui_data = conf.get_bool(PRESERVE_ADHOC_UI_DATA)
```
dkunitsk, line 185: let's be consistent about which instance variables are underscored and which are not

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+RELATION_START_KEY = 'START_KEY'
+# End node label
+RELATION_END_LABEL = 'END_LABEL'
+# Node node key
+RELATION_END_KEY = 'END_KEY'
+# Type for relationship (Start Node)->(End Node)
+RELATION_TYPE = 'TYPE'
+# Type for reverse relationship (End Node)->(Start Node)
+RELATION_REVERSE_TYPE = 'REVERSE_TYPE'
+# Required columns for Relationship
```
dkunitsk, line 108: get from graph_serializable.py

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,501 @@
...
+# Neo4j property name for last updated timestamp
+LAST_UPDATED_EPOCH_MS = 'publisher_last_updated_epoch_ms'
+
+# A boolean flag to indicate if publisher_metadata (e.g. published_tag,
+# publisher_last_updated_epoch_ms)
+# will be included as properties of the Neo4j nodes
+ADD_PUBLISHER_METADATA = 'add_publisher_metadata'
+
+# A boolean to indicate whether to publish reverse relationships, otherwise there will only be one direction
+PUBLISH_REVERSE_RELATIONSHIPS = 'publish_reverse_relationships'
```
dkunitsk, line 78: let's make sure to note that this is a super secret option which should not be used generally

ozandogrultan:
What is the minimum Neo4j version required to be able to use this new publisher? I would suggest to document it if there is such requirement

kristenarmes:
Hey @ozandogrultan, the Neo4j driver was upgraded in databuilder v7.0.0 (https://github.com/amundsen-io/amundsen/pull/1938) so it should work for anyone past that version

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
kristenarmes: I looked into this, it is used to specify fields where the values are not strings, so like `"order_pos:UNQUOTED"` would be for numbers 1, 2, etc

dkunitsk: I can't find a reference for this, could you share. I recall looking this up a year+ ago finding it hasn't been doing anything and wasn't really a thing since maybe the earliest versions of neo4j. I tried just now again, and didn't find anything concrete.

dkunitsk: Nevermind I found it. Clearly I was dumb back then.

dkunitsk: Okay I just looked more. IIUC: in the publisher, all that's happening is this suffix is being removed so it doesn't end up ingested. At some point, it was doing something, but not anymore (please correct me if I'm wrong). Instead of continuing to support this, instead can we simply remove this behavior from the loader and then, likewise, from all publishers?  I did a full code search and I'm quite sure this is safe! Maybe double check though.

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,471 @@
...
+
+
+class Neo4jCsvUnwindPublisher(Publisher):
+    """
+    A Publisher takes two folders for input and publishes to Neo4j.
+    One folder will contain CSV file(s) for Node where the other folder will contain CSV
+    file(s) for Relationship.
+
+    Neo4j follows Label Node properties Graph and more information about this is in:
+    https://neo4j.com/docs/developer-manual/current/introduction/graphdb-concepts/
```
dkunitsk, line 119: remove and say something about UNWIND

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,471 @@
...
+            for i in range(start_idx, stop_idx):
+                params_list.append(self._create_props_param(records[i]))
+
+            with self._driver.session(database=self._db_name) as session:
+                session.write_transaction(self._execute_statement, stmt, {'batch': params_list})
+
+            params_list.clear()
+            start_idx = start_idx + self._transaction_size
+
+    def _publish_node(self, node_file: str) -> None:
```
dkunitsk, line 268: misleading function name. Can we call it `_publish_node_file` instead. Same for rels.

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,471 @@
...
+    def _write_transactions(self,
+                            records: List[dict],
+                            stmt: str) -> None:
+        params_list = []
+        start_idx = 0
+        while start_idx < len(records):
+            stop_idx = min(start_idx + self._transaction_size, len(records))
+
+            for i in range(start_idx, stop_idx):
+                params_list.append(self._create_props_param(records[i]))
```
dkunitsk, line 260: suggestion: implement a helper function called `chunkify_list`:
```

def chunkify_list(self, records):
    """
    Generator to evenly split the input list into chunks
    """
    chunk_size = self._transaction_size
    for index in range(0, len(records), chunk_size):
        yield records[index:index + chunk_size]
```

Then here do:
```
for chunk in chunkify_list:
  params_list = []
  for record in chunk:
    params_list.append(self._create_props_param(record))
```

It's not shorter, but IMO more obviously lacking off-by-one errors.

Also, putting params_list initialization inside lets you not do params_list.clear()

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,471 @@
...
+                     col_test_id1.order_pos = 2,
+                     col_test_id1.type = 'bigint'
+
+        :param node_file:
+        :return:
+        """
+
+        with open(node_file, 'r', encoding='utf8') as node_csv:
+            all_node_records = pandas.read_csv(node_csv, na_filter=False).to_dict(orient="records")
+            if len(all_node_records) > 0:
```
dkunitsk, line 289: if this condition is false, then stmt will be undefined the the line below will fail. Maybe the case of empty files just doesn't need to be handled.

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,471 @@
...
+
+            with self._driver.session(database=self._db_name) as session:
+                session.write_transaction(self._execute_statement, stmt, {'batch': params_list})
+
+            params_list.clear()
+            start_idx = start_idx + self._transaction_size
+
+    def _publish_node(self, node_file: str) -> None:
+        """
+        Iterate over the csv records of a file, each csv record transform to Merge statement
```
dkunitsk, line 270: let's remove this lengthy comment. same for _publish_rel. The actual code is a better documentation than this re-statement of it.

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,471 @@
...
+            props.append(f'{identifier}.{k} = row.{k}')
+
+        if self._add_publisher_metadata:
+            props.append(f"{identifier}.{PUBLISHED_TAG_PROPERTY_NAME} = '{self._publish_tag}'")
+            props.append(f"{identifier}.{LAST_UPDATED_EPOCH_MS} = timestamp()")
+
+        # add additional metadata fields from config
+        for k, v in self._additional_publisher_metadata_fields.items():
+            val = v if isinstance(v, int) or isinstance(v, float) else f"'{v}'"
+            props.append(f"{identifier}.{k} = {val}")
```
dkunitsk, line 435: instead of adhoc handling, can we append this to record_dict and let it handle it.

dkunitsk: Allison recently added this and I should have pointed it out then.

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,471 @@
...
+        RETURN n1.key, n2.key
+
+        :param relation_file:
+        :return:
+        """
+
+        with open(relation_file, 'r', encoding='utf8') as relation_csv:
+            all_rel_records = pandas.read_csv(relation_csv, na_filter=False).to_dict(orient="records")
+            if len(all_rel_records) > 0:
+                stmt = self.create_relationship_merge_statement(rel_record=all_rel_records[0])
```
dkunitsk, line 336: it's a bit confusing that this takes a full record. suggestion:

```
csv_dataframe = pandas.read_csv(relation_csv, na_filter=False)
merge_statement = self.create_relationship_merge_statement(csv_dataframe.columns.tolist())
self._write_transactions(merge_statement, csv_dataframe.to_dict(orient="records"))
```
ref: https://stackoverflow.com/questions/24962908/how-can-i-read-only-the-header-column-of-a-csv-file-using-python

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,471 @@
...
+        # if NEO4J_VALIDATE_SSL or NEO4J_ENCRYPTED are set in config pass them to the driver
+        validate_ssl_conf = conf.get(NEO4J_VALIDATE_SSL, None)
+        encrypted_conf = conf.get(NEO4J_ENCRYPTED, None)
+        if validate_ssl_conf is not None:
+            driver_args['trust'] = neo4j.TRUST_SYSTEM_CA_SIGNED_CERTIFICATES if validate_ssl_conf \
+                else neo4j.TRUST_ALL_CERTIFICATES
+        if encrypted_conf is not None:
+            driver_args['encrypted'] = encrypted_conf
+
+        return GraphDatabase.driver(**driver_args)
```
dkunitsk, line 176: add the connectivity check here: https://docs.google.com/presentation/d/109rwL8_8AyRaMa2a_nyffmfRGMusVRhZSZUymFDHw3Y/mobilepresent?slide=id.g12e46523937_0_299

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,380 @@
...
+                                                                    RELATION_REQUIRED_KEYS,
+                                                                    self._additional_publisher_metadata_fields), 'r1')
+        props_body_r2 = self._create_props_body(get_props_body_keys(rel_keys,
+                                                                    RELATION_REQUIRED_KEYS,
+                                                                    self._additional_publisher_metadata_fields), 'r2')
+        props_body_create = props_body_template.render(props_body_r1=props_body_r1, props_body_r2=props_body_r2)\
+            if self._publish_reverse_relationships else props_body_r1
+
+        props_body_update = props_body_create
+        if self._preserve_adhoc_ui_data:
```
kristenarmes, line 292: I left the ifs outside of the jinja template on the create node/rel stmt functions bc it saves from doing extra work if the preserve data config isn't set, and I felt like it would still be messy looking in any case, wdyt?

databuilder/databuilder/utils/publisher_utils.py
```diff
@@ -0,0 +1,64 @@
...
+        yield records[index:index + chunk_size]
+
+
+def create_props_param(record_dict: dict, additional_publisher_metadata_fields: dict) -> dict:
+    """
+    Create a dict of all the params for a given record
+    """
+    params = {}
+
+    for k, v in {**record_dict, **additional_publisher_metadata_fields}.items():
```
kristenarmes, line 29: will remove/clean up some of these helpers in the next PR when removing UNQUOTED

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,380 @@
...
+            {% endif %}
+            RETURN n1.key, n2.key
+        """)
+
+        two_way_relationship = Template("""
+            (n1)-[r1:{{ TYPE }}]->(n2)-[r2:{{ REVERSE_TYPE }}]->(n1)
+        """).render(TYPE=relation_type, REVERSE_TYPE=relation_reverse_type)
+        one_way_relationship = Template("""
+            (n1)-[r1:{{ TYPE }}]->(n2)
+        """).render(TYPE=relation_type)
```
dkunitsk, line 278: this seems relatively straightforward to move inside the template above wdyt

dkunitsk (commented):
amazing. Minor comments raised. Very nearly done.

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        Publishes Nodes first and then Relations
+        """
+        start = time.time()
+
+        for node_file in self._node_files:
+            self.pre_publish_node_file(node_file)
+
+        for rel_file in self._relation_files:
+            self.pre_publish_rel_file(rel_file)
```
dkunitsk, line 144: move this to be after the node files are published

dkunitsk (approved):
😭 🥲 😭

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -0,0 +1,346 @@
...
+csv.field_size_limit(int(ctypes.c_ulong(-1).value // 2))
+
+# Required columns for Node
+NODE_REQUIRED_KEYS = {NODE_LABEL, NODE_KEY}
+# Required columns for Relationship
+RELATION_REQUIRED_KEYS = {RELATION_START_LABEL, RELATION_START_KEY,
+                          RELATION_END_LABEL, RELATION_END_KEY,
+                          RELATION_TYPE, RELATION_REVERSE_TYPE}
+
+DEFAULT_CONFIG = ConfigFactory.from_dict({Neo4jCsvPublisherConfigs.NEO4J_TRANSACTION_SIZE: 500,
```
dkunitsk, line 47: make this a nice round 1k i think. Don't have strong reasoning, but 500 seems small compared to the online blogs that talk about UNWIND.

feng-tao:
amazing! any perf number before vs after?

kristenarmes:
@feng-tao we've seen great improvements in a few tests I've ran, with having deleted the nodes/relations in between runs:

**Task # 1: 67676 statements**

Before: Successfully published. Elapsed: 203 seconds

After: Successfully published. Elapsed: 8 seconds

**Task # 2: 237734 statements**

Before: Successfully published. Elapsed: 876 seconds

After: Successfully published. Elapsed: 29 seconds

feng-tao:
nice, it would be great if you could share the change to the slack so that others could leverage :) (I still remembered the old days to take long time to update in the etl job)

PR #1983, opened 2022-09-13
feat: Allow null values set for empty props in neo4j unwind publisher and multiple rels between nodes

kristenarmes:

### Summary of Changes

- Added publisher config `PRESERVE_EMPTY_PROPS` to `Neo4jCsvUnwindPublisher`
  - True (default): existing behavior of publishing node or relation properties even if they have empty values
  - False: set empty properties to NULL so they will not show up on the nodes or relations
- Allow multiple relations between the same two nodes when not publishing reverse relations
  - Note: must add `key` as an attribute for the relation
  - Now has the relation `key` as part of the merge statement so unique relations will all be published
  - Appends the loader-generated key to the generated file names in case there are more than one for the same groupings of nodes and relations

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/databuilder/databuilder/loader/file_system_neo4j_csv_loader.py
+++ b/databuilder/databuilder/loader/file_system_neo4j_csv_loader.py
@@ -129,7 +129,7 @@ class FsNeo4jCSVLoader(Loader):
                     relation.type,
                     self._make_key(relation_dict))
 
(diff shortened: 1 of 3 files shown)


allisonsuarez (approved):
looks great! thanks Kristen

PR #1993, opened 2022-09-19
feat: Add configurable prop types to neo4j csv publisher

kristenarmes:

### Summary of Changes

Added publisher config `NEO4J_PROP_TYPES_TO_CONFIGURE` to `Neo4jCsvUnwindPublisher`. If used, this should be set to a dict using node or relation property names as keys mapped to the function name used to configure a specific type for that property. The values of the properties should be in the correct format that the function accepts.

For example, a config of `{'start_time': 'datetime', 'publish_tag': 'date'}` where the property values are in the format `<date>T<time>` and `<date>` would apply `datetime(n.start_time)` and `date(n.publish_tag)` in the prop merge statement to create the props as `DateTime` and `Date` types instead of strings.

### Tests

N/A

### Documentation

N/A

### CheckList

### DIFF
--- a/databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
+++ b/databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
@@ -91,6 +91,8 @@ class Neo4jCsvUnwindPublisher(Publisher):
         self._publish_reverse_relationships: bool = conf.get_bool(PublishBehaviorConfigs.PUBLISH_REVERSE_RELATIONSHIPS)
         self._preserve_adhoc_ui_data = conf.get_bool(PublishBehaviorConfigs.PRESERVE_ADHOC_UI_DATA)
         self._preserve_empty_props: bool = conf.get_bool(PublishBehaviorConfigs.PRESERVE_EMPTY_PROPS)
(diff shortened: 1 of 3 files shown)


allisonsuarez (approved):
the logic here makes perfect sense to me, thanks Kristen!

databuilder/databuilder/publisher/neo4j_csv_unwind_publisher.py
```diff
@@ -317,23 +319,38 @@ def _create_props_body(self,
         template = Template("""
             {% for k in record_keys %}
                 {% if preserve_empty_props %}
-                    {{ identifier }}.{{ k }} = row.{{ k }}
+                    {% if k in prop_types_to_configure %}
+                        {{ identifier }}.{{ k }} = {{ prop_types_to_configure[k] }}(row.{{ k }})
```
allisonsuarez, line 323: if you were to convert something to another type like string you would have to do toString instead of just string right?

kristenarmes: That's correct, whichever cypher functions are available would need to be used as the associated value in the dict (toString being one option)

PR #2004, opened 2022-10-05
fix: support capitalized table names

allisonsuarez:
Since search has to analyze a lowercase string, this change in frontend looks at the key to get the table name to display to users rather than the name coming from the search field `name`.

### DIFF
--- a/frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
+++ b/frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
@@ -24,8 +24,17 @@ export interface TableListItemProps {
   disabled?: boolean;
 }
 
+export const getName = (table) => {
+  const splitKey = table.key.split('/');
+  const keyName = splitKey[splitKey.length - 1];
+  if (keyName.toLowerCase() == table.name) {
+    return keyName;
+  }
+  return table.name;
+}
+
 export const getLink = (table, logging) =>
-  `/table_detail/${table.cluster}/${table.database}/${table.schema}/${table.name}` +
+  `/table_detail/${table.cluster}/${table.database}/${table.schema}/${getName(table)}` +
   `?index=${logging.index}&source=${logging.source}`;
 
 export const generateResourceIconClass = (databaseId: string): string =>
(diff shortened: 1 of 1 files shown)


kristenarmes (approved):
LGTM, thanks for adding the function comment for clarification about the usage!

frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
```diff
@@ -24,9 +24,26 @@ export interface TableListItemProps {
...
 }
 
-export const getLink = (table, logging) =>
-  `/table_detail/${table.cluster}/${table.database}/${table.schema}/${table.name}` +
-  `?index=${logging.index}&source=${logging.source}`;
+export const getName = (table) => {
+  /*
+  this function get's the table name from the key to preserve original
+  capitalization since search needs the names to be lowercase for analysis
+  */
```
Golodhros, line 31: In JS world we put the comments just outside the function :)

frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
```diff
@@ -24,9 +24,26 @@ export interface TableListItemProps {
...
+  const splitKey = table.key.split('/');
+  const keyName = splitKey[splitKey.length - 1];
+  if (keyName.toLowerCase() === table.name) {
+    return keyName;
+  }
+  return table.name;
+};
+
+export const getLink = (table, logging) => {
+  const name = getName(table);
```
Golodhros, line 41: nit: I would just inline the call instead of creating a temporal.

frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
```diff
@@ -24,9 +24,26 @@ export interface TableListItemProps {
...
+  */
+  const splitKey = table.key.split('/');
+  const keyName = splitKey[splitKey.length - 1];
+  if (keyName.toLowerCase() === table.name) {
+    return keyName;
+  }
+  return table.name;
+};
+
+export const getLink = (table, logging) => {
```
Golodhros, line 40: I would move this into a utils file, navigationUtils.ts will probably the best option.
I would also add a test for it.

frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
allisonsuarez: I tried but had a weird lint issue I couldn't fix or auto fix

frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
allisonsuarez: this is consistent with all of the other resource list item files

Golodhros: OK, could you add a test though?

Golodhros (approved):
LGTM

PR #2010, opened 2022-10-17
feat: added optional in_amundsen bool to lineage items

allisonsuarez:
- Allow metadata to return an `in_amundsen` field for each lineage item so that table links can be disabled if a table that showed up in lineage data doesn't currently exist in Amundsen (ex: temporary tables)


Golodhros (approved):
LGTM

kristenarmes (approved):

dkunitsk (approved):

MrwanBaghdad:
How's the `LineageItem` meant to be used with `TableLineage` object in the data builder?

PR #2014, opened 2022-10-18
fix: better styling for disabled items

allisonsuarez:
- when a list item is disabled it should look different than the ones that are still enabled to avoid user confusion

Screen Shot 2022-10-18 at 1 18 56 PM

### DIFF
--- a/frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
+++ b/frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
@@ -57,7 +57,9 @@ const TableListItem: React.FC<TableListItemProps> = ({
 }) => (
   <li className="list-group-item">
     <Link
-      className={`resource-list-item table-list-item ${disabled ? 'is-disabled' : 'clickable'}`}
+      className={`resource-list-item table-list-item ${
+        disabled ? 'is-disabled' : 'clickable'
+      }`}
       to={getLink(table, logging)}
       onClick={(e) =>
         logClick(e, {
--- a/frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
+++ b/frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
@@ -55,9 +55,9 @@ const TableListItem: React.FC<TableListItemProps> = ({
   tableHighlights,
   disabled,
 }) => (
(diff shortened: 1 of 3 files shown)


frontend/amundsen_application/static/js/components/ResourceListItem/styles.scss
```diff
@@ -14,6 +14,17 @@ $resource-item-height: 110px;
   padding: $spacer-3;
   text-decoration: none;
 
+  &.is-disabled {
+    pointer-events: none;
+    cursor: not-allowed;
+    .resource-icon {
+      filter: grayscale(80%);
```
Golodhros, line 21: Hmmm... Can you tell me more about why we use this? It is a non-conventional way.

allisonsuarez: I needed to add some sort of visual indicator that the table item is not actionable (won't go to a table page), and it felt like just the table name being gray was not enough. There is a ss showing what it looks like. Is there a better way of doing this?

frontend/amundsen_application/static/js/components/ResourceListItem/TableListItem/index.tsx
```diff
@@ -55,9 +55,9 @@ const TableListItem: React.FC<TableListItemProps> = ({
   tableHighlights,
   disabled,
 }) => (
-  <li className={`list-group-item ${disabled ? 'is-disabled' : 'clickable'}`}>
+  <li className="list-group-item">
     <Link
-      className="resource-list-item table-list-item"
+      className={`resource-list-item table-list-item ${disabled ? 'is-disabled' : 'clickable'}`}
```
Golodhros, line 60: It seems you don't have prettier and eslint configured properly, that's why you get this error.

frontend/amundsen_application/static/js/components/ResourceListItem/styles.scss
```diff
@@ -14,6 +14,17 @@ $resource-item-height: 110px;
...
+    }
+    .resource-info .resource-info-text .resource-name {
```
Golodhros, line 23: Would this work with just '.resource-name'?

allisonsuarez: it doesn't I tried to use a less specific one and the purple styling kept coming through

Golodhros (approved):
LGTM

PR #1920, opened 2022-07-02
fix: open file in read mode and close on cleanup for ElasticsearchPublisher

m1racoli:
* fix: open file in read mode
* fix: close file handler on cleanup

### Summary of Changes

Since we only read from the file, we do not need to open it in write mode. Furthermore we close the file on cleanup in order to avoid issues with open file handlers. In particular this affects the automatic cleanup of [temporary directories](https://docs.python.org/3.10/library/tempfile.html#tempfile.TemporaryDirectory) containing the file.

### Tests

This affects low level aspect of how we open and close files, which might be beyond the scope of tests.

### Documentation

Nothing to document as there is not effective change in behaviour.

### CheckList
Make sure you have checked **all** steps below to ensure a timely review.
- [x] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).
- [x] PR includes a summary of changes.
- [x] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [x] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public functions and the classes in the PR contain docstrings that explain what it does

PR #2023, opened 2022-11-07
fix: open file in read mode and close on cleanup for ElasticsearchPublisher

m1racoli:
* fix: open file in read mode
* fix: close file handler on cleanup

### Summary of Changes

Since we only read from the file, we do not need to open it in write mode. Furthermore we close the file on cleanup in order to avoid issues with open file handlers. In particular this affects the automatic cleanup of [temporary directories](https://docs.python.org/3.10/library/tempfile.html#tempfile.TemporaryDirectory) containing the file.

Continuation of stale PR #1920 

### Tests

This affects low level aspect of how we open and close files, which might be beyond the scope of tests.

### Documentation

Nothing to document as there is not effective change in behaviour.

### CheckList


kristenarmes (approved):

Golodhros:
@m1racoli not sure why the checks are not running, could you try pushing the branch again?

PR #2047, opened 2022-12-07
feat -- configurable homepage widgets

B-T-D:
### Summary of Changes

Refactors existing features on the home page to be widgets, and adds ability for Amundsen instances to customize their widgets.

### Tests

Modified some home page tests to reflect refactor.

### CheckList


frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -27,13 +27,57 @@ export interface DispatchFromProps {
   searchReset: () => UpdateSearchStateReset;
 }
 
+export type Widget = {
+  // TODO: What file does the Widget type belong in?
```
Golodhros, line 31: I would create an `interfaces/Widgets.ts` file or similar.

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -27,13 +27,57 @@ export interface DispatchFromProps {
...
+  name: string;
+  options: {
+    path: '/';
```
Golodhros, line 34:
```suggestion
    path: string;
```

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -27,13 +27,57 @@ export interface DispatchFromProps {
...
     this.props.searchReset();
   }
 
+  getHomePageWidgets(layout: HomePageLayout): React.ReactNode[] {
+    /* TODO: Make 100% sure it returns the pre-existing layout absent a custom config,
+    to not break any OSS users' Amundsen implementations */
+
+    return [
+      <div className="home-element-container">
+        <SearchBar />
```
Golodhros, line 53: These need to be updated to the Widgets, right?

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -27,13 +27,57 @@ export interface DispatchFromProps {
...
+      </div>,
+      <div className="home-element-container">
+        <PopularTables />
+      </div>,
+    ];
+  }
+
+  defaultLayout: HomePageLayout = [];
+
+  homePageWidgets = this.getHomePageWidgets(this.defaultLayout);
```
Golodhros, line 79: We don't really want to use instance variables, even if we are using classes. 
We'll be moving out of classes as soon as possible, so the less we depend on them, the better.

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -47,28 +91,7 @@ export class HomePage extends React.Component<HomePageProps> {
...
-            <div className="home-element-container">
-              <TagsListContainer shortTagsList />
-            </div>
-            <div className="home-element-container">
-              <MyBookmarks />
-            </div>
-            <div className="home-element-container">
-              <PopularTables />
-            </div>
+            {this.homePageWidgets}
```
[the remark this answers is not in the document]
B-T-D: What do you mean by having new components rendered? A new HomePageWidgetContainer (or whatever name) component that holds all the widgets that the user wants on their homepage? That seems logical to me.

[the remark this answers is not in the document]
B-T-D: So it could be something like:

```
  render() {
    /* TODO, just display either popular or curated tags,
    do we want the title to change based on which
    implementation is being used? probably not */
    return (
      <main className="container home-page">
        <div className="row">
          <div
            className={`col-xs-12 ${
              announcementsEnabled() ? 'col-md-8' : 'col-md-offset-1 col-md-10'
            }`}
          >
            <h1 className="sr-only">{HOMEPAGE_TITLE}</h1>
            ****<HomePageWidgets
               getHomePageWidgets={getHomePageWidgets} 
             />****
          </div>
          {announcementsEnabled() && (
            <div className="col-xs-12 col-md-offset-1 col-md-3">
              <Announcements />
            </div>
          )}
        </div>
      </main>
    );
  }
}
```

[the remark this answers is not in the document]
Golodhros: Yes, but you don't need to extract it into a different file for the moment, you can write it above line 40 as a functional component.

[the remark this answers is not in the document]
B-T-D: Thanks, will do!

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -27,13 +27,57 @@ export interface DispatchFromProps {
...
+      <div className="home-element-container">
+        <SearchBar />
+      </div>,
+      <div className="filter-breadcrumb pull-right">
+        <Breadcrumb
+          direction="right"
+          path="/search"
+          text={SEARCH_BREADCRUMB_TEXT}
+        />
+      </div>,
```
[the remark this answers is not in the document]
B-T-D: I agree conceptually but would like to leave that change until a little later if you don't mind--it'll require some tweaking beyond just moving the Breadcrumb to be inside a different render, since elsewhere in the UI we render the search bar without the "Advanced Search" Breadcrumb, for example on the advanced search page itself ('/search' URL):

Screenshot 2022-12-08 at 7 27 18 PM

(screenshot from our prod)

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
B-T-D: Yes, I've now updated the component names

frontend/amundsen_application/static/js/config/config-utils.ts
```diff
@@ -569,3 +570,10 @@ export function searchHighlightingEnabled(resource: ResourceType): boolean {
 export function getSearchResultsPerPage(): number {
   return AppConfig.searchPagination.resultsPerPage;
 }
+
+/**
+ * Returns the homepage widgets configuration
+ */
+export function getHomePageWidgets(): HomePageWidgetsConfig {
```
Golodhros, line 577: We'll need a test for this new method

frontend/amundsen_application/static/js/features/HomePageWidgets/BadgesWidget/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import * as React from 'react';
+
+import BadgesList, { BadgesListContainerProps } from 'features/Badges';
+
+type BadgesWidgetProps = BadgesListContainerProps;
+
+export const BadgesWidget: React.FC<BadgesWidgetProps> = (
+  props: BadgesWidgetProps
+) => (
+  <div>
```
Golodhros, line 13: Are these divs necessary?

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -29,6 +24,43 @@ export interface DispatchFromProps {
...
+    res.push(
+      <div className="home-element-container">
+        <React.Suspense fallback={<div>Loading...</div>}>
+          <WidgetComponent {...additionalProps} />
+        </React.Suspense>
+      </div>
+    );
+  });
+
+  return res;
```
Golodhros, line 54: I would do just a map here.
.forEach and .push are methods that we barely use in nowadays' JavaScript, as they imply mutations of data. We prefer to create new data with .map

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -29,6 +24,43 @@ export interface DispatchFromProps {
...
+        import('/js/features/HomePageWidgets/' + widget.options.path + '.tsx')
+    );
+
+    const additionalProps = widget.options.additionalProps
+      ? widget.options.additionalProps
+      : null;
+
+    res.push(
+      <div className="home-element-container">
+        <React.Suspense fallback={<div>Loading...</div>}>
```
Golodhros, line 47: We'll probably need to create a more elaborate loading state here. How does it show now?

We should ask Knowl for something here

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -29,6 +24,43 @@ export interface DispatchFromProps {
...
+          <WidgetComponent {...additionalProps} />
+        </React.Suspense>
+      </div>
+    );
+  });
+
+  return res;
+};
+
+export const HomePageWidgets = (props) => {
```
Golodhros, line 57:
```suggestion
export const HomePageWidgets = ({homePageLayout}) => {
```

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
```diff
@@ -29,6 +24,43 @@ export interface DispatchFromProps {
...
+  const { homePageLayout } = props;
+  const widgets = getHomePageWidgetComponents(homePageLayout);
+
+  return <div>{widgets}</div>;
```
Golodhros, line 61:
```suggestion
  return <div>{getHomePageWidgetComponents(homePageLayout)}</div>;
```

frontend/amundsen_application/static/js/features/HomePageWidgets/BadgesWidget/index.tsx
B-T-D: Nope, only the SearchBarWidget one I think

frontend/amundsen_application/static/js/pages/HomePage/index.tsx
B-T-D: Right now, I never see it actually do the fallback. Some of the components like TagsListWidget have their own more elaborate loading states (the tags list has a blur effect) which still work inside the React.Suspense.

Golodhros (approved):
LGTM

PR #2069, opened 2022-12-30
feat: Change default table lineage depth to 1

jsnb-devoted:
### Summary of Changes
Changing the default lineage depth from 5 to 1. Not sure if this is appealing to anyone but I thought I would put this out there. Typically people don't need to see the lineage at this broad a depth. Ideally this parameter would be selectable in the UI but until then defaulting it to 1 makes the page more usable for me. If other folks disagree we can just close this.

### CheckList


Golodhros:
Hi @jsnb-devoted,
I chatted with the team, and it seems there is a need here to expand our configuration object to include the default depth of the graph as you suggest.

The changes would be:
- Update the config-types.ts with an extra parameter for the default depth, maybe something like:
inAppPageConfig: {
   defaultLineageDepth: number,
}
- Give it a default of 5 on the config-default.ts
- Create a function (and test it) on config-utils.ts
- Use that function to set the default depth
- Remove the default depth on the Action (https://github.com/amundsen-io/amundsen/blob/cf7bf157bc8c85465b2ea96ceef6943a24cf2d03/frontend/amundsen_application/static/js/ducks/lineage/reducer.ts#L33)

I know is it a bit more work, would you be OK to do this?

jsnb-devoted:
> I know is it a bit more work, would you be OK to do this?

hi @Golodhros - That seems straightforward enough. Thanks for outlining all the places this would need to change.

Sounds like we want the depth to be configurable regardless but another thing that I doubt I have the bandwidth or front end experience for is defaulting all of the nodes one level of depth away from the root node to be collapsed. That way the user loads the page and it appears as if the graph is only one level of depth but they can click the nodes to expand them. I took a look at the code for the graph and it seemed a bit more advanced than this config change.

Golodhros:
> > I know is it a bit more work, would you be OK to do this?
> 
> hi @Golodhros - That seems straightforward enough. Thanks for outlining all the places this would need to change.
> 
> Sounds like we want the depth to be configurable regardless but another thing that I doubt I have the bandwidth or front end experience for is defaulting all of the nodes one level of depth away from the root node to be collapsed. That way the user loads the page and it appears as if the graph is only one level of depth but they can click the nodes to expand them. I took a look at the code for the graph and it seemed a bit more advanced than this config change.

I see. Sadly I won't be able to help on this, as I am not super familiar with the graph code and we don't use it at Lyft, so my priorities are not aligned with improving that code.

Maybe you could do a git blame and reach out the authors?

jsnb-devoted:
> > > I know is it a bit more work, would you be OK to do this?
> > 
> > 
> > hi @Golodhros - That seems straightforward enough. Thanks for outlining all the places this would need to change.
> > Sounds like we want the depth to be configurable regardless but another thing that I doubt I have the bandwidth or front end experience for is defaulting all of the nodes one level of depth away from the root node to be collapsed. That way the user loads the page and it appears as if the graph is only one level of depth but they can click the nodes to expand them. I took a look at the code for the graph and it seemed a bit more advanced than this config change.
> 
> I see. Sadly I won't be able to help on this, as I am not super familiar with the graph code and we don't use it at Lyft, so my priorities are not aligned with improving that code.
> 
> Maybe you could do a git blame and reach out the authors?

@ozandogrultan or @bbsbb any interest in this feature? Or maybe can describe the path to implementing?
The tl;dr is default all of the level 1 nodes in the table lineage view to be collapsed instead of expanded.

Golodhros (approved):
LGTM

jsnb-devoted:
I'm botching signing the old commits and I just accidentally pulled in another commit. I'm figuring this out.

jsnb-devoted:
Fixed in a follow up PR https://github.com/amundsen-io/amundsen/pull/2075

PR #2079, opened 2023-01-09
fix: int is too small for a unix timestamp

allisonsuarez:
Changed dashboard summary model in comment to use `last_successful_run_timestamp: Optional[float] = None` since a value like `1668725028.424` is too large to be casted to an `int`


Golodhros (approved):
LGTM

kristenarmes (approved):

PR #2151, opened 2023-05-03
style – Adjust margin on elements inside the Alert component

B-T-D:
## Description

The vertical alignment was inconsistent between different elements inside the Alert component. This PR adjusts the margins to make the Alert look better and properly aligned.

How the Alert component looked before:

before

After:

after

## How Has This Been Tested?

Manually. Building locally with the change and visually inspecting. 

### Documentation

None

### CheckList
* [ ] Adds tests

### DIFF
--- a/frontend/amundsen_application/static/js/components/Alert/styles.scss
+++ b/frontend/amundsen_application/static/js/components/Alert/styles.scss
@@ -7,7 +7,6 @@
 $alert-border-radius: 4px;
 $alert-warning-background: #fff0d4;
 $alert-alert-background: #ffe4dd;
-$alert-element-margin: auto 0 auto auto;
 
 .alert-list {
   margin-top: $spacer-3;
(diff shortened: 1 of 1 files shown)


frontend/amundsen_application/static/js/components/Alert/styles.scss
```diff
@@ -5,9 +5,9 @@
 @import 'typography';
 
 $alert-border-radius: 4px;
-$alert-message-line-height: 24px;
 $alert-warning-background: #fff0d4;
 $alert-alert-background: #ffe4dd;
+$alert-element-margin: auto 0 auto auto;
```
kristenarmes, line 10: the top and bottom margins look good in the screenshot now, but the message is a little far over from the icon IMO. maybe we can have two separate margin values here, this one you have used for the alert-action, then `auto auto auto 0` used for the alert-message so that we can have 0 on the left side like before

kristenarmes (approved):

PR #2152, opened 2023-05-15
style - Truncate long notice modal text with ellipsis

B-T-D:
## Description

When the text of an item in the notice modal is too long for its container, it looks better for it to truncate with an ellipsis rather than wrapping.

## How Has This Been Tested?

Manually.

### CheckList
* [ ] Adds tests

### DIFF
--- a/frontend/amundsen_application/static/js/components/DefinitionList/index.tsx
+++ b/frontend/amundsen_application/static/js/components/DefinitionList/index.tsx
@@ -37,7 +37,11 @@ export const DefinitionList: React.FC<DefinitionListProps> = ({
           {typeof description === 'string' ? (
             <SanitizedHTML
               html={description}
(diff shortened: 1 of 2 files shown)


frontend/amundsen_application/static/js/components/DefinitionList/index.tsx
```diff
@@ -35,7 +35,14 @@ export const DefinitionList: React.FC<DefinitionListProps> = ({
         </dt>
         <dd className="definition-list-definition">
           {typeof description === 'string' ? (
-            <SanitizedHTML html={description} />
+            <SanitizedHTML
```
allisonsuarez, line 38: we don't usually do inline styling, and it seems you added these to a class on css as well so can you use that instead?

allisonsuarez (approved):

PR #2162, opened 2023-06-08
fix: UnboundLocalError for msg in create_search_response

kristenarmes:
## Description
Initialize the `msg` and `status_code` variables at the top of the `create_search_response` util function for the search proxy

## Motivation and Context
We were getting this exception: `UnboundLocalError: local variable 'msg' referenced before assignment`
To mitigate this, now initialize the variables that were previously only set within the for loop at the top of the function.

### CheckList
* [ ] Adds tests


allisonsuarez (approved):
thank you!

Issue #2148, opened 2023-04-26
GPL-licensed dependency `unidecode`


hehao98:
(I think this is not a bug or a feature request, so I am opening an issue without a template)

Hello,

We are a group of researchers developing tools to monitor and remediate open-source license incompatibilities in the PyPI ecosystem.

We find that your `amundsen-databuilder` package has a **GPL-2.0-or-later** licensed dependency `unidecode`. This could be problematic because GPL-licensed software requires any of its derivative work to be also licensed under GPL, but `amundsen-databuilder` is licensed under Apache 2.0. 

To remove this license incompatibility, the following possible remediations may be considered:

* Relicense `amundsen-databuilder` under GPL 3.0 or AGPL 3.0
* Migrate `unidecode` to `text-unidecode`
* Change `unidecode` version to 0.04.6
* Remove `unidecode`

Note that the above remediations are generated by an automated tool that is still under test, may be incorrect, and does not represent legal advice. We welcome any suggestions and feedback!

kristenarmes:
Hi @hehao98, thanks for bringing this to our attention, I have opened a PR to address this issue

PR #2168, opened 2023-06-14
chore: Use text-unidecode instead of unidecode package to fix license incompatibility

kristenarmes:
## Description
- Update databuilder to use `text-unidecode` instead of `unidecode` due to license incompatibility issues

- Unrelated to the main change, now also ignoring the `None` type passed to `pyparsing`'s `nestedExpr` function. In the docstring for this function, it specifically says `if no expressions are to be ignored, then pass ``None`` for this argument.` so the type definition for the function is incorrect and this is a workaround

## Motivation and Context
This fixes license incompatibility issues, since Amundsen is licensed under Apache 2.0 and `unidecode` is GPL. See this issue for further context: https://github.com/amundsen-io/amundsen/issues/2148

## How Has This Been Tested?
Python tests pass

### Documentation
N/A

### CheckList
* [ ] Adds tests

### DIFF
--- a/databuilder/requirements.txt
+++ b/databuilder/requirements.txt
@@ -20,7 +20,7 @@ statsd>=3.2.1
 retrying>=1.3.3
 unicodecsv>=0.14.1,<1.0
 httplib2>=0.18.0
-text-unidecode
+text-unidecode>=1.3
 Jinja2>=2.10.0,<4
 pandas>=0.21.0,<1.5.0
 responses>=0.10.6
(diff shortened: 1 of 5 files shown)


databuilder/requirements.txt
```diff
@@ -20,7 +20,7 @@ statsd>=3.2.1
 retrying>=1.3.3
 unicodecsv>=0.14.1,<1.0
 httplib2>=0.18.0
-unidecode
+text-unidecode
```
ozandogrultan, line 23: Should we pin a version here to prevent backwards incompatibility issues?

MrwanBaghdad (approved):

ozandogrultan (commented):
Should we bump up the version of databuilder in setup.py?

ozandogrultan (approved):
LGTM!

PR #2182, opened 2023-08-08
Refactor Neo4jProxy table owners query for easier customization

B-T-D:
## Description

For some users of Amundsen, it may prove helpful to customize logic for deciding who a table's "owners" are. This can be accomplished by breaking up the table metadata API's query to make a modular query for only the owners (as is already done when getting table readers, in `Neo4jProxy._exec_usage_query`).

The results of the new `Neo4jProxy._exec_owners_query` are then combined back into the return value of the `Neo4jProxy.get_table method`, preventing any disruption to the JSON shape expected by the frontend. 

## Motivation and Context
This solves the problem of companies that use Neo4j-based Amundsen being unable to customize what data is returned as a table's owners. For example if a company wanted to apply a custom sorting logic to multiple owners, or impute table ownership based on specialized business logic. 

## How Has This Been Tested?

I updated existing unit tests such that they still pass. 

I also tested the service running locally and called it via an Amundsen frontend UI that was running locally. 

### Documentation

None

### CheckList
* [ ] Adds tests

### DIFF
--- a/metadata/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata/metadata_service/proxy/neo4j_proxy.py
@@ -159,10 +159,7 @@ class Neo4jProxy(BaseProxy):
         cols, last_neo4j_record = self._exec_col_query(table_uri)
 
         readers = self._exec_usage_query(table_uri)
-        LOGGER.info(f'about to call _exec_owners_query(). {table_uri=}')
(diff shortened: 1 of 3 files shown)


kristenarmes (approved):
looks good, thanks! just one thing, can you bump the version in setup.py so we can release a new metadata version?

PR #1817, opened 2022-04-15
feat: Introduce Nebula Metadata Proxy and Databuilder

wey-gu:
Metadata: Nebula Proxy
Databuilder:
- Nebula Extractor
- Nebula Search Data Extractor
- Nebula CSV Loader
- Nebula CSV Publisher
- Nebula Serializer
- Nebula Sample Data Loader

### Summary of Changes

see #1816

### Tests

All New things were UT covered.

### Documentation

```bash
docker-compose -f docker-amundsen-nebula.yml build
docker-compose -f docker-amundsen-nebula.yml up -d

cd databuilder
python3 -m venv venv
source venv/bin/activate
pip3 install --upgrade pip
pip3 install -r requirements.txt
python3 setup.py install
python3 example/scripts/sample_data_loader_nebula.py # this is necessary to trigger schema creation
```

### CheckList
Make sure you have checked **all** steps below to ensure a timely review.
- [x] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).
- [x] PR includes a summary of changes.
- [x] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [x] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public functions and the classes in the PR contain docstrings that explain what it does

PR #2193, opened 2023-09-21
Fix - Handle ES missing index edge case (bug fix)

mikaalanwar:
## Description
This pull request handles an edge case with basic search. If more than one `resource_types` is configured for amundsen (e.g Tables and Dashboards) and one of them have missing indices (i.e. no data got ingested for either of them in an initial run), it will crash the basic search from working for all `resource_types`. In a clean deployment (blank neo4j and elasticsearch), it is possible to not have data for all configured resource types at once. In such a case the search should not return any results for the resource_type with missing data instead of crashing basic search altogether. This PR addresses this problem and ensures that we trigger an elasticsearch "search" against only those `resource_types` which have a corresponding index/alias present in the elasticsearch database.

## Motivation and Context
For a fresh deployment of amundsen, in case any of the configured resource types is missing indices in elasticsearch, it will prevent the search from running altogether. Also, the error statement tends to be a bit confusing: 

`[ERROR] es_proxy_v2_1.execute_multisearch_query:349 (1:Thread-48) - Failed to execute ES search queries. TransportError(N/A, 'index_not_found_exception')`

As it does not mention which particular index (or alias) was missing. In my case, dashboards were missing but I was under the impression that maybe the search service was not looking in the right place because there were some indices in my elasticsearch instance already, which were searchable using other debugging tools (such as [elasticvue](https://elasticvue.com/)). I came across the aforementioned issue and wanted to push a fix for this. There wasn't a corresponding open issue against this already, but there is a reference to a similar issue [here](https://github.com/amundsen-io/amundsen/pull/1817#issuecomment-1126889050).

## How Has This Been Tested?
I prepared a clean environment locally by deleting all neo4j and elasticsearch data. Next, I ran an ingestion run for only one resource type (without knowing that I'd have to run it for all `resource_types` to actually get it working). In my case, I ingested tables only, without ingesting dashboards. And then I noticed that the advanced search (sidebar) was working but the main, basic search bar wasn't. Upon investigation, I narrowed it down and implemented this fix. Now, the main/basic search works fine even if I don't ingest dashboards. 

### Documentation
No updates to the documentation were needed. 

### CheckList
* [ ] Adds tests

### DIFF
--- a/search/search_service/proxy/es_proxy_v2_1.py
+++ b/search/search_service/proxy/es_proxy_v2_1.py
@@ -363,7 +363,7 @@ class ElasticsearchProxyV2_1(ElasticsearchProxyV2):
         multisearch = MultiSearch(using=self.elasticsearch)
 
         for resource in resource_types:
-            # guard clause to prevent search in missing indices 
+            # guard clause to prevent search in missing indices
             aliases_in_es = self.elasticsearch.indices.get_alias().keys()
             if self.get_index_alias_for_resource(resource_type=resource) not in aliases_in_es:
                 continue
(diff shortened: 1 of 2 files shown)


ozandogrultan (commented):
Please bump up the version in setup.py

ozandogrultan (approved):

PR #2216, opened 2023-11-28
fixed where

amarshinde150:
fixing where clause problem in redshift_metadata_extractor.py

### CheckList
* [ ] PR title addresses the issue accurately and concisely
* [ ] Adds tests

### DIFF
--- a/databuilder/databuilder/extractor/redshift_metadata_extractor.py
+++ b/databuilder/databuilder/extractor/redshift_metadata_extractor.py
@@ -29,7 +29,7 @@ class RedshiftMetadataExtractor(BasePostgresMetadataExtractor):
             cluster_source = f"'{self._cluster}'"
 
         if where_clause_suffix:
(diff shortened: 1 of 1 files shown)


databuilder/databuilder/extractor/redshift_metadata_extractor.py
```diff
@@ -29,7 +29,7 @@ def get_sql_statement(self, use_catalog_as_cluster_name: bool, where_clause_suff
             cluster_source = f"'{self._cluster}'"
 
         if where_clause_suffix:
-            if where_clause_suffix.lower().startswith("where"):
+            if "where" in where_clause_suffix.lower():
```
kristenarmes, line 32: is this change in case there is leading whitespace? I'd recommend instead using `lstrip()` if so, and then you can still check that the string starts with `where`

Alagappan:
This looks like a duplicate of https://github.com/amundsen-io/amundsen/pull/2217. Closing this one.

PR #2217, opened 2023-11-30
fixing where clause issue in redshift_metadata_extractor

amrshindhe:
Leading whitespaces causing errors with startsWith() python method, so using lstrip() before starts with will resolve the error


Alagappan:
ping @amarshinde150 any updates on this one?

PR #2226, opened 2024-01-05
chore: Remove py37 support, and unify compatibility matrix from 3.8-3.9

kristenarmes:
## Description
Remove python 3.7 support since it has reached its EOL, and unify the compatibility matrix across services to include 3.8 and 3.9

## Motivation and Context
We are going to be upgrading our services to py3.10 soon, so this is getting the project set up for that. py3.10 won't be added in this PR yet because it appears to require other package upgrades, so keeping the focus on removing py3.7 here. We will need to upgrade flask and werkzeug with our py3.10 upgrade, and werkzeug 3.0.1 does not support py3.7 so that is the main motivation.

### Documentation
Updated all relevant docs that show version requirements

### CheckList
* [X] Updates Documentation and Docstrings
* [ ] Adds tests

### DIFF
--- a/.github/workflows/common_pull_request.yml
+++ b/.github/workflows/common_pull_request.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python-version: ["3.7.x"]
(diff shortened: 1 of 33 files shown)


Golodhros (approved):
LGTM

PR #2230, opened 2024-02-12
feat: Aggregating Alerts Feature

jdslaugh:
## Description
Aggregates alert notices that are passed into the AlertList component and then parses them in the DefinitionList.

## Motivation and Context
The UI is cluttered when there are numerous alerts taking up space, especially when many of them have the same title. Aggregating them into one modal alleviates this and reduces the visual strain on the user.

#### Notices
Screenshot 2024-02-09 at 3 20 46 PM

#### Modal
Screenshot 2024-02-09 at 3 20 51 PM

#### Info SVG Icon Alignment
Screenshot 2024-02-23 at 12 51 57 PM

#### Notices
Screenshot 2024-02-23 at 12 56 24 PM

#### Modal
Screenshot 2024-02-09 at 3 18 18 PM

#### Storybook
Screenshot 2024-02-09 at 3 41 36 PM

#### Info SVG Icon Alignment
Screenshot 2024-02-23 at 12 56 24 PM

## How Has This Been Tested?
- Utilizing mock data to recreate the notices that can be aggregated as well as without aggregation.
- Unit testing through already existing tests in place
- Verification and updates to storybook and related components

### CheckList
* [x] Adds tests

### DIFF
--- a/frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
+++ b/frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
@@ -10,19 +10,46 @@ export interface AlertListProps {
   notices: NoticeType[];
 }
 
+export interface AggregatedAlertListProps {
+  notices: {
+    [key: string]: NoticeType;
+  };
+}
+
+const aggregateNotices = (notices) =>
+  notices.reduce((accum, notice: any) => {
+    if (notice) {
+      const { messageHtml, severity, payload } = notice;
+
+      if (payload) {
+        accum[messageHtml] ??= {
+          severity,
+          payload: { descriptions: [] },
+        };
+        accum[messageHtml].payload.descriptions.push(payload);
+      } else {
+        accum[messageHtml] = { ...notice };
+      }
+    }
+
+    return accum;
+  }, {});
+
 export const AlertList: React.FC<AlertListProps> = ({ notices }) => {
   if (!notices.length) {
     return null;
   }
 
+  const aggregated = aggregateNotices(notices);
+
   return (
     <div className="alert-list">
-      {notices.map((notice, idx) => (
+      {Object.keys(aggregated).map((notice, idx) => (
         <Alert
           key={idx}
-          message={notice.messageHtml}
-          severity={notice.severity}
-          payload={notice.payload}
+          message={notice}
+          severity={aggregated[notice].severity}
+          payload={aggregated[notice].payload}
         />
       ))}
     </div>
--- a/frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
+++ b/frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
@@ -3,7 +3,7 @@
 
 import * as React from 'react';
 
-import { NoticeType } from 'config/config-types';
+import { NoticeSeverity, NoticeType } from 'config/config-types';
 import { Alert } from './Alert';
 
 export interface AlertListProps {
@@ -17,21 +17,23 @@ export interface AggregatedAlertListProps {
 }
 
 const aggregateNotices = (notices) =>
-  notices.reduce((accum, notice: any) => {
+  notices.reduce((accum, notice: NoticeType) => {
     if (notice) {
       const { messageHtml, severity, payload } = notice;
 
-      /* eslint-disable no-param-reassign */
-      if (payload) {
-        accum[messageHtml] ??= {
-          severity,
-          payload: { descriptions: [] },
-        };
-        accum[messageHtml].payload.descriptions.push(payload);
-      } else {
-        accum[messageHtml] = { ...notice };
+      if (typeof messageHtml !== 'function') {
+        if (payload) {
+          accum[messageHtml] ??= {};
+          accum[messageHtml][severity] ??= {
+            payload: { descriptions: [] },
+          };
+          accum[messageHtml][severity].payload.descriptions.push(payload);
+        } else {
+          accum[messageHtml] = {
+            [severity]: { ...notice },
+          };
+        }
       }
-      /* eslint-enable no-param-reassign */
     }
 
     return accum;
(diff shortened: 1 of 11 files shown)


jdslaugh:
Opened in OSS here: https://github.com/amundsen-io/amundsen/pull/2230

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
```diff
@@ -10,19 +10,46 @@ export interface AlertListProps {
...
+  };
+}
+
+const aggregateNotices = (notices) =>
+  notices.reduce((accum, notice: any) => {
+    if (notice) {
+      const { messageHtml, severity, payload } = notice;
+
+      if (payload) {
+        accum[messageHtml] ??= {
```
kristenarmes, line 25: I just noticed this in your screenshot in the PR description where it groups the notices with the same message but different severities into 1 - I'm thinking if the severity is different we should group those separately if possible, so by unique message and severity? I don't think that would happen for us in our cases, but not sure if it might for others who use this feature

kristenarmes (commented):
looks good overall! just had one comment that I realized based on your screenshot of the grouped notices

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
jdslaugh: Ah this is a great point. Yeah I can group them by message + severity, let me get a fix up. Thanks for the find!

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
```diff
@@ -10,19 +10,48 @@ export interface AlertListProps {
...
+    [key: string]: NoticeType;
+  };
+}
+
+const aggregateNotices = (notices) =>
+  notices.reduce((accum, notice: any) => {
+    if (notice) {
+      const { messageHtml, severity, payload } = notice;
+
+      /* eslint-disable no-param-reassign */
```
Golodhros, line 24: We could get rid of this rule, we have better things to care about!

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
```diff
@@ -10,19 +10,48 @@ export interface AlertListProps {
...
 }
 
+export interface AggregatedAlertListProps {
+  notices: {
+    [key: string]: NoticeType;
+  };
+}
+
+const aggregateNotices = (notices) =>
+  notices.reduce((accum, notice: any) => {
```
Golodhros, line 20: Any way to avoid this 'any'?

Golodhros (approved):
LGTM

Golodhros:
Maybe I would add a count on the aggregated notices, something like a (X) at the end of the text.

jdslaugh:
> Maybe I would add a count on the aggregated notices, something like a (X) at the end of the text.

Ooh that's a good idea, yeah I can do that!

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
jdslaugh: Yeah should be able to, might've just been leftover from my first implementation :)

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
jdslaugh: Sounds good!

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
jdslaugh: Alright, took a few days to get back to this but the changes are now up and screenshots are updated. So it will now group by message and then sub-group by severity, which is then sorted to show up

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
jdslaugh: Adjusted the rule to turn it off

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
jdslaugh: Mapped it to `NoticeType`

frontend/amundsen_application/static/js/components/Alert/AlertList.tsx
```diff
@@ -3,28 +3,67 @@
...
+    [key: string]: NoticeType;
+  };
+}
+
+const aggregateNotices = (notices) =>
+  notices.reduce((accum, notice: NoticeType) => {
+    if (notice) {
+      const { messageHtml, severity, payload } = notice;
+
+      if (typeof messageHtml !== 'function') {
```
jdslaugh, line 24: Only case I'm not sure of is when `messageHtml` is a function and takes a `resourceName`. Not really sure what I would use for that case but if anyone has any ideas I can do it or if we assume it's not happening for this we can just ignore it I guess.

kristenarmes: good question, I just checked and we have one instance where we were using this functionality at Lyft. I looked into this a bit and I don't think it would ever be a function type at this point. I included my investigation below, but TLDR I think this is fine to leave as you have it here

in the table detail page it passes the results from this function [aggregateResourceNotices](https://github.com/amundsen-io/amundsen/blob/890bfaf92c4b01ff524791b0447f2f57fee1b0a1/frontend/amundsen_application/static/js/pages/TableDetailPage/index.tsx#L122) to the AlertList, and that function calls [getResourceNotices](https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/static/js/config/config-utils.ts#L114). in there it transforms the notices that have function types as their messageHtml to call the function and set set the field to its string message. I investigated backward from the original implementation of function messageHtml fields [here](https://github.com/amundsen-io/amundsen/pull/1280/files)

kristenarmes (approved):
looks great, thanks for all the updates!

I see there is a failure from betterer as the last thing, you can run `npm run betterer` under the directory `frontend/amundsen_application/static` to test locally. I'm not sure what the best solution is for this case bc I think this is what you tried to turn off...if it is too much of a pain to fix, I'm fine with you using the comment like you had before to bypass

PR #2234, opened 2024-02-26
fix: Update mssql extractor to utilize proper case for case sensitive MSSQL databases

ardcarmvk:
Correct SQL code's table and column case so mssql extractor works with case sensitive instances

## Description
Some SQL Server databases are configured as case sensitive.  Changes were made for utilizing proper case in 
required areas.

## Motivation and Context
Extractor will not work on mssql databases that are case sensitive

## How Has This Been Tested?

Ran the SQL code on SQL database with case sensitivity, tested the extractor against non-case sensitive and case sensitive
databases.  Ran on windows 10 with python 3.12 on local machine.

### CheckList
* [ x ] PR title addresses the issue accurately and concisely
* [ x ] Updates Documentation and Docstrings
* [ ] Adds tests

### DIFF
--- a/databuilder/databuilder/extractor/mssql_metadata_extractor.py
+++ b/databuilder/databuilder/extractor/mssql_metadata_extractor.py
@@ -31,29 +31,29 @@ class MSSQLMetadataExtractor(Extractor):
             {cluster_source} AS cluster,
             TBL.TABLE_SCHEMA AS [schema_name],
             TBL.TABLE_NAME AS [name],
-            CAST(PROP.VALUE AS NVARCHAR(MAX)) AS [description],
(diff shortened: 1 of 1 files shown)


Golodhros:
Please check the DCO task above

ardcarmvk:
DCO task should be resolved now, just need review.  Thanks

kristenarmes (approved):

kristenarmes:
> @ardcarmvk it seems it got broken again

I think the issue is the original commit has to be modified to include the sign off, check out the details in the failed task [here](https://github.com/amundsen-io/amundsen/pull/2234/checks?check_run_id=22035077518) for instructions

PR #2242, opened 2024-03-27
chore: Add py3.10 to compatibility matrix across all services except for metadata

kristenarmes:
## Description
Adding py3.10 to compatibility matrix across all services except for metadata.
Metadata is blocked by amundsengremlin:
- `overrides` package needs to be upgraded, which is doable
- `gremlinpython` package needs to be upgraded, but there are some extra changes required for this when you go >=3.5.0 since it stops using the `tornado` package, and there are some restrictions with Neptune ([see some details here](https://docs.aws.amazon.com/neptune/latest/userguide/access-graph-gremlin-python.html)). Since I don't use Neptune and have a good way of testing this to make sure it works, I am going to leave metadata out for now.

## Motivation and Context
We are going to be upgrading our services to py3.10, so want to make sure everything will work as expected

## How Has This Been Tested?
Tests will run on py3.10 as part of this PR

### CheckList
* [ ] Adds tests

### DIFF
--- a/.github/workflows/common_pull_request.yml
+++ b/.github/workflows/common_pull_request.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python-version: ['3.8.x', '3.9.x']
+        python-version: ['3.8.x', '3.9.x', '3.10.x']
     steps:
       - name: Checkout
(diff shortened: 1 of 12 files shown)


feng-tao (approved):

Alagappan:
Looks good! Please fix broken tests before merging in

PR #2258, opened 2024-08-14
feat: Add config for implicit transactions to neo4j extractor

kristenarmes:
## Motivation and Context
- Adding the ability to configure the neo4j extractor to use [implicit transactions](https://neo4j.com/docs/python-manual/current/query-advanced/#implicit-transactions). The default behavior of the extractor will remain the same.
- Most cases can continue to use the default explicit transactions, but implicit transactions are required for `CALL {} IN TRANSACTIONS` [subqueries](https://neo4j.com/docs/cypher-manual/current/subqueries/subqueries-in-transactions/). Using this can help improve performance for queries that use too much memory, for example when a `Neo.ClientError.General.TransactionOutOfMemoryError` exception is thrown this can help break up the query.
- Also adding `default_access_mode=neo4j.READ_ACCESS` to the session configuration for all queries for better performance.

## How Has This Been Tested?
Ran this in an ETL job that set the config to use implicit transactions for one, and another job that used the default explicit transactions with no config passed and confirmed expected behavior.

### Documentation
N/A

### CheckList
* [ ] Adds tests


B-T-D (approved):
LGTM

PR #2260, opened 2024-08-14
feat -- table owner categories

B-T-D:
## Description

This PR adds config support for differentiating multiple categories of table owners in the Amundsen UI. For example, if an Amundsen instance wants to differentiate human users who are responsible for remediating operational issues with a table from non-human team or service emails.

To implement this, we propose supporting open-ended owner categories identified by a name and a definition text, to be configured in non-open-source code. The relevant React components are updated to render differently when owner categories are provided, while preserving current behavior for Amundsen instances that don't want to display different owner types differently in the UI.

- Add OwnerCategory interface
- Update User interface to support passing `other_key_values` field that already exists on the backend's User model
- Update OwnerEditor component to render table owners grouped by category, when categories are configured

Example of owners rendered by category (where there's no owner of a configured category for the table, we still show the category name and definition, e.g. no "operational" owners here):
owner categories and definition info text

## Motivation and Context

At Lyft, we'd like to be able to differentiate human-user table owners--who could actually be contacted to address data quality problems--from non-human owners.

## How Has This Been Tested?

Existing unit tests and linting pass. I locally tested using the new configs to have the React app render owners by category.

### CheckList
* [ ] Adds tests


Golodhros (approved):
LGTM

frontend/amundsen_application/static/js/interfaces/User.ts
```diff
@@ -6,6 +6,7 @@ export interface User {
   email: string;
   profile_url: string;
   user_id: string;
+  other_key_values?: Record<string, string>;
```
kristenarmes, line 9: what is this? I think I vaguely remember you did something with additional values for users, but can't remember the details. thinking this name should be made a little more descriptive so people know what it is/can be used for

frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
```diff
@@ -238,21 +241,19 @@ export class OwnerEditor extends React.Component<
...
+  renderOwnersSection = (section: OwnerCategory | null) => {
+    const { resourceType } = this.props;
+    const { itemProps } = this.state;
 
-    const ownerList = hasItems ? (
+    return (
       <ul className="component-list">
+        {section ? (
+          <div>
+            <span>{section.label}</span>
```
kristenarmes, line 252: can we add `className="title-3"` to this so it looks more like a title?

frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
```diff
@@ -238,21 +241,19 @@ export class OwnerEditor extends React.Component<
...
-        </div>
-      );
-    }
+  renderOwnersSection = (section: OwnerCategory | null) => {
+    const { resourceType } = this.props;
+    const { itemProps } = this.state;
 
-    const ownerList = hasItems ? (
+    return (
       <ul className="component-list">
```
kristenarmes, line 249: can you add a css class to the `<ul>` here that sets `margin-bottom: $spacer-1` (8px) but then sets it to 0 for the last-child? the larger sections have their own spacing so we want to skip it for the last item

frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
```diff
@@ -238,21 +241,19 @@ export class OwnerEditor extends React.Component<
...
-    const hasItems = Object.keys(itemProps).length > 0;
-
-    if (errorText) {
-      return (
-        <div className="owner-editor-component">
-          <span className="status-message">{errorText}</span>
-        </div>
-      );
-    }
+  renderOwnersSection = (section: OwnerCategory | null) => {
```
kristenarmes, line 244: we might want to consider doing some indication if there are no owners in a certain category. right now those titles look like floating text but isn't fully clear that it is an empty list

frontend/amundsen_application/static/js/interfaces/User.ts
B-T-D: It was named `other_key_values` on the backend User model so I thought the pattern might be just match that naming here?

frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
B-T-D: updated -- i used the styling that FrequentUsers uses for "No frequent users exist": https://github.com/amundsen-io/amundsen/blob/main/frontend/amundsen_application/static/js/pages/TableDetailPage/FrequentUsers/index.tsx#L64

kristenarmes (approved):
LGTM, thanks for making the updates!

PR #2268, opened 2024-10-03
fix -- handle uncategorized owners

B-T-D:
## Description

Prior PR didn't handle the case of having owner categories configured, but also having some owners without any categories.

## How Has This Been Tested?

Added unit tests for key cases, tested in our staging environment.

### CheckList
* [ ] PR title addresses the issue accurately and concisely
* [x] Adds tests

### DIFF
--- a/frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
+++ b/frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
@@ -252,7 +252,7 @@ export class OwnerEditor extends React.Component<
 
     // if any owners uncategorized, set section to null to render an uncategorized owners list
     Object.keys(itemProps).forEach((key) => {
-      if (!itemProps[key].additionalOwnerInfo.owner_category) {
+      if (!itemProps[key].additionalOwnerInfo?.owner_category) {
         section = null;
       }
     });
--- a/frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
+++ b/frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
@@ -324,8 +324,15 @@ export class OwnerEditor extends React.Component<
 
   renderOwnersList = () => {
     const sections = getOwnersSectionConfig().categories;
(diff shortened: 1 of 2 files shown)


frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
```diff
@@ -248,14 +248,25 @@ export class OwnerEditor extends React.Component<
     // check if rendering an owner category that lacks any entries
     let isEmptySection = false;
 
-    if (section) {
+    console.log(`itemProps: ${JSON.stringify(itemProps)}`);
```
kristenarmes: don't forget to remove log lines

frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
```diff
@@ -248,14 +248,25 @@ export class OwnerEditor extends React.Component<
...
+
+    // if any owners uncategorized, set section to null to render an uncategorized owners list
+    Object.keys(itemProps).forEach((key) => {
+      if (!itemProps[key].additionalOwnerInfo?.owner_category) {
+        section = null;
+      }
+    });
```
kristenarmes, line 256: is this needed if you do the other check before this function is called? this feels like it might be more complicated than it needs to be

kristenarmes (commented):
ideally this should have a couple unit tests to help avoid issues around this feature in the future, might be worth it to write a couple quick ones

frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
kristenarmes: nvm you beat me to it

frontend/amundsen_application/static/js/components/OwnerEditor/index.tsx
B-T-D: Yeah it wasn't needed both places, I deleted it from here.

kristenarmes (approved):

PR #2274, opened 2024-10-23
fix: Change databuilder search data extractors to publish `name` in user document.

glipR:
## Description
Changes the search data databuilder extractors to extract a `name` field, for publishing to elasticsearch.

Currently, the user document schema in elasticsearch expects a `name` keyword, which is the primary field for searching users. (Expected from [Databuilder publish](https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/task/search/document_mappings.py#L262), and expected from [search service](https://github.com/amundsen-io/amundsen/blob/main/search/search_service/proxy/es_proxy_v2_1.py#L82))

## Motivation and Context
The change is required because currently the elasticsearch query does not use the `name` field to create better matches for search results, and instead is likely just using first/last and key of the user. This is fine when searching for just first or last, but when searching both the search results currently aren't great for this reason.

This is actually fixed in the [example query](https://github.com/amundsen-io/amundsen/blob/main/databuilder/databuilder/task/search/search_data_queries.py#L127) ([PR](https://github.com/amundsen-io/amundsen/pull/1717)) for posting from neo4j to elasticsearch, but this query is never used.

## How Has This Been Tested?
This has only been tested with the neo4j_search_data_extractor, and not the other extractors. Testing was done by:
1. Loading from databuilder via the search_data_extractor script
2. Navigating to the index on `localhost:9200` and inspecting document values
3. Attempting a search in the frontend and comparing results

### Documentation
No change in documentation

### CheckList

Believe this fix has no need for:
* [X] N/A: ~Updates Documentation and Docstrings~
* [X] N/A: ~Adds tests~
* [X] N/A: ~Adds instrumentation (logs, or UI events)~

### DIFF
--- a/databuilder/databuilder/extractor/atlas_search_data_extractor.py
+++ b/databuilder/databuilder/extractor/atlas_search_data_extractor.py
@@ -204,6 +204,7 @@ class AtlasSearchDataExtractor(Extractor):
              lambda x: AtlasSearchDataExtractorHelpers.get_badges_from_classifications(x), [])
         ],
         'User': [
+            ('name', 'attributes.full_name', None, ''),
             ('email', 'attributes.qualifiedName', None, ''),
             ('first_name', 'attributes.first_name', None, ''),
             ('last_name', 'attributes.last_name', None, ''),
--- a/databuilder/databuilder/extractor/mysql_search_data_extractor.py
+++ b/databuilder/databuilder/extractor/mysql_search_data_extractor.py
@@ -426,6 +426,7 @@ def _user_search(session: Session, published_tag: str, limit: int) -> List[Dict]
 
             manager_email = user.manager.email if user.manager else ''
             user_result = dict(email=user.email,
+                               name=user.full_name,
                                first_name=user.first_name,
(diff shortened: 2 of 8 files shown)


kristenarmes:
@glipR thanks for your contribution! I triggered the CI tests and see a couple minor test failures. if you can resolve, I'd be happy to approve

glipR:
Thanks @kristenarmes ! I've edited the two failing tests to reflect the new return values / expected arguments.

Apologies I'm having trouble triggering the tests locally due to some version conflicts so may need a retrigger to confirm that that has fixed everything 🙏

glipR:
Didn't notice they'd be rerun, apologies for the delay :)

Looks like everything is passing now, thanks again for triggering these.

kristenarmes (approved):
LGTM, thanks!

kristenarmes:
I was about to merge, but realized did you want to bump the databuilder version? [here's the version spot](https://github.com/amundsen-io/amundsen/blob/main/databuilder/setup.py#L8)

PR #2278, opened 2024-11-15
fix: Adjustments to tabs components to prevent overflows

jdslaugh:
## Description
- `css/_layouts.scss` - Removed the min-heights and instead let it utilize the height of its container, which wont force it past acceptable limits

- `Table/styles.scss` - increased `min-width` to 100% to fill the container, set the header widths to auto and removed the top margin

- `TabsComponent/styles.scss` - Set the height to 100%, utilizing flex to allow the contents to grow but not overflow. Limited the overflow scrolling to the inner table versus the entire container

- `TableDetailPage/styles.scss` - Switched from fixed to absolute positioning so it would stop overflowing, removed the background and adjusted the positioning at smaller sizes so it wouldn't overlap

- `TableDetailPage/index.tsx` - Removed the visibility handlers that hid the Collapse all functionality and instead rotated it above the tabs at small enough sizes

## Motivation and Context
Scrolling in the container would cause the tabs components to overflow out of position

### Before
before

### After
after

## How Has This Been Tested?
Manually and comparison against production

### CheckList
* [ ] Adds tests

### DIFF
--- a/frontend/amundsen_application/static/js/pages/TableDetailPage/styles.scss
+++ b/frontend/amundsen_application/static/js/pages/TableDetailPage/styles.scss
@@ -60,10 +60,6 @@ $width-increment-5: $width-increment * 5;
     }
   }
 
-  .has-open-right-panel .nav.nav-tabs {
-    width: calc(100% - #{$total-panel-width});
-  }
-
   @media (max-width: $base-max-width) {
     .has-open-right-panel .nav.nav-tabs {
       width: calc(100% - #{$total-panel-width - $width-increment-1});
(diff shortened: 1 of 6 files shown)


frontend/amundsen_application/static/js/pages/TableDetailPage/styles.scss
```diff
@@ -113,20 +109,40 @@ $width-increment-5: $width-increment * 5;
...
-    position: fixed;
+    position: absolute;
     z-index: 6;
     padding: 20px $spacer-2 3px $spacer-1;
 
+    @mixin action-button-positions {
+      right: unset;
+      position: relative;
+      padding: 8px 0px 0px 14px;
+      margin-bottom: -16px;
```
Golodhros, line 124: Could we use spacer- tokens here?

Golodhros (approved):
LGTM

frontend/amundsen_application/static/js/pages/TableDetailPage/styles.scss
jecr, line 124:
```suggestion
      padding: $spacer-1 0px 0px 14px;
      margin-bottom: -$spacer-2;
```

B-T-D (approved):
Looks good based on the test in our staging environment.

